PR #1, opened 2019-03-07
Minor  fixes/improvements

smalleni:
1. Use field selector to get runnign pods
2. Remove unneeded export


smalleni:
@jtaleric PTAL

roles/uperf-bench/tasks/main.yml
```diff
@@ -52,6 +52,8 @@
     namespace: '{{ meta.namespace }}' 
     label_selectors:
       - app = bench-server
+    field_selectors:
+      - status.phase = running
```
jtaleric: This failed for me in the past... I will retry.

PR #4, opened 2019-03-08
Removed unneeded export

smalleni:


smalleni:
This needs updating of the operator image as well. So a new image has to be built after merging this.

smalleni:
@jtaleric PTAL

jtaleric:
Testing locally...

PR #6, opened 2019-03-10
Add sysbench

smalleni:
This commit adds sysbench as a new workload type


jtaleric:
@smalleni rebase please

roles/sysbench/meta/main.yml
```diff
@@ -0,0 +1,60 @@
+galaxy_info:
```
jtaleric: Any reason to include this?

roles/sysbench/README.md
```diff
@@ -0,0 +1,38 @@
+Role Name
```
jtaleric: Want to either delete or update this boilerplate stuff?

roles/sysbench/tasks/main.yml
```diff
@@ -0,0 +1,42 @@
...
+      metadata:
+        name: '{{ meta.name }}-sysbench-config'
+        namespace: '{{ meta.namespace }}'
+      data:
+        sysbenchScript: "{{ lookup('template', 'sysbench.sh.j2') }}"
+
+- name: start sysbench job
+  k8s:
+    definition:
+      kind: Job
```
jtaleric: You could just use a pod here instead of a job? Any specific reason for the Job?

roles/sysbench/templates/sysbench.sh.j2
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
```
jtaleric: Should we template this out more (considering we could add additional workloads using sysbench?).  We pass 
```
test:cpu
```
We could add a `if test == cpu` in the template

smalleni: All the basic tests (CPU, Memory) etc unless you go for a database workload(in which case you need to create a database first) have the same format sysbench --test=<name> params.  So it doesn't look like it is needed now?

roles/sysbench/meta/main.yml
smalleni: Boilerplate stuff

roles/sysbench/templates/sysbench.sh.j2
smalleni: Actually you are right..might need to change it.

roles/sysbench/tasks/main.yml
smalleni: I preferred the job because it terminated the pod as well. Just felt that was cleaner.

smalleni: Terminates pods but holds onto logs.

roles/sysbench/README.md
smalleni: Done.

smalleni:
@jtaleric Made the changes you requested. Enhanced the template to be a lot more smarter based on workload type. Also, now sysbench will take a list of tests like cpu, memory, fileio instead of one single tests. PTAL.

deploy/crds/bench_v1alpha1_bench_cr.yaml
```diff
@@ -21,3 +21,11 @@ spec:
     runtime: 57
     rw: write
     filesize: 1
+  sysbench:
+    tests:
```
aakarshg: It'd be in best interest to add a boolean like enabled, which can be defaulted to false.

playbook.yml
```diff
@@ -7,3 +7,6 @@
   - include_role:
       name: "fio-bench"
     when: fio is defined and fio.clients > 0
+  - include_role:
+      name: sysbench
+    when: sysbench is defined and sysbench.tests is defined sysbench.tests is not none
```
aakarshg: This condition can be then updated to also look at sysbench.enabled

smalleni: Well not really, you need to check if sysbench is defined first otherwise you would get an undefined var error.

deploy/crds/bench_v1alpha1_bench_cr.yaml
smalleni: The existing workloads don't have that. I was under the impression that if you didn't want to run the workload, you wouldn't have it in the CR in the first place. @jtaleric thoughts?

jtaleric: The existing workloads do have a concept of # of pairs or clients. which I think might make sense here too...

```yaml
sysbench:
  clients: 0
  tests:
   ...
```

Thoughts? 

And yes, of course users could just have a workload defined in their CR. I would prefer they could go either route.

smalleni: @jtaleric Ack. I can add the enabled flag. However, I have a questions regarding number of client filed. Is that a hard requirement? I run my workload as a job. I don't think there is a way to set replicas with job, so the only way would be loop over with ansible and start multiple jobs, which I'm not sure is what we want. Thoguhts?

smalleni:
@aakarshg @jtaleric PTAL :)

jtaleric:
ok to test

aakarshg:
rerun smoke

smalleni:
@aakarshg @jtaleric thanks for reviews!

PR #10, opened 2019-03-14
Adding logic for leftover servers

jtaleric:
Previously, if I removed the CR and added a new CR, the pods in
terminating state would get picked up. This is to fix that logic issue

### DIFF
--- a/roles/uperf-bench/tasks/main.yml
+++ b/roles/uperf-bench/tasks/main.yml
@@ -1,47 +1,59 @@
 ---
 
-- block:
+- name: Check for current_run
+  stat:
+    path: /tmp/current_run
+  register: state_file
 
-  - name: Wait for old pods to destroy...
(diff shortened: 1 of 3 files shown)


aakarshg:
/block 
this isn't working locally

roles/uperf-bench/tasks/main.yml
```diff
@@ -1,109 +1,126 @@
...
-        selector:
-          matchLabels:
-            app: uperf-bench-server
-        template:
-          metadata:
-            labels:
+
+- block:
+
+  - name: Wait for old pods to destroy...
```
aakarshg: There seems to be an issue with this logic, as it;s going into an inifinite loop and is not working.

jtaleric: This was working against my OCP4 cloud? I will retest

PR #32, opened 2019-03-24
Using kind:jobs for uperf/fio

aakarshg:
Also updated tests to look for job completion.
uperf: clients and fio both don't need to be running forever, so switched to jobs

### DIFF
--- a/roles/fio-bench/tasks/main.yml
+++ b/roles/fio-bench/tasks/main.yml
@@ -14,28 +14,32 @@
 - name: Start Server(s)
   k8s:
     definition:
-      kind: Pod
-      apiVersion: v1
+      kind: Job
+      apiVersion: batch/v1
       metadata:
         name: '{{ meta.name }}-fio-client-{{item}}-bench'
         namespace: '{{ meta.namespace }}'
-        labels:
-          app: fio-bench-server
       spec:
-        containers:
-        - name: bench-server
-          image: "quay.io/jtaleric/fio:testing"
-          command: ["/bin/sh","-c"]
-          args:
-           - "cat /tmp/fio/fiojob;
-              fio /tmp/fio/fiojob"
-          volumeMounts:
-          - name: config-volume
-            mountPath: "/tmp/fio"
-        volumes:
-        - name: config-volume
-          configMap:
-            name: fio-test
-        restartPolicy: Never
+        ttlSecondsAfterFinished: 600
+        template:
+          metadata:
+            labels:
+              app: fio-bench
+          spec:
+            containers:
+            - name: bench-server
+              image: "quay.io/jtaleric/fio:testing"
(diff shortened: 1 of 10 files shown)


aakarshg:
rerun smoke

roles/uperf-bench/tasks/main.yml
```diff
@@ -81,30 +81,33 @@
...
-        kind: Pod
-        apiVersion: v1
+        kind: Job
+        apiVersion: batch/v1
         metadata:
           name: '{{ meta.name }}-uperf-client-{{item.status.podIP}}'
           namespace: '{{ meta.namespace }}'
-          labels:
-            app : uperf-bench-client
         spec:
```
jtaleric: Should we use : `ttlSecondsAfterFinished` to automatically cleanup?

aakarshg:
rerun smoke

roles/uperf-bench/tasks/main.yml
aakarshg: Yes I think that'll be helpful, also I was thinking we can clean up server as soon as the client job completes successfully. Thus we'll keep the client pod but delete the serve pod once client finishes running. what'd you think?

jtaleric: Hm - I would think the other way around. Client completes, deletes. Server deletes? Not sure if the order really matters...

aakarshg: After discussion with @jtaleric on bluejeans, we've agreed upon deleting the server once the client job completes and use the `ttlSecondsAfterFinished` for client job.

aakarshg:
rerun smoke

aakarshg:
rerun smoke kube

aakarshg:
rerun smoke kube

aakarshg:
rerun smoke kube

aakarshg:
rerun smoke kube

aakarshg:
rerun smoke kube

aakarshg:
rerun smoke kube

aakarshg:
rerun smoke kube

aakarshg:
rerun smoke kube

aakarshg:
rerun smoke kube

roles/fio-bench/tasks/main.yml
```diff
@@ -14,28 +14,30 @@
...
+        template:
+          metadata:
+            labels:
+              app: fio-bench
+          spec:
+            containers:
+            - name: bench-server
+              image: "quay.io/jtaleric/fio:testing"
+              command: ["/bin/sh","-c"]
+              args: ["cat /tmp/fio/fiojob;fio /tmp/fio/fiojob"]
```
jtaleric: I don't think this reads as well, can you change it back so it reads easier?

aakarshg:
rerun smoke

PR #39, opened 2019-03-27
Sysbench template workload

### LINKED ISSUE #35: RFE : Jinja template workloads
Right now we are defining the job/pods in the tasks. 

We should consider jinja templates for a two reasons

1. Customizations passed by users (ie, Pass a PV or not)
2. Virtual Machines are treated differently

jtaleric:
Instead of defining the job/pod/etc in the task we should create a
template which can easily be modified to support different scenarios.

partially fixes #35

### DIFF
--- a/deploy/cluster_role_kubevirt.yaml
+++ b/deploy/cluster_role_kubevirt.yaml
@@ -1,0 +1,46 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: benchmark-operator-kube
+rules:
+- apiGroups:
+  - rbac.authorization.k8s.io
+  resources:
+  - roles
+  - rolebindings
+  - clusterroles
+  - clusterrolebindings
+  verbs:
+  - '*'
+- apiGroups:
+    - subresources.kubevirt.io
+  resources:
+    - virtualmachineinstances/console
+    - virtualmachineinstances/vnc
+  verbs:
+    - get
+- apiGroups:
+    - kubevirt.io
+  resources:
+    - virtualmachineinstances
+    - virtualmachines
+    - virtualmachineinstancepresets
+    - virtualmachineinstancereplicasets
+  verbs:
+    - get
+    - delete
+    - create
+    - update
+    - patch
+    - list
+    - watch
+    - deletecollection
+- apiGroups: [""]
+  resources:
+    - configmaps
+  resourceNames:
+    - kubevirt-config
+  verbs:
+    - update
+    - get
+    - patch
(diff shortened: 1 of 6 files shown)


roles/sysbench/tasks/main.yml
```diff
@@ -11,32 +11,14 @@
...
-                mountPath: "/tmp/"
-            volumes:
-            - name: sysbench-volume
-              configMap:
-                name: '{{ meta.name }}-sysbench-config'
-                defaultMode: 0777
-            restartPolicy: OnFailure
+    state: present
+    definition: "{{ lookup('template', 'workload.yml') | from_yaml }}"
+  when: workload.sysbench.kind != "VirtualMachineInstance"
```
aakarshg: workload.? shouldn't this be just sysbench.kind?

roles/sysbench/tasks/main.yml
```diff
@@ -11,32 +11,14 @@
...
+
+- name: Start sysbench vm job
+  k8s:
+    state: present
+    definition: "{{ lookup('template', 'workload_vm.yml') | from_yaml }}"
+  when:  workload.sysbench.kind == "VirtualMachineInstance"
```
aakarshg: workload.? shouldn't this be just sysbench.kind?

roles/sysbench/templates/workload_vm.yml
```diff
@@ -0,0 +1,34 @@
...
+          bus: virtio
+        name: containerdisk
+      - disk:
+          bus: virtio
+        name: workload
+    machine:
+      type: ""
+    resources:
+      requests:
+        memory: "{{ workload.sysbench.mem }}"
```
aakarshg: This should also read from cr file

jtaleric:
Going to change the logic here after discussing with @aakarshg 

CR will simply have,`vm:true|false (default: false)` I will create a different issue/PR to discuss the `workload` definition. Since I think we need a way to configure VMs / Containers with Memory/CPU shares/etc.

deploy/cluster_role_kubevirt.yaml
```diff
@@ -0,0 +1,46 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: benchmark-operator
```
aakarshg: we already have a ClusterRole with the name `benchmark-operator` at https://github.com/cloud-bulldozer/honeybadger/blob/master/deploy/cluster_role.yaml not sure how it'd work with 2 clusterrole sharing same names.

aakarshg:
Needs a rebase

aakarshg:
Merging this for now, but we'll need to look at documenting how we can use kubevirt for sysbench and if so whats the documentation required and I'm guessing this is something that'll be needed to be done not just at workload level.

PR #42, opened 2019-03-27
Re-organized resources and rbacs

### LINKED ISSUE #17: Consider restructuring deploy/
There is a handy k8s shortcut that allows you to `kubectl/oc create -f <directory>` and it will load all of the yaml files in that directory. The deployment instructions could be simplified by leveraging this shortcut, but the current deploy/ structure is problematic for this since the CRD needs to be loaded before the operator. If the CRD were in the deploy/ directory, then setup would be as simple as `kubectl/oc create -f deploy` and the operator would be up-and-running.

If the existing structure follows some established standard, then we could at least simplify the deployment instructions to say:

```bash
# oc create -f deploy/crds/bench_v1alpha1_bench_crd.yaml
# oc create -f deploy
```

aakarshg:
Also disabled couchbase test
fixes #17


aakarshg:
rerun minishift

aakarshg:
rerun minikube

jtaleric:
ready to merge?

Issue #37, opened 2019-03-27
RFE: Allow editing of versions and URLs for couchbase operator and pods in the userspace CR


dustinblack:
Currently, the URLs and versions for the CB operator and pods are defined within the role, and therefore coded into the operator image. We should allow for overriding default values with the CR.

dustinblack:
Completed with PR #43

PR #43, opened 2019-03-27
Better defaults and user modification of defaults through CR

dustinblack:
For issue #37 

Adding couchbase deployment

make operator and pod versions and urls variables changeable by CR

move default vars to defaults dir

making operator and pod versions and URLs user-configurable

### DIFF
--- a/docs/couchbase.md
+++ b/docs/couchbase.md
@@ -36,32 +36,42 @@ spec:
       use_persistent_storage: True
       class_name: "rook-ceph-block"
       volume_size: 10Gi
-    on_openshift: True
-    rh_pull_secret: <your registry.redhat.io pull secret>
 ```
 
 If you set `spec.couchbase.stroage.use_persistent_storage` to `true`, then you will need to provide a valid
 StorageClass name for `spec.couchbase.storage.class_name` and a valid volume size for `spec.couchbase.storage.volume_size`.
+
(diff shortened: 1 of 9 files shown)


dustinblack:
rerun smoke

aakarshg:
Seems to be failing on my 3.11 minishift environment at creating clusterrolebinding for the cb-operator :/ despite the other clusterrole, clusterrolebinding being applied successfully.

dustinblack:
> Please apply rebase and also update the docs reflecting the new changes.

Updates done, and tests passed locally on minishift.

dustinblack:
> Looks good for most part, we just need to edit the templates to make use of benchmark namespace instead of defaulting to current namespace. Another way we can do this, for now we can merge this if I'm able to get it to work locally and then make changes in #50

We currently use meta.namespace across other workloads, and the existing CI system (minishift, at least) would run tests within the benchmark namespace context already. I don't like the idea of hard-coding the benchmark namespace all over the place; I already don't like how much of this is in the deploy/ directory files.

dustinblack:
Test passed for me on local minikube (after dealing with lab-related network annoyances)

aakarshg:
Test passed on minishift if all other images were cleared from cache and rerun thrice. Will merge this as discussion related to namespace can take place at #54 .

PR #45, opened 2019-03-29
Breakout workload per CR

jtaleric:
Currently we ship a single CR. This PR allows for that same single PR
with `workload_all_cr`. Each worklaod will also have it's own CR with
this change.


jtaleric:
We need to merge #43 prior to this.

jtaleric:
I _could_ leave the old file (bench_v1alpha1_bench_cr.yaml) there (instead of placing in the workload_all file)... That way documentation would still be correct.

aakarshg:
> Currently we ship a single CR. This PR allows for that same single PR
> with `workload_all_cr`. Each worklaod will also have it's own CR with
> this change.

From this I thought we want users/contributers to create a workload specific cr file for the new workload they add right? so in that case we'll need to update docs, and add it to the list of things thats required before a PR is submitted. 
Also looks like there's another conflict that arose.

PR #48, opened 2019-04-01
removed molecule

aakarshg:
We don't use molecule currently, so removing it.


aakarshg:
The [documentation ](https://operatorhub.io/contribute) doesn't mention anything about molecule needing to be present, operator-sdk probably had created the directory and since we don't use it nor add to it, it's better to remove it. The presence of it might cause unnecessary confusion as the user will expect us to have molecule tests.
I feel the best way to maintain any code and keep it clean is to remove what we don't use nor maintain.

PR #49, opened 2019-04-01
New version of ansible-operator

jtaleric:
This version is necessary to work with KNI/RHHI.


aakarshg:
Tested locally on minishift okd 3.11 and k8s 1.13.4 through minikube. This is a good to go.

PR #50, opened 2019-04-01
Restructure deploy dir and default namespace to 'ripsaw'

dustinblack:
…amespace-level services to the benchmark namespace


dustinblack:
> Needs a rebase

Rebase complete

dustinblack:
> This will require updating the test.sh and tests/common.sh as well as they use it to apply definition.

This brings up an interesting question. If the operatory.yaml file is in the deploy/ directory, then a `kubectl apply -f deploy` will ultimately deploy the operator itself. Currently, the tests are designed to create and destroy the operator within each test functional test script. Would it not make more sense to start the operator once, run the sequence of tests, and then stop the operator? Or is there some reason that I'm not considering for the operator to start/stop with each functional test?

aakarshg:
> This brings up an interesting question. If the operatory.yaml file is in the deploy/ directory, then a `kubectl apply -f deploy` will ultimately deploy the operator itself. Currently, the tests are designed to create and destroy the operator within each test functional test script. 

Thats right, I didn't even think about that.  Maybe we'll need to leave out operator and crd in resource dir. 

> Would it not make more sense to start the operator once, run the sequence of tests, and then stop the operator? Or is there some reason that I'm not considering for the operator to start/stop with each functional test?

That's what we're doing with full_test.sh, except we're not running that in CI yet, but we'll eventually be doing that. The functional test idea was to sort of test each workload independently as if the operator was created just for the workload, and not worrying about resources that were left behind from previous workloads.

dustinblack:
> > This brings up an interesting question. If the operatory.yaml file is in the deploy/ directory, then a `kubectl apply -f deploy` will ultimately deploy the operator itself. Currently, the tests are designed to create and destroy the operator within each test functional test script.
> 
> Thats right, I didn't even think about that. Maybe we'll need to leave out operator and crd in resource dir.
> 
> > Would it not make more sense to start the operator once, run the sequence of tests, and then stop the operator? Or is there some reason that I'm not considering for the operator to start/stop with each functional test?
> 
> That's what we're doing with full_test.sh, except we're not running that in CI yet, but we'll eventually be doing that. The functional test idea was to sort of test each workload independently as if the operator was created just for the workload, and not worrying about resources that were left behind from previous workloads.

I guess I don't see the added value of re-starting the operator for each functional test. Maybe I'm missing something?

I would recommend one test.sh script, that does _not_ require an edit every time a test is added to the tests/ directory. Just glob the files from that directory that match the `test_*.sh` naming convention, start the operator, run all tests, stop the operator. If we want to get fancy, we can add a command line option to test.sh that allows you to list by name the tests you want to run, defaulting to all tests if no options are passed.

aakarshg:
The added value of re-starting operator is for the operator logs to have nothing old, and the new CR file being the only CR file it has ever encountered thus ensuring nothing was left behind that the operator is dealing with or was dealing with mid execution when we applied a new CR file. 
We already have the [full_test.sh](https://github.com/cloud-bulldozer/ripsaw/blob/master/full_test.sh) that accomplishes pretty much the same you're referring to, and we want to start with a fresh operator to ensure that everything is clean before we kick off a new workload and test it.

dustinblack:
> The added value of re-starting operator is for the operator logs to have nothing old, and the new CR file being the only CR file it has ever encountered thus ensuring nothing was left behind that the operator is dealing with or was dealing with mid execution when we applied a new CR file.
> We already have the [full_test.sh](https://github.com/cloud-bulldozer/ripsaw/blob/master/full_test.sh) that accomplishes pretty much the same you're referring to, and we want to start with a fresh operator to ensure that everything is clean before we kick off a new workload and test it.

Hmm. This bugs me a bit, as I feel like there is a more sustainable way to handle this. I'd like to be able to `kubectl apply -f deploy/` and have the operator up-and-running in its namespace, but making that happen right now with these changes breaks too much about the tests. So I will revert that part of the change for now and we can circle back on this later.

aakarshg:
What I was trying to say was that instead of doing `kubectl apply -f deploy` in our tests/common.sh we can split it like we'd earlier thus we'll allow for users to do `kubectl apply -f deploy` and then pick a cr file while testing we'd do `kubectl apply -f 001..` and so on.

dustinblack:
> What I was trying to say was that instead of doing `kubectl apply -f deploy` in our tests/common.sh we can split it like we'd earlier thus we'll allow for users to do `kubectl apply -f deploy` and then pick a cr file while testing we'd do `kubectl apply -f 001..` and so on.

That could make sense, but it could also make maintaining the test scripts more cumbersome since any additions to deploy/ would also mean that common.sh would need to be updated.

Let's think through this a bit more...

group_vars/all.yml
```diff
@@ -0,0 +1,2 @@
+---
+operator_namespace: ripsaw
```
jtaleric: to keep meta.namespace could we just do :

```yaml
meta :
  namespace: ripsaw
```

aakarshg:
This is closed now as #69 was merged.

PR #52, opened 2019-04-02
Minor tweaks for CI tests

aakarshg:
No wait is required if we're waiting for job to complete
Added a small wait for 5 seconds after deploying operator
Cleaning up benchmark-operator before deleting cb operator


aakarshg:
@dustinblack PTAL

tests/test_couchbase.sh
```diff
@@ -6,8 +6,8 @@ source tests/common.sh
 function finish {
   echo "Cleaning up couchbase"
   kubectl delete -f tests/test_crs/valid_couchbase.yaml
-  kubectl delete deployment couchbase-operator
   delete_operator
+  kubectl delete deployment couchbase-operator
```
dustinblack: I think this delete deployment may only need to be here at all in the case of a failed or aborted run. Anecdotally, it seems like it gets cleaned up properly without this line if the job finishes properly.

dustinblack:
Small comment, but it's not exactly related to this commit. Otherwise LGTM.

tests/test_couchbase.sh
aakarshg: yep that's right if the job finishes properly it gets cleaned up automatically, I think in that case I can probably add a check to see if the deployment is present and delete only if. Good catch @dustinblack

aakarshg:
Closing this and will work on a new PR since #69 got merged

PR #68, opened 2019-04-12
Upload uperf results

### LINKED ISSUE #26: RFE: Central collection point for results
Add a feature to tag workload results and upload them to a central repository (likely an object store). Could be configurable as a private repo, or could default to a public one where we can collect broad result sets for analysis.

aakarshg:
- Added new role result which will upload base64 encoded data to es
- Currently only uperf supports it

Closes #26
Co-authored-by: Joe (rook) <joe.talerico@gmail.com>

### DIFF
--- a/docs/uperf.md
+++ b/docs/uperf.md
@@ -35,6 +35,37 @@ Once done creating/editing the resource file, you can run it by:
 # kubectl apply -f <path_to_file> # if created a new cr file
 ```
 
+Note: If you'd like to try to experiment with storing results in a pv and have followed
+instructions to deploy operator with attached pvc and would like to send results to ES(elasticsearch).
+you can instead define a custom resource as follows:
+
+```yaml
+apiVersion: benchmark.example.com/v1alpha1
+kind: Benchmark
+metadata:
+  name: example-benchmark
+  namespace: ripsaw
+spec:
+  elasticsearch:
+    server: <es_host>
+    port: <es_port>
+    index: <es_index>
+  user: rht_perf_ci # user is a key that points to user triggering ripsaw, useful to search results in ES
+  store_results: true
+  results:
+    path: /opt/result-data/ # has to be what's passed to mountpath for operator pod
+  workload:
+    name: uperf
+    args:
+      # To disable uperf, set pairs to 0
+      pair: 1
+      proto: tcp
+      test_type: stream
+      nthr: 2
+      size: 16384
+      runtime: 10
+```
+
 Deploying the above(assuming pairs is set to 1) would result in
 
 ```bash
--- a/roles/result/tasks/main.yml
+++ b/roles/result/tasks/main.yml
@@ -1,0 +1,38 @@
+---
+
+- name: Get files
+  shell: find {{results.path}} -maxdepth 1 -printf '%f\n' | grep '^[0-9]'
+  ignore_errors: true
+  register: result_files
+
+- block:
+  - debug:
+      msg: "{{result_files}}"
+
+  - name: Result data
+    slurp:
+      src: "{{results.path}}/{{ item }}"
+    register: result_data
+    with_items: "{{result_files.stdout_lines}}"
+
+  - name: Index data
+    uri:
+      url: "http://{{ elasticsearch.server }}:{{ elasticsearch.port }}/{{ elasticsearch.index }}/doc"
+      method: POST
+      status_code: 201
+      body: "{{ lookup('template','result.json') }}"
+      body_format: json
+    with_items: "{{result_data.results}}"
+    register: indexed_files
+
+  - name: ensure the directory is present
+    file:
+      path: "{{results.path}}/indexed"
+      state: directory
+
+  - name: move all files
+    command: mv {{results.path}}/{{ item.item.item }} {{results.path}}/indexed/{{ item.item.item }}
+    with_items: "{{ indexed_files.results }}"
+    when: item.status == 201
+
+  when: result_files.rc == 0
--- a/roles/result/templates/result.json
+++ b/roles/result/templates/result.json
@@ -1,0 +1,7 @@
+{
+  "uuid": "{{ item.item | regex_replace('-.*','') | to_uuid }}",
+  "user": "{{user}}",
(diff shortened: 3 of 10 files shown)


tests/test_crs/valid_uperf.yaml
```diff
@@ -3,7 +3,14 @@ kind: Benchmark
 metadata:
   name: example-benchmark
 spec:
-  cleanup: false
+  es:
+    server: elk-b09-h30-r720xd.rdu.openstack.engineering.redhat.com
```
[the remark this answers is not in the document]
aakarshg: This is the test cr, so it can stay I believe

aakarshg:
Tested locally and have a sample document indexed into ES [here](http://elk-b09-h30-r720xd.rdu.openstack.engineering.redhat.com/app/kibana#/doc/AWnuKgN3geXZ7MzXQLuu/test/doc?id=AWoyI67tgeXZ7MzXQOkZ&_g=())

tests/test_crs/valid_uperf.yaml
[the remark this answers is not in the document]
jtaleric: My issue is more about the hostname, no the CR.

docs/uperf.md
```diff
@@ -35,6 +35,34 @@ Once done creating/editing the resource file, you can run it by:
...
+instructions to deploy operator with attached pvc and would like to send results to ES.
+you can instead define a custom resource like
+```yaml
+apiVersion: benchmark.example.com/v1alpha1
+kind: Benchmark
+metadata:
+  name: example-benchmark
+  namespace: ripsaw
+spec:
+  es:
```
jtaleric: should we not use an abbreviated name here? maybe elasticsearch?

docs/uperf.md
```diff
@@ -35,6 +35,34 @@ Once done creating/editing the resource file, you can run it by:
...
+    server: <es_host>
+    port: <es_port>
+    index: <es_index>
+  user: rht_perf_ci
```
jtaleric: Add context here, the `user:` here is the one executing the test. This will make finding results easier in ES.

docs/uperf.md
```diff
@@ -35,6 +35,34 @@ Once done creating/editing the resource file, you can run it by:
...
+  store_results: true
+  results:
```
jtaleric: I wonder if we could store this in the defaults, and allow the user to override (which should never be the case).

roles/result/tasks/main.yml
```diff
@@ -0,0 +1,41 @@
+---
+
```
jtaleric: We should also display the results, right?

docs/uperf.md
aakarshg: Yep probably would make more sense.

docs/uperf.md
aakarshg: Agreed will update the document.

roles/result/templates/result.json
```diff
@@ -0,0 +1,7 @@
+{
+  "uuid": "{{ item.item | regex_replace('-.*','') | to_uuid }}",
+  "user": "{{user}}",
+  "spec": {{  _benchmark_example_com_benchmark.spec | to_nice_json  }},
```
jtaleric: Shouldn't this just be a list of values?

jtaleric: Additionally, `_benchmark_example_com_benchmark` isn't many any assumptions? right? IE the user can pass a different `metadata.name` ?

aakarshg: Currently it's not a list of values, but once we finalize on #7 this would change accordingly, and nope only assumption is that the cr's apiVersion is `benchmark.example.com` and the kind is `benchmark`. So if we ever updated our CRD and thus CRs, we'll also have to update this. And yes, it shouldn't matter what `metadata.name` is

roles/result/tasks/main.yml
aakarshg: I'm against it, because the logs would be super ugly, if the user wants to look at results, they can always look at logs of the pods.

roles/result/templates/result.json
jtaleric: So it is predicated on some definition we have... The user _could_ change that... i would also recommend we remove "example" from the name.

roles/result/tasks/main.yml
jtaleric: How do you figure it would be ugly? If they look at the pod, it will be the same output :)

docs/uperf.md
aakarshg: You mean `results.path` should be set in `group_vars/all.yaml` and then for it to be overwritten through CR? but what about the inconsistency that's bound to be produced if the user doesn't mount it on the same locally for operator pod? I'd say it leave it as is, as it doesn't seem like we'd be able to solve a problem but if anything for a person developing on ripsaw it adds to more confusion

roles/result/tasks/main.yml
aakarshg: Yes, exactly my point if they can already get the output by looking at the pod, why also push it to logs of the operator pod which is already quite difficult to read.

roles/result/templates/result.json
aakarshg: The user can change the cr you mean? Then the operator will not be able to read it or work against it, I think its outside the scope of this PR and even ripsaw. we can't remove example from the name as its how we defined in crd.

aakarshg: When the user wants to create the resource based on the cr which if user changes apiversion to say something like `benchmark.exampl2e.com/v1alpha1` then they'd possibly hit an error like `[root@smicro-6029p-04 ripsaw]# kubectl apply -f valid_uperf_store.yaml
error: unable to recognize "valid_uperf_store.yaml": no matches for kind "Benchmark" in version "benchmark.exampl2e.com/v1alpha1"
`

tests/test_crs/valid_uperf.yaml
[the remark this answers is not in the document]
aakarshg: I'm trying to come up with a solution for the CI to use environment vars as well, so this can also be eliminated soon. For now, I can remove it if you think its best.

PR #70, opened 2019-04-23
Better structure to our CR

### LINKED ISSUE #7: RFE : New CR / Playbook structure
Right now our Custom resource looks like 
```yaml
spec:
  uperf:
    # To disable uperf, set pairs to 0
    pair: 1
    proto: tcp
    test_type: stream
    nthr: 2
    size: 16384
    runtime: 10
  fio:
    # To disable fio, set clients to 0
    clients: 0
    jobname: test-write
    bs: 4k
    iodepth: 4
    runtime: 57
    rw: write
    filesize: 1
```

We only have a limited definition. I would like to move to a list

```yaml
spec:
    infra:
      couchbase:
        servers: 1
    workloads:
      - uperf:
        # To disable uperf, set pairs to 0
        pair: 1
        ...
     - fio:
       # To disable fio, set clients to 0
       clients: 0
       jobname: test-write
     - ycsb:
       ...
```

Which would allow us to have a pipeline of workloads that we iterate through.

Also, defining the infrastructure (ie databases we will run against) allows us to kick off infrastructure and run many different workloads against it, vs just always relying on YCSB...

jtaleric:
This commit will provide a structure to our CR.

Closes #7


tests/test_crs/valid_fio.yaml
```diff
@@ -5,7 +5,8 @@ metadata:
   namespace: ripsaw
 spec:
   cleanup: false
-  fio:
+  name: fio
```
aakarshg: `workload` is missing and thats why CI was failing

PR #73, opened 2019-04-24
Use Redis for backend key/value store

jtaleric:
This commit was built to create a mechanism to sync workloads to run at
the sametime. There was a want to do this with FIO, however FIO might
have a built in mechanism to do this. We feel there is still a good
reason to keep this functionality for workloads that do not have built
in sync.

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -4,3 +4,7 @@ COPY group_vars/ ${HOME}/group_vars/
 COPY roles/ ${HOME}/roles/
 COPY watches.yaml ${HOME}/watches.yaml
 COPY playbook.yml ${HOME}/playbook.yml
+
+USER root
+RUN yum install -y redis
+USER 1001
--- a/resources/operator.yaml
+++ b/resources/operator.yaml
@@ -28,5 +28,23 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.name
+            - name: ANSIBLE_VERBOSITY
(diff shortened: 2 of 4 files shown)


resources/operator.yaml
```diff
@@ -28,5 +28,23 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.name
+            - name: ANSIBLE_VERBOSITY
```
aakarshg: Do we want to just add this to the normal definition of operator? We've another definition in `operator_store_results` which allows for indexing uperf/fio(pr_pending) results too. Should we look to add this work into that definition as well?

build/Dockerfile
```diff
@@ -4,3 +4,7 @@ COPY group_vars/ ${HOME}/group_vars/
 COPY roles/ ${HOME}/roles/
 COPY watches.yaml ${HOME}/watches.yaml
 COPY playbook.yml ${HOME}/playbook.yml
+
+USER root
+RUN yum install -y redis
+USER 1001
```
aakarshg: I don't have much understanding of what's going on here, but still curious why USER 1001?

jtaleric:
Great question! This is the user I noticed when building the container... I wonder if it could simply be ansible-operator?

resources/operator.yaml
jtaleric: I don't see why not? Since we build our operator, I would prefer we have everything built...

PR #58, opened 2019-04-03
fio-tests

shekharberry:
It will RUN FIO on using templates both on persistent storage provided by CEPH
Also if needed we can run FIO without using PV from CEPH.

Signed-off-by: Shekhar Berry <shberry@redhat.com>

PR #74, opened 2019-04-25
Distributed FIO

jtaleric:
FIO that uses the FIO server/client built in mechanism


roles/fio-distributed/tasks/main.yml
```diff
@@ -0,0 +1,55 @@
...
+    label_selectors:
+      - app = fio-benchmark
+  register: server_pods
+  until: "fiod.servers == (server_pods | json_query('resources[].status.podIP')|length)"
+  retries: 10
+  delay: 10
+
+- name: Create IP list
+  set_fact:
+    hosts: |
```
aakarshg: `hosts` isn't a safe variable to use and its a bit vague, so perhaps replace with `fio_servers`

roles/fio-distributed/templates/client.yaml
```diff
@@ -0,0 +1,36 @@
...
+apiVersion: batch/v1
+metadata:
+  name: '{{ meta.name }}-fio-client'
+  namespace: '{{ operator_namespace }}'
+spec:
+  ttlSecondsAfterFinished: 600
+  template:
+    metadata:
+      labels:
+        app: fio-client
```
aakarshg: Replace this with `fiod-client` would ensure no conflict takes place even if a user runs both together somehow

tests/test_crs/valid_fiod.yaml
```diff
@@ -0,0 +1,18 @@
...
+metadata:
+  name: example-benchmark
+  namespace: ripsaw
+spec:
+  cleanup: false
+  workload:
+    name: "fio_distributed"
+    args:
+      # To disable fio, set clients to 0
+      servers: 1
```
aakarshg: best to run with servers set 2 ?

tests/test_fiod.sh
```diff
@@ -0,0 +1,27 @@
...
+  kubectl delete -f tests/test_crs/valid_fiod.yaml
+  delete_operator
+}
+
+trap finish EXIT
+
+function functional_test_fio {
+  apply_operator
+  kubectl apply -f tests/test_crs/valid_fiod.yaml
+  check_pods 1
```
aakarshg: should be 3 as it'll be 2 server pods and 1 client pods

roles/fio-distributed/tasks/main.yml
jtaleric: I don't regard it as unsafe, but noted.

roles/fio-distributed/templates/client.yaml
jtaleric: Highly unlikely but alright.

tests/test_crs/valid_fiod.yaml
jtaleric: Why?

tests/test_fiod.sh
jtaleric: I will set to 2, good catch.

roles/fio-distributed/tasks/main.yml
aakarshg: That's right unsafe isn't the right word, the variable name doesn't represent the value it will hold, and is generic so can cause conflict. Best to avoid it if possible.

tests/test_crs/valid_fiod.yaml
aakarshg: We'd want to test with at least 2 servers so that we've better coverage say at a task like this `https://github.com/cloud-bulldozer/ripsaw/pull/74/files#diff-7aa81d437a25b88f200c4b77a2f138f9R32`

docs/fio_distributed.md
```diff
@@ -0,0 +1,28 @@
+# FIO Distributed
+
+FIO (Flexible I/O Tester) has a native mechanism to run multiple servers concurrently and summarize
```
aakarshg: Would be better if we can add a bit more documentation on how there's a client that gets launched, and the interaction of the client pod with the server pods if possible. I think this would be useful for a newcomer.

tests/test_fiod.sh
```diff
@@ -0,0 +1,27 @@
...
+
+trap finish EXIT
+
+function functional_test_fio {
+  apply_operator
+  kubectl apply -f tests/test_crs/valid_fiod.yaml
+  check_pods 3
+  fio_pod=$(kubectl get pods -l app=fio-client --namespace ripsaw -o name | cut -d/ -f2)
+  kubectl wait --for=condition=Initialized "pods/$fio_pod" --namespace ripsaw --timeout=200s
+  kubectl wait --for=condition=complete -l app=fio-client jobs --namespace ripsaw --timeout=100s
```
aakarshg: has to be be `app=fiod-client` as the label has been updated in the client definition

docs/fio_distributed.md
jtaleric: I added a single line, but that interaction is documented in the fio client pretty well. I would advise people to look there.

aakarshg:
The reason the minikube run ended succesfully despite the label issue is because test_fiod hasn't been added to test.sh. If you think its ready and would like to merge before #58 then we can add it to test.sh and start testing.

PR #77, opened 2019-05-01
Add support for more network scenarios

jtaleric:
Allow for :
- hostNetwork testing -- not using OpenShift SDN
- Pods to be scheduled to specific nodes using hostname

### DIFF
--- a/docs/uperf.md
+++ b/docs/uperf.md
@@ -5,20 +5,25 @@
 ## Running UPerf
 
 Given that you followed instructions to deploy operator,
-you can modify [cr.yaml](../resources/crds/benchmark_v1alpha1_benchmark_cr.yaml)
+you can modify [cr.yaml](../resources/crds/benchmark_v1alpha1_uperf_cr.yaml)
 
 Note: please ensure you set 0 for other workloads if editing the
-[cr.yaml](../resources/crds/benchmark_v1alpha1_benchmark_cr.yaml) file otherwise
+[cr.yaml](../resources/crds/benchmark_v1alpha1_uperf_cr.yaml) file otherwise
 your resource file should look like this:
 
 ```yaml
 apiVersion: benchmark.example.com/v1alpha1
 kind: Benchmark
 metadata:
-  name: example-benchmark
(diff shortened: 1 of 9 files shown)


docs/uperf.md
```diff
@@ -28,6 +33,22 @@ spec:
     runtime: 60
 ```
 
+`hostnetwork` will test the performance of the node the pod will run on.
+
+*Note:* If you want to run with hostnetwork, you will need to execute the following:
+
+```bash
+
+$ oc adm policy add-scc-to-user privileged -z benchmark-operator
```
aakarshg: What's the way to do this through `kubectl` ? It'd be helpful to add that here and then put forward saying if you've oc installed then you can accomplish this through the above mentioned.

jtaleric: Not sure. You can try this commit on a pure kube deployment to see if this is even needed.

jtaleric: Added a note for OpenShift env. Unsure about Kube.

roles/uperf-bench/templates/workload_store.yml.j2
```diff
@@ -12,14 +12,19 @@ spec:
...
+{% endif %}
       containers:
       - name: benchmark
         image: "quay.io/benchmark-operator/uperf:latest"
         command: ["/bin/sh", "-c"]
         args:
           - "cat /tmp/uperf-test/uperfXML;
              export h={{item.status.podIP}};
-             uperf -v -x -a -m /tmp/uperf-test/uperfXML -i 1 | tee {{results.path}}/$(date +%s)-uperf"
+             seq {{uperf.rerun}} | xargs -I -- uperf -v -x -a -m /tmp/uperf-test/uperfXML -i 1 | tee {{results.path}}/$(date +%s)-uperf"
```
aakarshg: I don't like creating new results such that there's no concrete manner to link the reruns, but this seems to be out of scope of this PR. This should definitely be addressed when we retrospect on indexing results.

aakarshg:
LGTM, merging it.

PR #81, opened 2019-05-13
Adding iperf3 benchmark

aakarshg:


resources/crds/benchmark_v1alpha1_iperf3_cr.yaml
```diff
@@ -0,0 +1,24 @@
...
+kind: Benchmark
+metadata:
+  name: iperf3-benchmark
+  namespace: ripsaw
+spec:
+  name: iperf3
+  args:
+    pairs: 1
+    hostnetwork: false
+    pin: true
```
jtaleric: default to false.

roles/iperf3-bench/tasks/main.yml
```diff
@@ -0,0 +1,54 @@
...
+  - include_role:
+      name: pre-result
+    when: store_results is defined and store_results
+
+  - name: Start Server(s)
+    k8s:
+      definition: "{{ lookup('template', 'server.yml.j2') | from_yaml }}"
+    with_sequence: start=1 count={{ iperf3.pairs }}
+
+  # TODO(aakarsh): check if all server pods are running
```
jtaleric: This still a todo?

roles/iperf3-bench/tasks/main.yml
```diff
@@ -0,0 +1,54 @@
...
+  - name: Waiting for Jobs to complete....
+    k8s_facts:
+      kind: Job
+      api_version: v1
+      namespace: '{{ operator_namespace }}'
+      label_selectors:
+        - app = iperf3-bench-client
+    register: client_jobs
+    until: "iperf3.pairs == (client_jobs | json_query('resources[].status.succeeded')|length)"
+    retries: 30
```
jtaleric: If the user passes `transmit_type: time` and `transmit_value: 400` this will fail

tests/test_iperf3.sh
```diff
@@ -0,0 +1,26 @@
...
+
+function functional_test_iperf {
+  apply_operator
+  kubectl apply -f tests/test_crs/valid_iperf3.yaml
+  check_pods 2
+  iperf_client_pod=$(kubectl get pods -l app=iperf3-bench-client -o name | cut -d/ -f2)
+  kubectl wait --for=condition=Initialized "pods/$iperf_client_pod" --timeout=200s
+  kubectl wait --for=condition=complete -l app=iperf3-bench-client jobs --timeout=100s
+  sleep 5
+  # ensuring that uperf actually ran and we can access metrics
```
jtaleric: typo

roles/iperf3-bench/tasks/main.yml
aakarshg: That was a smart catch, updated it to a max of 900 seconds.

jtaleric: Well... I would suggest we check if transmit_type is time, and if it is time, it is ` {{ transmit_value + 300 }} ` for example... If transmit_type isn't time, the current logic is fine (not the 900 sec)

tests/test_crs/valid_iperf3.yaml
```diff
@@ -0,0 +1,21 @@
...
+      transmit_type: time
+      transmit_value: 60
+      omit_start: 0
+      length_buffer: 128K
+      window_size: 64k
+      ip_tos: 0
+      mss: 900
+      streams: 1
+      extra_options_server: ' '
+      extra_options_client: ' '
```
jtaleric: Should you test w/ a client option like -u?

aakarshg: ack that's right let me add a couple of options to test.

workloads/ycsb-Dockerfile
```diff
@@ -1,8 +1,6 @@
 FROM centos:7
 
 USER root
-#RUN rpm -ivh https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
```
jtaleric: Random clean up? nice

aakarshg: yeah was annoying me, sorry.

PR #83, opened 2019-05-16
change all benchmark.example.com references to ripsaw.cloudbulldozer.io

dustinblack:
rename files to match new ripsaw.cloudbulldozer.io domain
fix names, update docs, and default CRs to on

We own cloudbulldozer.io now!

### DIFF
--- a/CONTRIBUTE.md
+++ b/CONTRIBUTE.md
@@ -77,7 +77,7 @@ These should be buildable by our CI system for maintaining a central public imag
 
 ### Workload triggers
 [CRD](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/) holds the definition of the resource.
-The operator triggers roles based on the conditions defined in [cr](resources/crds/benchmark_v1alpha1_benchmark_cr.yaml) which will influence which roles the
+The operator triggers roles based on the conditions defined in a CR ([example](resources/crds/ripsaw_v1alpha1_uperf_cr.yaml)) which will influence which roles the
 [playbook](playbook.yml) executes.
 Other vars may be defined that can modify the workload run conditions.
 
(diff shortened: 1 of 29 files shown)


jtaleric:
Nice! Thanks Dustin! The documentation fixes were much needed! Thanks!!

CONTRIBUTE.md
```diff
@@ -77,15 +77,15 @@ These should be buildable by our CI system for maintaining a central public imag
...
+The operator triggers roles based on the conditions defined in a CR ([example](resources/crds/ripsaw_v1alpha1_uperf_cr.yaml)) which will influence which roles the
 [playbook](playbook.yml) executes.
 Other vars may be defined that can modify the workload run conditions.
 
 For the sake of the example CR, please default all workloads to disabled.
 
 Example CR:
 ```yaml
-apiVersion: benchmark.example.com/v1alpha1
```
aakarshg: we'll also need to update the spec to follow the new structure.

roles/result/templates/result.json
```diff
@@ -1,7 +1,7 @@
 {
   "uuid": "{{ item.item | regex_replace('-.*','') | to_uuid }}",
   "user": "{{user}}",
-  "spec": {{  _benchmark_example_com_benchmark.spec | to_nice_json  }},
+  "spec": {{  _ripsaw.cloudbulldozer.io_benchmark.spec | to_nice_json  }},
```
aakarshg: I think this should be updated to `_ripsaw_cloudbulldozer_io_benchmark` as that's how ansible-operator sends the extra vars, more on this is here https://github.com/operator-framework/operator-sdk/blob/master/doc/ansible/dev/developer_guide.md#extra-vars-sent-to-ansible

PR #85, opened 2019-05-20
Indent properly volumes for operator container

jtudelag:
result-data volume was not indented properly in the operator Deployment. I was getting this error:

```
oc create -f resources/operator_store_results.yaml                                                                                                                                                                                                

The Deployment "benchmark-operator" is invalid: spec.template.spec.containers[0].volumeMounts[0].name: Not found: "result-data" 
```


aakarshg:
Thanks @jtudelag I had a much larger PR which also addressed this #78.

PR #89, opened 2019-05-22
Couchbase updates

dustinblack:
This is intended to go after PR #83. Please review for now, and I will rebase it later.


roles/result/templates/result.json
```diff
@@ -1,7 +1,7 @@
 {
   "uuid": "{{ item.item | regex_replace('-.*','') | to_uuid }}",
   "user": "{{user}}",
-  "spec": {{  _ripsaw_cloudbulldozer_io_benchmark.spec | to_nice_json  }},
+  "spec": {{  _ripsaw.cloudbulldozer.io_benchmark.spec | to_nice_json  }},
```
aakarshg: Please undo this change, there shouldn't be a `.` as its replaced with `_`

tests/test_ycsb-couchbase.sh
```diff
@@ -6,7 +6,9 @@ source tests/common.sh
 function finish {
   echo "Cleaning up couchbase"
   kubectl delete -f tests/test_crs/valid_ycsb-couchbase.yaml
-  kubectl delete deployment couchbase-operator
+  kubectl delete csv --all -n ripsaw
+  kubectl delete secret 1979710-benchmark-operator-ci-pull-secret -n ripsaw
```
aakarshg: Like this +1 for this.

PR #90, opened 2019-05-23
Small file workload

chawlanikhil24:
With this PR comes the small file operator in line for ripsaw operator framework.


playbook.yml
```diff
@@ -30,7 +30,7 @@
         name: sysbench
       vars:
         sysbench : "{{ workload.args }}"
-      when: workload.name == "sysbench" and workload.args.enabled is defined and workload.args.enabled
+      when: workload.name == "sysbench" and workload.args.client > 0 and workload.args.enabled
```
jtaleric: Any specific reason for this change?

resources/operator.yaml
```diff
@@ -17,7 +17,7 @@ spec:
       containers:
         - name: benchmark-operator
           # Replace this with the built image name
-          image: quay.io/benchmark-operator/benchmark-operator:latest
+          image: "quay.io/nichawla/benchmark_op_nikhil:testing"
```
jtaleric: Let's keep this using the main image.

roles/smallfile-bench/README.md
```diff
@@ -0,0 +1,38 @@
+Role Name
```
jtaleric: Boiler plate file, can be removed or update to reflect the smallfile workload.

roles/smallfile-bench/tasks/main.yml
```diff
@@ -0,0 +1,50 @@
...
+              image: "quay.io/nichawla/smallfile_container:testing"
+              command: ["/bin/sh", "-c"]
+              workingDir: /root/smallfile-master/
+              args:
+                 - "mkdir /mnt/test_smallfile/;cat /mnt/smallfile/smallfilejob ; python smallfile_cli.py --yaml-input-file /mnt/smallfile/smallfilejob --output-json /root/SMALLFILE_RUN_RESULT; ls -R /mnt/"
+              volumeMounts:
+              - name: config-volume
+                mountPath: "/mnt/smallfile"
+
+              runAsUser: 1000 #0 means run as root
```
jtaleric: is this due to the mountpoint or something that smallfile needs? 

With FIO I default the mount point to /tmp

roles/smallfile-bench/tasks/main.yml
```diff
@@ -0,0 +1,50 @@
...
+            labels:
+              app: smallfile-benchmark
+          spec:
+            containers:
+            - name: benchmark-server
+              image: "quay.io/nichawla/smallfile_container:testing"
+              command: ["/bin/sh", "-c"]
+              workingDir: /root/smallfile-master/
+              args:
+                 - "mkdir /mnt/test_smallfile/;cat /mnt/smallfile/smallfilejob ; python smallfile_cli.py --yaml-input-file /mnt/smallfile/smallfilejob --output-json /root/SMALLFILE_RUN_RESULT; ls -R /mnt/"
```
jtaleric: You can break with newlines so it is easier on the eyes.

roles/smallfile-bench/tasks/main.yml
```diff
@@ -0,0 +1,50 @@
...
+      spec:
+        ttlSecondsAfterFinished: 600
+        template:
+          metadata:
+            labels:
+              app: smallfile-benchmark
+          spec:
+            containers:
+            - name: benchmark-server
+              image: "quay.io/nichawla/smallfile_container:testing"
```
jtaleric: We will want to build this into our benchmark-operator namespace

roles/smallfile-bench/tasks/main.yml
```diff
@@ -0,0 +1,50 @@
...
+        namespace: '{{ operator_namespace }}'
+        name: smallfile-test
+      data:
+        smallfilejob: "{{ lookup('template', 'smallfilejob.yaml.j2') }}"
+  when: smallfile.clients > 0
+
+- name: Start Run(s)
+  k8s:
+    definition:
+      kind: Job
```
jtaleric: I would recommend we create a template from this jobfile, so we can easily manipulate it -- ie add a pvc eventually

bengland2:
@jtaleric @chawlanikhil24 The Dockerfile that I use at present is:

https://github.com/distributed-system-analysis/smallfile/blob/master/examples/docker/smallfile/Dockerfile

am just seeing this PR now, will get back to you on it.

resources/crds/benchmark_v1alpha1_smallfile.yaml
```diff
@@ -0,0 +1,16 @@
...
+  workload:
+    name: smallfile
+    args:
+      clients: 1
+      operation: create
+      threads: 1
+      file-size: 4
+      files: 100
+      top: /tmp/smallfile/
+      fsync: y
```
bengland2: I don't think you want fsync: y.    This will cause an fsync after every file create, which is very expensive and not the typical use case.    Also, you want to add these parameters to improve your data collection:

```
args:
...
   threads: 1
   response-times: Y
   output-json:  <wherever you want to put it>
```

output-json means that you don't have to screen-scrape to get results from smallfile.    

response-times: Y will cause each smallfile thread (in each pod) to save response times for all of its file operations in $top/network_shared/rsptimes_<thread_id>_<host>_<operation>_<time>.csv .   The smallfile_rsptimes_stats.py program postprocesses those response times into response time percentiles, including percentiles as a function of time.

threads: 1 means that only 1 thread per pod will be run.   Perhaps we want this to be user-settable but typically you can get more concurrency by just adding more pods, and this has the advantage of letting Kubernetes spread your threads across the available physical hosts more evenly.

roles/smallfile-bench/tasks/main.yml
bengland2: I built one in [quay.io](https://quay.io/repository/bengland2/smallfile)  here using the aforementioned Dockerfile.  How does your image build differ from the one in the [smallfile tree here](https://quay.io/repository/bengland2/smallfile?tab=tags)?   I don't want to have too many of these images - for example if smallfile sources change, we need to be able to know which images to rebuild to distribute those changes.

roles/smallfile-bench/templates/smallfilejob.yaml.j2
```diff
@@ -0,0 +1,6 @@
+operation: create
```
bengland2: The biggest problem right now is that we can only pass one operation type to the operator, since a single smallfile run takes only 1 operation type.   Ideally I'd like to be able to define a test matrix by passing lists of parameter values and have the smallfile operator iterate on them.   Do we need to change the operator or smallfile?  In the past Alex C and I have let [Ceph Benchmarking Tool handle this](https://nuget.pkg.github.com/bengland2/cbt/blob/smallfile/example/smallfile-test-set.yaml) - it accepts parameter lists and can invoke smallfile multiple times to implement this.   If it's possible for Ripsaw to do this, then I'd prefer that.

roles/smallfile-bench/tests/inventory
```diff
@@ -0,0 +1,2 @@
+localhost
```
bengland2: what does localhost mean here?     Typically I use cluster loader to create a set of pods, and each pod is assigned a unique ID so that it can listen to the shared directory for test requests from a smallfile initiator.   The Dockerfile just starts the [launch.sh script](https://github.com/distributed-system-analysis/smallfile/blob/master/examples/docker/smallfile/launch.sh)  with the 2 specified environment variables, the script waits for a test to do, that's all a smallfile pod does by default.

roles/smallfile-bench/tests/test.yml
```diff
@@ -0,0 +1,5 @@
+---
+- hosts: localhost
```
bengland2: again, why is "localhost" used?

roles/smallfile-bench/tasks/main.yml
chawlanikhil24: There is nothing special needed by smallfile, I just changed the mountpoint for experimental purposes

roles/smallfile-bench/tasks/main.yml
chawlanikhil24: There is a template created in roles/smallfile-bench/templates/ , which is further served as yaml input to smallfile

roles/smallfile-bench/tasks/main.yml
chawlanikhil24: Ben, since I didn't have the idea about the available smallfile containers already , hence I constructed my own, but it can be replaced.

roles/smallfile-bench/tasks/main.yml
chawlanikhil24: Also, yes, I went through fio-bench's tasks/main.yaml, I saw that there were tasks to create a PVC and add it to the container. I can copy the same procedure here.

resources/crds/benchmark_v1alpha1_smallfile.yaml
chawlanikhil24: I agree to the idea of 1 thread per pod, which we followed for fio as well. I will omit this and add another patch. 

For  gluster use case, we generally use to do a fsync after every run to clear the buffers to create more stable run ( when using fio). Hence, in smallfile , i thought it would be good to add fsync after the run. If you still think , fsync is not need, then i can clear it out. 

output-json, I am already using this option in the cli command, and it is capturing the results at the desired location.

roles/smallfile-bench/tests/inventory
chawlanikhil24: This script was added by ansible-galaxy command directly , I've no idea about this yet, how test cases will be executed here.  @jtaleric I might need help on this.

roles/smallfile-bench/tests/test.yml
chawlanikhil24: Just mentioned the reason in the above comment.

roles/smallfile-bench/templates/smallfilejob.yaml.j2
chawlanikhil24: Yupp, I've kept it in the phase 2 of the task . I think we can do this, but there will be a limitation. I still need to explore it a bit more.

playbook.yml
chawlanikhil24: reverting this , AFAIR it was for experimental puposes

resources/operator.yaml
```diff
@@ -17,7 +17,7 @@ spec:
       containers:
         - name: benchmark-operator
           # Replace this with the built image name
-          image: quay.io/benchmark-operator/benchmark-operator:latest
+          image: quay.io/nichawla/benchmark_op_nikhil:testing
```
aakarshg: update required

roles/smallfile-bench/tests/inventory
aakarshg: You can remove the tests directory altogether.

jtaleric: Boiler plate, can be removed.

roles/smallfile-bench/tests/test.yml
jtaleric: More boilerplate.

aakarshg:
Needs a rebase.

resources/operator.yaml
aakarshg: yep, when the PR is ready update it to master as it should be after the rebase.

roles/smallfile-bench/README.md
aakarshg: Remove the file altogether.

.gitignore
```diff
@@ -1 +1,2 @@
 # Ignore
+resources/operator.yaml
```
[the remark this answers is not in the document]
chawlanikhil24: I add it for my case only, cause I am building and pushing the operator to my repo. .gitignore should not be a part of this PR. I'll revert this.

resources/crds/benchmark_v1alpha1_smallfile.yaml
bengland2: In smallfile, "--fsync Y" means that it issues an [fsync call before each file is closed](https://github.com/distributed-system-analysis/smallfile/blob/master/smallfile.py#L1089).   While this can sometimes be a useful test for debugging certain performance problems, it is not what you want to do by default.    I assume that smallfile parameters can be added to the CR yaml if needed.

roles/smallfile-bench/tasks/main.yml
bengland2: this command is missing ```--host-set pods.list --launch-by-daemon Y``` .    To see what I mean, see examples/docker/run-smallfile-client-tests.sh in the source tree.   It demonstrates all this.  For example, 

/run-smallfile-client-tests.sh /var/tmp/smf ~/smallfile quay.io/bengland2/smallfile:20190411.b

resources/operator.yaml
jtaleric: Still needs fixing

roles/smallfile-bench/tasks/main.yml
```diff
@@ -0,0 +1,159 @@
...
+- set_fact:
+    smallfile_path: /mnt/pvc
+  when: smallfile.storageclass is defined
+
+- set_fact:
+    smallfile_path: /tmp/test_smallfile/
+  when: smallfile.storageclass is not defined
+
+- debug:
+    var: smallfile_path
```
jtaleric: Scrap debug

roles/smallfile-bench/tasks/main.yml
```diff
@@ -0,0 +1,159 @@
...
+          spec:
+            ttlSecondsAfterFinished: 600
+            template:
+              metadata:
+                labels:
+                  app: smallfile-benchmark
+              spec:
+                containers:
+                  - name: benchmark-server
+                    image: "quay.io/nichawla/smallfile_container:testing"
```
jtaleric: Let's get this into Magazine.

bengland2:
I think this pull request should be withdrawn since you replaced it with PR 109.

aakarshg:
> I think this pull request should be withdrawn since you replaced it with PR 109.

likewise, @chawlanikhil24 do you want to close this out in favor of #109 ?

jtaleric:
Can we abandon this @chawlanikhil24

chawlanikhil24:
> Can we abandon this @chawlanikhil24

Yupp, no issues in closing this PR

PR #95, opened 2019-05-31
Add operator marketplace to dependencies for CI

aakarshg:


tests/common.sh
```diff
@@ -18,7 +18,30 @@ function delete_operator {
   kubectl delete -f resources/operator.yaml
 }
 
+function olm_dep {
```
dustinblack: So technically this is the marketplace, not the OLM (the OLM is a dependency of the marketplace, and it is already on the CI systems).

tests/common.sh
```diff
@@ -18,7 +18,30 @@ function delete_operator {
...
+  kubectl apply -f https://raw.githubusercontent.com/operator-framework/operator-marketplace/master/deploy/upstream/01_namespace.yaml
+  kubectl apply -f https://raw.githubusercontent.com/operator-framework/operator-marketplace/master/deploy/upstream/02_catalogsourceconfig.crd.yaml
+  kubectl apply -f https://raw.githubusercontent.com/operator-framework/operator-marketplace/master/deploy/upstream/03_operatorsource.crd.yaml
+  kubectl apply -f https://raw.githubusercontent.com/operator-framework/operator-marketplace/master/deploy/upstream/04_service_account.yaml
+  kubectl apply -f https://raw.githubusercontent.com/operator-framework/operator-marketplace/master/deploy/upstream/05_role.yaml
+  kubectl apply -f https://raw.githubusercontent.com/operator-framework/operator-marketplace/master/deploy/upstream/06_role_binding.yaml
+  kubectl apply -f https://raw.githubusercontent.com/operator-framework/operator-marketplace/master/deploy/upstream/07_upstream_operatorsource.cr.yaml
+  kubectl apply -f https://raw.githubusercontent.com/operator-framework/operator-marketplace/master/deploy/upstream/08_operator.yaml
```
dustinblack: Applying each file technically works, but if they change the structure then we have to maintain against it.

tests/common.sh
aakarshg: ack will update it.

tests/common.sh
aakarshg: it makes sense, i'll try the sparse checkout

tests/common.sh
```diff
@@ -18,7 +18,21 @@ function delete_operator {
   kubectl delete -f resources/operator.yaml
 }
 
+function marketplace_setup {
+  git clone --depth 1 --no-checkout --filter=blob:none   "https://github.com/operator-framework/operator-marketplace/" local_repo
+  cd local_repo
```
dustinblack: Maybe `pushd local_repo` and then `popd` below instead of `cd ..`

tests/common.sh
```diff
@@ -18,7 +18,21 @@ function delete_operator {
   kubectl delete -f resources/operator.yaml
 }
 
+function marketplace_setup {
+  git clone --depth 1 --no-checkout --filter=blob:none   "https://github.com/operator-framework/operator-marketplace/" local_repo
```
dustinblack: This still seems to do a full clone, which I thought you were trying to avoid.

aakarshg:
Falling back to the original approach of applying the 8 files files and deleting on exit. The original approach works better for the following reasons:
- No need to clone the repo or a part of it
- Simplifying the task
- If there's any change to directory structure, the builds would fail so we would notice immediately.

PR #98, opened 2019-06-04
Updating image tag to master

aakarshg:
Moving to use quay.io trigger for git push, this would mean that the tag
will be updated to branch name.

Also updated documentation and operator definition to pull from master tag.


aakarshg:
Example build is [here](https://quay.io/repository/benchmark-operator/benchmark-operator?tab=builds) which was triggered as #81 got merged.

PR #100, opened 2019-06-04
UPerf sync all clients

jtaleric:
Using our Redis backend we needed to sync all the uperf clients to start
up at the same time. This commit finally makes use of said backend.


tests/test_uperf.sh
```diff
@@ -17,7 +17,7 @@ function functional_test_uperf {
   check_pods 2
   uperf_client_pod=$(kubectl get pods -l app=uperf-bench-client -o name | cut -d/ -f2)
   kubectl wait --for=condition=Initialized "pods/$uperf_client_pod" --timeout=200s
-  kubectl wait --for=condition=complete -l app=uperf-bench-client jobs --timeout=100s
+  kubectl wait --for=condition=complete -l app=uperf-bench-client jobs --timeout=300s
```
aakarshg: We can avoid the long timeout by reducing the test payload, but it doesn't have to happen in this PR. Just something we should keep in back of our mind.

PR #103, opened 2019-06-06
Adding logic to generate UUID

aakarshg:
Useful when storing/indexing results from running workloads.

### DIFF
--- a/playbook.yml
+++ b/playbook.yml
@@ -13,6 +13,8 @@
 
   - name: Run Workload
     block:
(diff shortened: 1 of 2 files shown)


playbook.yml
```diff
@@ -13,6 +13,9 @@
 
   - name: Run Workload
     block:
+    - include_role:
+        name: "uuid"
+      when: store_results is defined and store_results
```
jtaleric: Not necessary.

PR #105, opened 2019-06-07
Pulling up logic to capture state

aakarshg:
Logic used by uperf to ensure subsequent runs of ansible-runner
doesn't recreate jobs/pods, is pulled up.


aakarshg:
WIP currently

playbook.yml
```diff
@@ -1,6 +1,55 @@
 - hosts: localhost
   gather_facts: no
   tasks:
+    - name: Check for running uperf pods...
```
jtaleric: s/uperf//g

playbook.yml
```diff
@@ -1,6 +1,55 @@
...
+      set_fact:
+        prev_values: |
+            [
+            {% for item in contents %}
+              "{{ item }}:{{ contents[item] }}",
+            {% endfor %}
+            ]
+        curr_values: |
+            [
+            {% for item in uperf %}
```
jtaleric: s/uperf/workload/g

playbook.yml
```diff
@@ -1,6 +1,55 @@
...
+              "{{ item }}:{{ uperf[item] }}",
```
jtaleric: s/uperf/workload/g

aakarshg:
> I see this is a WIP, but lets add a document here... Additionally, for debugging, maybe throw a clause in the when to ignore this logic, if `debug: true`, thoughts?

Ooooh, I like the thought and although it maybe confusing. We'll add documentation on what advanced users can tweak about wth ripsaw and this will be an option in there.

aakarshg:
Noticed some problems with the current approach, we should discuss this at length.

jtaleric:
> Noticed some problems with the current approach, we should discuss this at length.

Do we want to discuss here?

aakarshg:
> > Noticed some problems with the current approach, we should discuss this at length.
> 
> Do we want to discuss here?

I'll try to table a reply here as well as open this up for broad discussion on the wednesday meeting.

aakarshg:
#176 is a better approach to solve the issue, closing this.

PR #106, opened 2019-06-11
UPerf Indexing

jtaleric:
Work to index UPerf data


docs/uperf.md
```diff
@@ -111,61 +133,36 @@ Note: If cleanup is not set in the spec file then the client pods will be killed
...
+          median: 72580096.0
+          average: 63342843.6066
+          95th: 75016192.0
++-------------------------------------------------------------------------------------------------------------------+
+
 ```
+
+# Dashboard example
+
+Using the Elasticsearch storage describe above, we can build dashboards like the below.
```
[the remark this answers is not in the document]
jtaleric: No template yet. Noted. Added documentation on the uperf-results index.

docs/uperf.md
```diff
@@ -71,34 +83,46 @@ metadata:
   name: example-benchmark
   namespace: ripsaw
 spec:
+  user: test_user # user is a key that points to user triggering ripsaw, useful to search results in ES
```
dustinblack: If the 'user' key is an arbitrary value for us to search/match on in Elastic, I'd prefer to use a different key name that makes this more clear. The 'user' key to me implies a username/userid that I need to get right in order for things to work.

resources/crds/ripsaw_v1alpha1_uperf_cr.yaml
```diff
@@ -12,7 +12,7 @@ spec:
       pin: false
       pin_server: "node-0"
       pin_client: "node-1"
-      rerun: 1
+      samples: 1
       pair: 1
       test_types:
         - stream
```
dustinblack: The spec.user and spec.elasticsearch sections are in the docs above, but they are not included in the example CR here.

PR #107, opened 2019-06-13
Add role for postgres-operator (zalando)

dustinblack:


roles/postgres-infra/README.md
```diff
@@ -0,0 +1,11 @@
+postgres-infra
```
aakarshg: Remove this file please

roles/postgres-infra/handlers/main.yml
```diff
@@ -0,0 +1,2 @@
+---
```
aakarshg: This can also be removed

roles/postgres-infra/vars/main.yml
```diff
@@ -0,0 +1 @@
+---
```
aakarshg: This file is also not required, so can be removed

tests/test_postgres.sh
```diff
@@ -0,0 +1,54 @@
...
+  apply_operator
+  #marketplace_setup
+  #kubectl apply -f /root/.1979710-benchmark-operator-ci-pull-secret.yaml -n ripsaw
+  sleep 15
+  kubectl apply -f tests/test_crs/valid_postgres.yaml
+  pg_operator_pod=$(get_pod 'name=postgres-operator' 300)
+  kubectl wait --for=condition=Initialized "pods/$pg_operator_pod" --namespace ripsaw --timeout=60s
+  kubectl wait --for=condition=Ready "pods/$pg_operator_pod" --namespace ripsaw --timeout=300s
+  pg_app_pod=$(get_pod 'spilo-role=master' 600)
+  kubectl wait --for=condition=Initialized "pods/$cb_app_pod" --namespace ripsaw --timeout=60s
```
aakarshg: cb_app_pod ? is it supposed to be pg_operator_pod instead?

tests/test_postgres.sh
```diff
@@ -0,0 +1,54 @@
...
+  echo $pod_name
+  return 0
+}
+
+trap finish EXIT
+
+
+# Note we don't test persistent storage here
+function functional_test_postgres{
+  apply_operator
```
aakarshg: CI chokes on this line http://porthos.perf.lab.eng.rdu2.redhat.com:8080/job/machinist_githubpr_minikube/246/console

dustinblack:
Because CI was choking on some related things after adding the finalizer concept, I have a couple of additional commits in this PR now. I'm not going to squash these because they are not explicitly part of the same change and should be considered independently, though given the CI problems they should probably be taken together in this one PR.

resources/crds/ripsaw_v1alpha1_postgres_cr.yaml
```diff
@@ -0,0 +1,34 @@
...
+  namespace: ripsaw
+spec:
+  infrastructure:
+    name: postgres
+    args:
+      servers:
+        # Typical deployment size is 3
+        size: 1
+      storage:
+        # persistent storage volume is required
```
jtaleric: In the docs you mention, `use_persistent_storage: True` however here it mentions persistent storage is required. Do we require it?

roles/couchbase-infra/tasks/main.yml
```diff
@@ -1,121 +1,198 @@
...
+    #  kind: ClusterServiceVersion
+    #  namespace: "{{ operator_namespace }}"
+    #  name: "{{ item.metadata.name }}"
+    #register: cb_op_pod
+    #until: "'Running' in (cb_op_pod | json_query('resources[].status.phase'))"
+    #with_items: "{{ csvs.resources }}"
+    #retries: 10
+    #delay: 10
+    pause:
+      seconds: 100
```
jtaleric: Instead of a fixed amount of time, should we query and have a retry w/ delay?

roles/postgres-infra/templates/postgres-cluster.yaml.j2
```diff
@@ -0,0 +1,89 @@
...
+    zalando:
+    - superuser
+    - createdb
+  enableMasterLoadBalancer: true
+  enableReplicaLoadBalancer: true
+  allowedSourceRanges: # load balancers' source ranges for both master and replica services
+  - 127.0.0.1/32
+  databases:
+    foo: zalando
+#Expert section
```
jtaleric: Do we want to expose any of this in the CR to allow users to overwrite these values?

resources/crds/ripsaw_v1alpha1_postgres_cr.yaml
dustinblack: Required persistent storage was a later discovery about the operator. I'll update the doc to match.

roles/couchbase-infra/tasks/main.yml
dustinblack: Yes, but there is the problem with the above commented-out block, so the sleep was a stop-gap. I haven't come up with a way to test for "upness" that is consistent between k8s/OCP3 and OCP4 because of a difference in metadata.

roles/postgres-infra/templates/postgres-cluster.yaml.j2
dustinblack: Most likely, yes. My initial aim was to enable automation for CI validation of pgbench, mostly, so defaults are fine. I might need feedback from a postgres expert to best determine what to expose in the CR.

roles/ycsb-bench/tasks/main.yml
```diff
@@ -1,98 +1,127 @@
...
+
+
+# Destroy block is run when the finalizer has set teardown=true
+- name: Destroy
+  when: teardown is defined
+  block:
+
+  - name: Get YCSB workload pod
+    k8s_facts:
+      kind: Pod
```
aakarshg: This should be a job, and wouldn't operator-framework by default cleanup resources the operaator created which in this case is the jobs

roles/ycsb-bench/tasks/main.yml
```diff
@@ -1,98 +1,127 @@
...
+      namespace: "{{ operator_namespace }}"
+      label_selectors:
+        - job-name = ycsb-couchbase-ycsb-bench-job
```
aakarshg: This is hardcoded for the couchbase job, and I don't think any of this is required, as operator framework should take care of this

tests/test_postgres.sh
```diff
@@ -0,0 +1,81 @@
...
+
+
+# Note we don't test persistent storage here
+function functional_test_postgres {
+  apply_operator
+  ripsaw_pod=$(get_pod 'name=benchmark-operator' 300)
+  kubectl wait --for=condition=Initialized "pods/$ripsaw_pod" --namespace ripsaw --timeout=60s
+  kubectl wait --for=condition=Ready "pods/$ripsaw_pod" --namespace ripsaw --timeout=300s
+  # first teardown any previous resources from this test
+  kubectl apply -f tests/test_crs/teardown_postgres.yaml
```
aakarshg: I don't think this is necessary.

aakarshg:
closing this re: #121 and this made its way into https://github.com/cloud-bulldozer/cloud-builder

PR #109, opened 2019-06-19
Single pod implementation of smallfile

chawlanikhil24:


roles/smallfile-bench/tasks/main.yml
```diff
@@ -0,0 +1,93 @@
+---
+- set_fact:
+    smallfile_path: /mnt/pvc
+  when: smallfile.storageclass is defined
+
+- set_fact:
+    smallfile_path: /tmp/test_smallfile/
+  when: smallfile.storageclass is not defined
```
bengland2: why is it necessary to support the case where no storage class is defined?   If we just always required a storage class this would be simpler.   Minikube offers local storage class so there should not be a problem even when there is no Ceph.

roles/smallfile-bench/tasks/main.yml
```diff
@@ -0,0 +1,93 @@
...
+            containers:
+              - name: benchmark-server
+                image: "quay.io/nichawla/smallfile_container:testing"
+                command: ["/bin/sh", "-c"]
+                workingDir: /root/smallfile-master/
+                args:
+                  - arr=$(python <<< "print(' '.join({{smallfile.operation}}))");
+                    rm -rf {{smallfile_path}}/RESULTS ;
+                    mkdir {{smallfile_path}}/RESULTS;
+                    mkdir {{smallfile_path}}/smallfile_test_data;
```
bengland2: do you mean "mkdir -p {{smallfile_path}}/smallfile_test_data"  ?   Otherwise this will be an error, though because of the shell syntax you are using, the error will be ignored.

roles/smallfile-bench/tasks/main.yml
```diff
@@ -0,0 +1,93 @@
...
+                    for i in ${arr[@]};do echo $i;
+                    echo RUnning_OPERATION_$i;
+                    python smallfile_cli.py --operation ${i} --yaml-input-file /tmp/smallfilejob --output-json {{smallfile_path}}/RESULTS/SMALLFILE_${i}_RUN_RESULT;
+                    sleep 10;
```
bengland2: why the "sleep 10"?

playbook.yml
```diff
@@ -1,60 +1,66 @@
...
-        byowl: "{{ workload.args }}"
-      when: workload.name == "byowl" and workload.args.clients > 0
-
-    - include_role:
-        name: "fio-distributed"
-      vars:
-        fiod: "{{ workload.args }}"
-      when: workload.name == "fio_distributed" and workload.args.servers > 0
-    when: workload is defined
+    - name: Build Infrastructure
```
bengland2: why did all of this  change when most of it has nothing to do with smallfile?  Was this a whitespace/indentation problem?

roles/smallfile-bench/tasks/main.yml
chawlanikhil24: Ok, then, there is no problem to remove this.

roles/smallfile-bench/tasks/main.yml
chawlanikhil24: will add -p in mkdir

roles/smallfile-bench/tasks/main.yml
chawlanikhil24: to create a small interval before executing the next test right after the previous one. Also, I am thinking to add "echo 3 > /proc/sys/vm/drop_caches"

playbook.yml
chawlanikhil24: It was due to a merge conflict resolution

chawlanikhil24:
> Any reason why `.gitignore` file was deleted?

Now,  it wasn't looking as a required thing.

bengland2:
@jtaleric, definitely an improvement over last PR, see [results](https://pastebin.com/NzmiUeUw).   I created a cluster in AWS, installed rook/ceph on it, created a Cephfs secret and storage class, and used this with smallfile.   However, cache dropping doesn't work.     Also, no result aggregation here.   Will play with it a little more today and post results.   If it doesn't like something you input, it's a little hard to figure out why.  Suggest that you cat the YAML being input to smallfile and each smallfile_cli.py command should be logged so it is really clear what you are doing.  For example, the whole thing about converting from underscores to dashes in the parameter names will confuse people.

bengland2:
I meant @chawlanikhil24 , sorry.

bengland2:
To show you what I mean about result aggregation, here is a result from a different smallfile run where smallfile synchronizes the pods and [aggregates the results](http://pbench.perf.lab.eng.bos.redhat.com/results/e23-h03-740xd/pbench-user-benchmark_bensmf_fsz.64.op.create_2019.04.16T19.40.37/1/sample1/result.txt) for the entire cluster.   This is not possible in the ripsaw implementation because each pod only knows about its own activity.   In contrast, distributed fio does output an aggregate result.    We could postprocess results outside of smallfile, using JSON output from each pod, as shown in [output_results.py](https://github.com/distributed-system-analysis/smallfile/blob/master/output_results.py).

jtaleric:
@chawlanikhil24 nice, glad you were able to fix the merge! Can you add a smallfile test?

roles/smallfile-bench/tasks/main.yml
```diff
@@ -0,0 +1,92 @@
...
+                    cat /tmp/smallfilejob;
+                    python smallfile_cli.py --operation ${i} --yaml-input-file /tmp/smallfilejob --output-json {{smallfile_path}}/RESULTS/SMALLFILE_${i}_RUN_RESULT;
+                    sleep 5;
+                    sync; echo 3 > /proc/sys/vm/drop_caches;
+                    done;
+                    rm -rf {{ smallfile_path }}/smallfile_test_data
+                volumeMounts:
+                  - name: config-volume
+                    mountPath: "/tmp"
+
```
aakarshg: remove the empty line

roles/smallfile-bench/tasks/main.yml
```diff
@@ -0,0 +1,92 @@
...
+                  - name: ceph-volume
+                    mountPath: "{{ smallfile_path }}"
+
```
aakarshg: remove the empty line

roles/smallfile-bench/tasks/main.yml
```diff
@@ -0,0 +1,92 @@
...
+            volumes:
+              - name: config-volume
+                configMap:
+                  name: smallfile-test
+                  defaultmode: 0777
+
```
aakarshg: remove the empty line

roles/smallfile-bench/tasks/main.yml
```diff
@@ -0,0 +1,92 @@
...
+              - name: ceph-volume
+                persistentVolumeClaim:
+                  claimName: claim"{{ item }}"
+
```
aakarshg: remove the empty line

roles/smallfile-bench/tasks/main.yml
```diff
@@ -0,0 +1,92 @@
...
+            containers:
+              - name: benchmark-server
+                securityContext:
+                  runAsUser: 0 #0 means run as root
+                  privileged: true
+                image: "quay.io/nichawla/smallfile_container:testing"
+                command: ["/bin/sh", "-c"]
+                workingDir: /root/smallfile-master/
+                args:
+                  - arr=$(python <<< "print(' '.join({{smallfile.operation}}))");
```
aakarshg: just a suggestion but if the args seems to be having a lot of work, maybe you can make this a script and make it the entry point for the container?

resources/crds/ripsaw_v1alpha1_smallfile_cr.yaml
```diff
@@ -0,0 +1,17 @@
...
+spec:
+  workload:
+    name: smallfile
+    args:
+      clients: 0
+      operation: []
+      threads: 5
+      file_size: 64
+      files: 100
+      storageclass:
```
aakarshg: are these parameters supposed to be empty? that is the value null?

tests/test_smallfile.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+set -xeo pipefail
+
+source tests/common.sh
+
+function finish {
+  echo "Cleaning up Fio"
```
aakarshg: `cleaning up smallfile` ?

tests/test_crs/valid_smallfile.yaml
```diff
@@ -0,0 +1,16 @@
...
+spec:
+  workload:
+    name: smallfile
+    args:
+      clients: 1
+      operation: ["create"]
+      threads: 5
+      file_size: 64
+      files: 100
+      storageclass: rook-ceph-block # Provide if PV is needed
```
aakarshg: will fail 2/3 CI envs - minikube/minishift because the `rook-ceph-block` sc is not available.

aakarshg: so if PV isn't required we can probably remove it?

tests/test_crs/valid_smallfile.yaml
```diff
@@ -0,0 +1,16 @@
...
+kind: Benchmark
+metadata:
+  name: example-benchmark
+  namespace: ripsaw
+spec:
+  workload:
+    name: smallfile
+    args:
+      clients: 1
+      operation: ["create"]
```
aakarshg: Can you add more `operations` to test here, if possible cover all usecases.

docs/smallfile.md
```diff
@@ -0,0 +1,168 @@
...
+      clients: 1
+      operation: ["create","read","append", "delete"]
+      threads: 5
+      file_size: 64
+      files: 100
+      storageclass: rook-ceph-block # Provide if PV is needed
+      storagesize: 5Gi # Provide if PV is needed
+```
+
+NOTE: While running the sequence of tests using smallfile operator , please make sure that the initial operation must be create, and the "delete" operation should come in termination, else smallfile might produce error due to meaningless sequence of tests.
```
aakarshg: even `operation: ["read","append","create", "delete"]` wouldn't make sense right? we should cover that too in documentation.

roles/smallfile-bench/tasks/main.yml
aakarshg: @chawlanikhil24 what are your thoughts on this?

aakarshg:
can you please rebase and squash commits?

docs/smallfile.md
```diff
@@ -0,0 +1,168 @@
+# Smallfile Benchmark
+
+[Smallfile](https://github.com/distributed-system-analysis/smallfile) is a python-based distributed POSIX workload generator which can be used to quickly measure performance for a variety of metadata-intensive workloads across an entire cluster.
+
+## Running Smallfile Benchmark using Ripsaw
+Once the operator has been installed following the instructions, one needs to modify the clients parameter(which is currently set to 0), to value greater than 0 in  [cr.yaml](../resources/crds/ripsaw_v1alpha1_smallfile_cr.yaml) to run default "create" the test. Also, in addtion to that, smallfile operator is completely dependent on storageclass and storagesize. Please make sure to double check the parameters in CRD file.
```
jtaleric: name the CR so it is obvious to someone who is just reading the documents. 

s/[cr.yaml]/[ripsaw_v1alpha1_smallfile_cr.yaml]

docs/smallfile.md
jtaleric: s/smallfile operator/smallfile workload/g

docs/smallfile.md
```diff
@@ -0,0 +1,168 @@
...
+      files: 100
+      storageclass: rook-ceph-block # Provide if PV is needed
+      storagesize: 5Gi # Provide if PV is needed
+```
+
+NOTE: While running the sequence of tests using smallfile operator , please make sure that the initial operation must be create, and the "delete" operation should come in termination, else smallfile might produce error due to meaningless sequence of tests.
+
+## Adding More options in smallfile tests
+
+Smallfile also comes with a variety of configurable options for running tests, following are the parameters that can be added to CRD file:
```
jtaleric: Added to the CR file

jtaleric:
Any examples were samples > 2 (indexed)? The example provided looked like `samples:1`

roles/smallfile-bench/tasks/main.yml
chawlanikhil24: @aakarshg I tried to implement, but it was not a successful attempt. May be we can keep this for future track and meanwhile keep it here. Although, I've move the definition of the pod into template.

resources/crds/ripsaw_v1alpha1_smallfile_cr.yaml
chawlanikhil24: Fixed that , they are not supposed to be empty

tests/test_crs/valid_smallfile.yaml
chawlanikhil24: I've added the ability to run on root filesystem

docs/smallfile.md
chawlanikhil24: @aakarshg I am giving your operation as an example in the docs. 
@jtaleric  fixing that

docs/smallfile.md
chawlanikhil24: fixed

docs/smallfile.md
chawlanikhil24: fixed

docs/smallfile.md
```diff
@@ -0,0 +1,172 @@
...
+```bash
+operation: ["read","append","create", "delete"]
+#This will be meaningless sequence of test as trying to read something, which has not been created yet.The same logic applies for the append test as well. Hence, smallfile will produce error.
+```
+
+## Adding More options in smallfile tests
+
+Smallfile also comes with a variety of configurable options for running tests, following are the parameters that can be added to CR file:
+
+* response_times – if Y then save response time for each file operation in a
```
jtaleric: Small nit, but could you bold the param names so they stand out?

playbook.yml
```diff
@@ -1,63 +1,68 @@
 - hosts: localhost
   gather_facts: no
   tasks:
+    - name: Run Workload
+      block:
+        - include_role:
+            name: "uuid"
 
-  - name: Run Workload
```
jtaleric: Did the spacing change?

docs/smallfile.md
chawlanikhil24: Agree with the idea. Will add that

playbook.yml
chawlanikhil24: Yess, due to merge conflict.

roles/smallfile-bench/templates/workload_job.yml.j2
```diff
@@ -0,0 +1,55 @@
...
+apiVersion: batch/v1
+metadata:
+  name: "{{ meta.name }}-smallfile-client-{{ item }}-benchmark"
+  namespace: "{{ operator_namespace }}"
+spec:
+  ttlSecondsAfterFinished: 600
+  template:
+    metadata:
+      labels:
+        app: smallfile-benchmark
```
aakarshg: same label is applied for the job that creates the publisher pod.

README.md
```diff
@@ -5,15 +5,16 @@ a performance baseline of Kubernetes cluster on your provider.
...
-| [YCSB](docs/ycsb.md)           | Database Performance  | Working            |
-| [Byowl](docs/byowl.md)         | User defined workload | Working            |
-| [Pgbench](docs/pgbench.md)     | Postgres Performance  | Working            |
+| Workload                       | Use                    | Status in Operator |
+| ------------------------------ | ---------------------- | ------------------ |
+| [UPerf](docs/uperf.md)         | Network Performance    | Working            |
+| [Iperf3](docs/iperf3.md)       | Network Performance    | Working            |
+| FIO-lite                       | Storage IO             | Working            |
+| [Sysbench](docs/sysbench.md)   | System Performance     | Working            |
+| [YCSB](docs/ycsb.md)           | Database Performance   | Preview            |
```
aakarshg: YCSB is in working state

playbook.yml
```diff
@@ -1,54 +1,68 @@
...
+          vars:
+            sysbench: "{{ workload.args }}"
+          when: workload.name == "sysbench" and workload.args.enabled is defined and workload.args.enabled
+
+        - include_role:
+            name: "ycsb-bench"
+          vars:
+            ycsb: "{{ workload.args }}"
+            ycsb_load: "{{ workload.args.loaded | default(false) }}"
+          loop:
```
aakarshg: not a neat rebase as the loop control was changed, can you please make sure that there is no other change in here other than adding smallfile. i.e. remove the spacing change as its not necessary.

roles/smallfile-bench/templates/workload_job.yml.j2
aakarshg: still seems to be the case

tests/test_list
```diff
@@ -1,7 +1,7 @@
+test_smallfile.sh
 test_pgbench.sh
 test_fiod.sh
 test_iperf3.sh
 test_sysbench.sh
 test_uperf.sh
 test_byowl.sh
-test_ycsb.sh
```
aakarshg: This shouldn't be removed.

chawlanikhil24:
rerun

PR #110, opened 2019-06-24
Adding additional namespace specifics

mike10010100:
This allows each of these commands to be run and not error out.


aakarshg:
Thanks for the submission @mike10010100 . The scripts in tests/ directory are for CI, and when we setup CI, we switch context to ripsaw so thus this won't be necessary at all.

mike10010100:
Ah, I see. Fair enough. It was more that 90% of the scripts have this explicit `--namespace ripsaw` specification, but these did not. It also allows people to run all of the tests locally on their machines.

aakarshg:
Ah I see what you mean, that's a fair point. I've just went over the other scripts and you're right, the others have the `--namespace` but these don't. I'll go ahead and merge the PR. Thanks again for the contribution.

mike10010100:
Thank you! Love the project!

PR #113, opened 2019-06-24
Adding MongoDB infrastructure

aakarshg:
This is using statefulset and not through the enterprise operator.

### DIFF
--- a/docs/mongo.md
+++ b/docs/mongo.md
@@ -1,0 +1,54 @@
+# MongoDB
+
+[MongoDB](https://www.mongodb.com/) is a document-based, distributed database
+
+## Using the MongoDB Infra
+
+### Pre-requisites
+MongoDB uses volumeClaimTemplates to request volumes, thus you'll need to create a storageclass. Creating a storageclass
+is outside the scope of this documentation.
+
+### Customizing your CR
+
+An example to enable only the MongoDB infra (this does _not_ run a workload):
+
+```yaml
+apiVersion: ripsaw.cloudbulldozer.io/v1alpha1
+kind: Benchmark
+metadata:
+  name: mongo-infra
+  namespace: ripsaw
+spec:
+  infrastructure:
+    name: mongo
+    args:
+      servers: 3 # has to be > 0
+      storageclass: # uses the sc with the annotation storageclass.kubernetes.io/is-default-class: "true" if the option is not specified
+      storagesize: 10Gi # default value if option not specified
+      port: 27017 # default value if option not specified
+      extra_options: # list of extra options that need to be passed as is to starting mongo daemon
+        - "--smallfiles"
+        - "--noprealloc"
+```
+
+### Starting the Infra
+Once you are finished creating/editing the custom resource file and the Ripsaw benchmark operator is running, you can start the infra with:
+
+```bash
+$ kubectl apply -f /path/to/cr.yaml
+```
+
+You can then check if mongo pods are created as follows
+
+```bash
+$ kubectl get pods -l "role=mongo"
+NAME      READY   STATUS    RESTARTS   AGE
+mongo-0   2/2     Running   0          6m51s
+mongo-1   2/2     Running   0          6m48s
+mongo-2   2/2     Running   0          6m45s
+```
+
+The connection string URI is "mongodb://mongo/ycsb?replicaSet=rs0"
+
+**Note that the MongoDB role is only an infrastructure role, and no workloads will be triggered directly
+by running the CR as described here. You will need to separately define a workload in the CR (such as [YCSB](ycsb.md)).**
--- a/roles/mongo-infra/defaults/main.yml
+++ b/roles/mongo-infra/defaults/main.yml
@@ -1,0 +1,2 @@
+---
+mongo.storagesize: 10Gi
--- a/roles/mongo-infra/files/initiate_rs.py
+++ b/roles/mongo-infra/files/initiate_rs.py
@@ -1,0 +1,42 @@
+#!/bin/bash
+
+''':'
+vers=( /usr/bin/python[2-3] )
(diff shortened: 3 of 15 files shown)


resources/crds/ripsaw_v1alpha1_couchbase_cr.yaml
```diff
@@ -1,42 +1,42 @@
...
-        #default_bucket_password: "password"
-        #pull_secret_name: "rhsecret"
-        #
-        ## For OpenShift v4 w/ Built-In OLM and Marketplace
-        #catalogsource_namespace: "openshift-marketplace"
-        #cb_operator_package: "couchbase-enterprise-certified"
-        #pod_base_image: "registry.connect.redhat.com/couchbase/server"
-        #pod_version: "5.5.4-1"
-        #default_bucket_password: "password"
+apiVersion: ripsaw.cloudbulldozer.io/v1alpha1
```
aakarshg: I honestly don't know wh this even happened. Like there's no change. I even tried to redo copy-paste but this still shows as a change. I tried with and without editor too.

aakarshg:
@ekuric can you please review this work, and point out anything that I might be missing. You've more experience with running ycsb against mongodb so adding you here.

docs/mongo.md
```diff
@@ -0,0 +1,47 @@
...
+kind: Benchmark
+metadata:
+  name: mongo-infra
+  namespace: ripsaw
+spec:
+  infrastructure:
+    name: mongo
+    args:
+      servers: 3 # has to be > 0
+      storageclass: "default" # uses the default sc by default if option not specified
```
ekuric: by default there is not storageclass name with name `default`. 
When storageclass is created one can specify 
`annotations:
    storageclass.kubernetes.io/is-default-class: "true"`
what will make an storage class `preferred` storageclass ( or default ) in case there are other storage classes - then one with `is-default-class` will be used. 
In this case, it will search for storageclass with name `default` and it will not find it and then  

- fail back to storage class with `is-default-class` 
- fail if there is not any with `is-default-class` configured

I propose to put here and storage class name which is always present , eg. on ec2 it is `gp2` or leave it empty and state to update this file with working storage class from where storage will be taken.

aakarshg: Thanks for the review. I've updated it to empty and added a bit more detailed description.

resources/crds/ripsaw_v1alpha1_couchbase_cr.yaml
ekuric: there is an "space" at end, I found that having space in .yaml files can be sometimes problematic and I usually open files in `vim` and then run `:%s/\s\+$//e`  this will remove all trailing spaces in file, also check https://github.com/kubernetes/kubernetes/issues/36222

resources/crds/ripsaw_v1alpha1_ycsb-couchbase_cr.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: ripsaw.cloudbulldozer.io/v1alpha1
 kind: Benchmark
 metadata:
-  name: ycsb-couchbase-benchmark
+  name: couchbase-infra
   namespace: ripsaw
```
jtaleric: Very curious change?

roles/mongo-infra/defaults/main.yml
```diff
@@ -0,0 +1,2 @@
+---
+mongo.storagesize: 100Gi
```
jtaleric: Maybe default to something a bit more sensible? 10Gi?

roles/mongo-infra/files/initiate_rs.py
```diff
@@ -0,0 +1,42 @@
...
+        # The ismaster command is cheap and does not require auth.
+        client.admin.command('ismaster')
+    except ConnectionFailure:
+        print("Server not available")
+        return 1
+    return 0
+
+if __name__ == '__main__':
+    sys.exit(main())
+# EOF
```
jtaleric: Should this exist in SNAFU?

roles/mongo-infra/templates/mongo-sfset.yaml.j2
```diff
@@ -0,0 +1,53 @@
...
+       - name: mongo
+         image: mongo
+         command:
+           - mongod
+           - "--replSet"
+           - rs0
+           - "--bind_ip"
+           - 0.0.0.0
+           - "--smallfiles"
+           - "--noprealloc"
```
jtaleric: Any reason for this? maybe we allow the user to allow preallocation?

roles/mongo-infra/files/initiate_rs.py
aakarshg: Not really, because this is just initiating the replicaset, unfortunately ansible doesn't have a nice mongo module to do this so a simple python script

aakarshg:
@dustinblack implemented the finalizer here, because repeated mongo tests would fail as the pvs aren't deleted

roles/mongo-infra/files/initiate_rs.py
jtaleric: I am personally against maintaining python code directly in Ripsaw. Is there a mongocli or similar we can use instead?

aakarshg: unfortunately from my knowledge this is the best way forward, mongo shell interaction through ansible is painful i.e. you can start a mongo shell and then run the initiate, but there's not a single cli command to do this. This python code is only temporary in a way, until a proper mongodb module is added to ansible. Think of it as our custom module

aakarshg:
closing this out following #122 merge

aakarshg:
closing this re: #121 and this made its way into https://github.com/cloud-bulldozer/cloud-builder

PR #116, opened 2019-06-28
UPerf : Pod to Service

### LINKED ISSUE #24: RFE: Enhance uperf workload to work between ServiceIPs
Since we have the pod to pod tests in place, it would be good to have a way to do serviceIP to serviceIP tests also.

jtaleric:
UPerf uses mutliple ports, which makes things a bit tricky for using K8s
services. Since with services you need to define the ports in use.

By default UPerf will use port `20000` for data plane, but then it opens
another port random(35000,4XXXX?) for control it seems. Orgionally I
was using using Jinja to create thousands of ports to open in the
service definiton, however that caused issues with the server pod.
Instead I limited the ports that were availabile for use to
20001->20041 for the control.

fixes #24

### DIFF
--- a/roles/uperf-bench/templates/service.yml.j2
+++ b/roles/uperf-bench/templates/service.yml.j2
@@ -1,0 +1,24 @@
+---
+kind: Service
+apiVersion: v1
+metadata:
+  name: uperf-service-{{item}}
+  namespace: '{{ operator_namespace }}'
+spec:
+  selector:
+    app: uperf-bench-server-{{item}}
+  ports:
+  - name: uperf
+    port: 20000
+    targetPort: 20000
+    protocol: TCP
+{% for num in range(20001,20011,1) %}
+  - name: uperf-control-tcp-{{num}}
+    port: {{num}}
+    targetPort: {{num}}
+    protocol: TCP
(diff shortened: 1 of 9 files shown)


jtaleric:
@smalleni PTAL

jtaleric:
@smalleni Currently, this PR will create a single svc... I think we want to do 1 svc per 1pair. 

I need to add documentation, but I want to capture if the above fits the bill... If it does, I need to add a bit more logic and of course a test... @aakarshg

roles/uperf-bench/templates/service.yml.j2
```diff
@@ -0,0 +1,20 @@
...
+  name: uperf-service
+  namespace: '{{ operator_namespace }}'
+spec:
+  selector:
+    app: uperf-bench-server
+  ports:
+  - name: uperf
+    port: 20000
+    targetPort: 20000
+    protocol: TCP
```
smalleni: No need for UDP for dataplane?

smalleni:
Where is the CR for the test, can you help me find it? Shouldn't we have uperf.serviceIP defined there?

jtaleric:
@aakarshg @smalleni Ready for eyes.

PR #120, opened 2019-07-01
Iterate over test_list for all ci tests

dry923:
Iterate over tests/test_list as well as cleanup/re-deploy the deploy and operator.yaml files for each CI test. This will always run all the tests and give a summary. If any fail it will be marked with a failure.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -1,46 +1,41 @@
 #!/usr/bin/env bash
-set -xeo pipefail
+set -x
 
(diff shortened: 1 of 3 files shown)


aakarshg:
@dry923 can you also increase the value here https://github.com/cloud-bulldozer/ripsaw/blob/master/tests/common.sh#L74 to 15, helps with some tests timing out.

PR #123, opened 2019-07-02
Improve CI pod checking, including ripsaw itself

dustinblack:
This adds a `git_pod()` function to `common.sh` that is more robust than the existing `check_pods()` function. It also uses the new function to validate the ripsaw pod before allowing the test CR to be applied. All test scripts are updated to use the new function.


aakarshg:
rebased this @dustinblack, gonna wait for it to finish the run

PR #124, opened 2019-07-02
FIO multiple samples

jtaleric:
Enable FIO to run multiple samples


aakarshg:
@jtaleric closing this altogether, as this isn't what we want after the discussion we'd

PR #125, opened 2019-07-02
Stop waiting for ripsaw crd to delete when cleaning up

dry923:
Added --wait=false when deleting ripsaw crd as it was frequently hanging during CI.


dustinblack:
Is this still a problem we need to address?

aakarshg:
This can be closed, because `--wait=false` only results in a background deletion of the objects and thus doesn't address the problem directly. However given that we no longer maintain infra/other operators with in Ripsaw, I don't think this would be needed.

PR #127, opened 2019-07-03
running multiple samples with fio-d

aakarshg:
Also takes care of indexing the results from multiple samples, but doesn't do any statistics.

### DIFF
--- a/playbook.yml
+++ b/playbook.yml
@@ -50,6 +50,9 @@
 
     - include_role:
         name: "fio-distributed"
+      loop: "{{ range(0, (workload.args.samples | int)) | list | flatten(1)}}"
+      loop_control:
+        loop_var: fiod_sample
       vars:
         fiod: "{{ workload.args }}"
       when: workload.name == "fio_distributed" and workload.args.servers > 0
--- a/roles/fio-distributed/tasks/main.yml
+++ b/roles/fio-distributed/tasks/main.yml
@@ -31,7 +31,7 @@
       kind: PersistentVolumeClaim
       apiVersion: v1
       metadata:
-        name: "claim-{{item}}"
+        name: "claim-{{item}}-{{fiod_sample}}"
         namespace: '{{ operator_namespace }}'
         annotations:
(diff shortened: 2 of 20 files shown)


aakarshg:
Tested locally, looks good. results in elasticsearch - http://marquez.perf.lab.eng.rdu2.redhat.com:5601/goto/a5f5641147d5cbbc10c030a310884dba

roles/fio-distributed/tasks/main.yml
```diff
@@ -104,12 +104,19 @@
     api_version: batch/v1
     namespace: '{{ operator_namespace }}'
     label_selectors:
-      - app = fiod-client
+      - app = fiod-client-{{fiod_sample}}
```
dry923: Not to nit-pick but consistency should be -{{ fiod_sample }}

playbook.yml
```diff
@@ -50,6 +50,9 @@
 
     - include_role:
         name: "fio-distributed"
+      loop: "{{ range(0, (workload.args.samples | int)) | list | flatten(1)}}"
```
dustinblack: I don't think the loop control belongs in the `playbook.yaml`. We've used a different convention with at least uperf and pgbench, leveraging jinja2 loops in the template. I think the same thing can be applied to fio-d in the client.yaml file.

Also... Wouldn't looping here result in a separate UUID generated for each loop iteration and mess up the indexing?

aakarshg:
holding until we can come to a consensus on what `rerun`/`samples`/`iteration` mean. This PR assumes `samples` mean rebuilding/configuring infra needed to run the benchmark, instead of just running the benchmark.

aakarshg:
This is temporarily put on hold, as this PR doesn't just run fio `n` times but rather recreates the configuration and then runs fio. #137 achieves the `samples` aspect, and if theres any need for the `rerun` then we shall focus on this PR.

dustinblack:
IMO we should close this one.

dustinblack:
@aakarshg Can we close this out? I don't think there is code here that we'll end up using.

aakarshg:
Closing this, given that theres been no ask.

PR #132, opened 2019-07-05
selectively running tests

aakarshg:
To ensure that a specific test is ran only if the roles directory is changed.
And run all the tests if higher level files/dirs are changed, which are stored
in tests/full_test_file_trigger

### DIFF
--- a/tests/common.sh
+++ b/tests/common.sh
@@ -1,5 +1,27 @@
 #!/usr/bin/env bash
 
+function check_full_trigger {
+python - <<END
+import os
+with open('tests/full_test_file_trigger', 'r') as file:
+    a = file.read()
+a = filter(None,a.split('\n'))
+with open('tests/git_diff', 'r') as file:
+    b = file.read()
+b = filter(None,b.split('\n'))
+print(any((x in a for x in b)))
+END
+}
+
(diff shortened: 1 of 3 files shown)


tests/common.sh
```diff
@@ -1,5 +1,27 @@
...
+END
+}
+
+function populate_test_list {
+  if 'roles/uperf-bench' in $1; then echo "test_uperf.sh" >> tests/iterate_tests; fi
+  if 'roles/fio-distributed' in $1; then echo "test_fiod.sh" >> tests/iterate_tests; fi
+  if 'roles/fio-bench' in $1; then echo "test_fio.sh" >> tests/iterate_tests; fi
+  if 'roles/iperf3-bench' in $1; then echo "test_iperf.sh" >> tests/iterate_tests; fi
+  if 'roles/byowl' in $1; then echo "test_byowl.sh" >> tests/iterate_tests; fi
+  if 'roles/sysbench' in $1; then echo "test_sysbench.sh" >> tests/iterate_tests; fi
```
dustinblack: I'd like to see a better way to maintain this than having to add a line here for each new role, which seems prone to error. But I can live with this for now.

PR #137, opened 2019-07-05
Fio-distributed running multiple samples

aakarshg:
runnning/indexing with 2 samples - http://marquez.perf.lab.eng.rdu2.redhat.com:5601/goto/855d7859a84278374ed87c5cbe124076

### DIFF
--- a/docs/fio_distributed.md
+++ b/docs/fio_distributed.md
@@ -20,6 +20,7 @@ spec:
     name: "fio_distributed"
     args:
       servers: 1
+      samples: 1 # number of times to run the fio test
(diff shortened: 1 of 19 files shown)


chawlanikhil24:
@aakarshg samples parameter is added only in docs, and test_crs, I think , its good to add the parameter in [crd]( resources/crds/ripsaw_v1alpha1_fio_distributed_cr.yaml)  and set the default value to 1, like the way we do for clients.

aakarshg:
nice catch @chawlanikhil24 , updated the crd.

PR #138, opened 2019-07-06
Changing ycsb status to working

aakarshg:


docs/ycsb.md
```diff
@@ -16,7 +14,7 @@ Your resource file may look like this:
 apiVersion: ripsaw.cloudbulldozer.io/v1alpha1
 kind: Benchmark
 metadata:
-  name: ycsb-couchbase-benchmark
+  name: ycsb-mongo-benchmark
```
jtaleric: Should you rename this to `docs/ycsb-mongo.md` to be specific that this is for mongodb?

tests/test_list
```diff
@@ -4,3 +4,4 @@ test_iperf3.sh
 test_sysbench.sh
 test_uperf.sh
 test_byowl.sh
+test_ycsb.sh
```
jtaleric: Should the testname reflect the database?

docs/ycsb.md
aakarshg: Only the example here is specifically for running against mongodb, so I dont think so.

tests/test_list
aakarshg: We're not testing the database at all here, all the test tries to see if ycsb can actually run. For sanity purposes it'll load the database.

docs/ycsb.md
jtaleric: The remove mongo?

PR #142, opened 2019-07-08
Remove fio

jtaleric:


aakarshg:
also update https://github.com/cloud-bulldozer/ripsaw/blob/master/docs/installation.md#running-workload for fio to point at fio-d instead.

aakarshg:
looks like I missed another file that would need to be deleted in the [crd directory](https://github.com/cloud-bulldozer/ripsaw/blob/master/resources/crds/ripsaw_v1alpha1_fio_cr.yaml) . that should be it, and if we find anything else still present, we can deal with it in a later PR.

Issue #144, opened 2019-07-08
[pgbench] pin_node is NOT optional in ripsaw_v1alpha1_pgbench_cr.yaml


netzzer:
Leaving pin-node blank caused the client pod to not create. Adding a value such as `pin_node: "ip-10-0-134-143"` in ripsaw_v1alpha1_pgbench_cr.yaml fixed the problem and the test ran as expected after this change.

```
$ oc get benchmark
NAME                TYPE      AGE
pgbench-benchmark   pgbench   56s
[ec2-user@ip-172-31-14-128 ripsaw-dustin]$ oc describe benchmark pgbench-benchmark
Name:         pgbench-benchmark
Namespace:    ripsaw
Labels:       <none>
Annotations:  <none>
API Version:  ripsaw.cloudbulldozer.io/v1alpha1
Kind:         Benchmark
Metadata:
  Creation Timestamp:  2019-07-08T23:39:49Z
  Generation:          1
  Resource Version:    91783
  Self Link:           /apis/ripsaw.cloudbulldozer.io/v1alpha1/namespaces/ripsaw/benchmarks/pgbench-benchmark
  UID:                 ad4debce-a1d9-11e9-acf0-0268146ce15c
Spec:
  Workload:
    Args:
      Clients:
        4
        8
      cmd_flags:  
      Databases:
        db_name:       sampledb
        Host:          172.30.86.214
        Password:      wTRfg5vxpmtfkYKA
        pin_node:      <nil>
        Port:          <nil>
        User:          user8LH
      init_cmd_flags:  
      run_time:        300
      Samples:         2
      scaling_factor:  30
      Threads:         4
      Timeout:         5
      Transactions:    <nil>
    Name:              pgbench
Status:
  Conditions:
    Last Transition Time:  2019-07-08T23:40:53Z
    Message:               Running reconciliation
    Reason:                Running
    Status:                False
    Type:                  Running
    Ansible Result:
      Changed:             2
      Completion:          2019-07-08T23:40:56.93001
      Failures:            1
      Ok:                  5
      Skipped:             6
    Last Transition Time:  2019-07-08T23:40:57Z
    Message:               An unhandled exception occurred while running the lookup plugin 'template'. Error was a <class 'ansible.errors.AnsibleError'>, original message: Unexpected templating type error occurred on (---
kind: Job
apiVersion: batch/v1
metadata:
  name: '{{ meta.name }}-pgbench-client-{{ item.0|int + 1 }}'
[...]

PR #145, opened 2019-07-09
fix jinja2 if logic for pin_node feature of pgbench workload

dustinblack:
Fix for issue #144

### DIFF
--- a/roles/pgbench/templates/workload.yml.j2
+++ b/roles/pgbench/templates/workload.yml.j2
@@ -41,7 +41,7 @@ spec:
                done;
              fi"
       restartPolicy: OnFailure
-{% if item.1.pin_node is defined and item.1.pin_node|length and item.1.pin_node is not sameas "" %}
+{% if item.1.pin_node is defined and item.1.pin_node %}
       nodeSelector:
         kubernetes.io/hostname: '{{ item.1.pin_node }}'
 {% endif %}


roles/pgbench/templates/workload.yml.j2
```diff
@@ -41,7 +41,7 @@ spec:
                done;
              fi"
       restartPolicy: OnFailure
-{% if item.1.pin_node is defined and item.1.pin_node|length and item.1.pin_node is not sameas "" %}
+{% if item.1.pin_node is defined and item.1.pin_node %}
```
jtaleric: `item.1.pin_node` covers both the empty and none case? 

Should we add a test to our CI to catch/cover this?

dry923: Assuming item.1.pin_node covers both cases that's good. Otherwise would the bellow suffice? It would vet there is some length to the pin node, hence not being blank.

```
{% if item.1.pin_node is defined and item.1.pin_node|length > 0 %}
```

dustinblack: > `item.1.pin_node` covers both the empty and none case?

Yes. Per the docs, you should also be able to cover the undefined case with simply '{% if item.1.pin_node %}`, but in testing it this failed for me. The if logic in this patch was tested against the "full" list of scenarios and passed, as documented in issue #144 .

> Should we add a test to our CI to catch/cover this?

I can't really think of how we would do this. Suggestions?

aakarshg: If this change covers the list of scenarios you've documented in the issue, then I don't think we'll need another test in our CI.

jtaleric: @aakarshg you don't think any of our other CRs will suffer from this exact same issue? I tend to think they will.

PR #149, opened 2019-07-11
Allow user provide more detail to YCSB

jtaleric:
This allows the user to modify the YCSB workload a bit.

### DIFF
--- a/roles/load-ycsb/defaults/main.yml
+++ b/roles/load-ycsb/defaults/main.yml
@@ -1,0 +1,5 @@
+---
+user: ""
+elasticsearch:
+        server: ""
+        port: ""
--- a/roles/load-ycsb/tasks/main.yml
+++ b/roles/load-ycsb/tasks/main.yml
@@ -1,27 +1,27 @@
 ---
 - name: setting ycsb_workload var
   set_fact:
-    ycsb_workload: "{{ item }}"
+    ycsb_workload: "workloada"
 
 - name: Generate workloads
   k8s:
     definition: "{{ lookup('template', 'configmap.yml.j2') | from_yaml }}"
 
-- name: Run YCSB Workload
+- name: Load Data Into Database
   k8s:
     definition:
       kind: Job
       apiVersion: batch/v1
       metadata:
-        name: '{{ meta.name }}-ycsb-bench-job-{{ ycsb_workload }}'
+        name: '{{ meta.name }}-ycsb-data-load-job-{{ ycsb_workload }}'
         namespace: '{{ operator_namespace }}'
       spec:
         template:
           metadata:
-            name: '{{ meta.name }}-ycsb-bench'
+            name: '{{ meta.name }}-ycsb-data-load'
             namespace: '{{ operator_namespace }}'
             labels:
-              name: 'ycsb-run'
+             name: 'ycsb-load'
           spec:
             containers:
             - name: ycsb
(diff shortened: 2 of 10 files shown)


roles/load-ycsb/tasks/main.yml
```diff
@@ -0,0 +1,66 @@
...
+        template:
+          metadata:
+            name: '{{ meta.name }}-ycsb-data-load'
+            namespace: '{{ operator_namespace }}'
+            labels:
+             name: 'ycsb-load'
+          spec:
+            containers:
+            - name: ycsb
+              image: "quay.io/jtaleric/ycsb-server:testing"
```
dry923: Should the image location be in quay/benchmark-operator?

roles/ycsb-bench/tasks/main.yml
```diff
@@ -66,11 +25,33 @@
           spec:
             containers:
             - name: ycsb
-              image: "quay.io/cloud-bulldozer/ycsb-server:latest"
+              image: "quay.io/jtaleric/ycsb-server:testing"
```
dry923: Same as above w/ re the image location.

dry923:
This is failing in kni and minishift because it is trying to write to /data/db which would be a privileged location. Changing the mongod default db path to tmp fixes this in my test on kni:

 args:  ["-c", "mkdir -p /tmp/data/db; mongod --smallfiles --bind_ip 0.0.0.0 --dbpath /tmp/data/db"]

roles/load-ycsb/defaults/main.yml
```diff
@@ -0,0 +1,5 @@
+---
+user: ""
+elasticsearch:
+        server: ""
+        port: ""
```
dustinblack: I've been struggling to get defaults to work the way I expect -- applying for missing, empty, or blank values. This method only takes care of missing, but empty or blank will be passed on as blank. Not a blocker problem at all, but something that could reduce user error if we can fix it.

roles/load-ycsb/tasks/main.yml
```diff
@@ -0,0 +1,66 @@
...
+              args: ["-c", "/opt/snafu/ycsb-wrapper/ycsb-wrapper.py -r 1 -l -d {{ ycsb.driver }} -w {{ ycsb_workload }} -x \"{{ ycsb.options_load }}\""]
+              volumeMounts:
+                - name: config-volume
+                  mountPath: "/tmp/ycsb"
+            restartPolicy: Never
+            volumes:
+            - name: config-volume
+              configMap:
+                name: ycsb-test
+        backoffLimit: 4
```
dustinblack: Do you want to pull this job out into a template to clean up the task file a bit?

roles/load-ycsb/templates/configmap.yml.j2
```diff
@@ -0,0 +1,245 @@
...
+    recordcount={{ycsb.recordcount}}
+    operationcount={{ycsb.operationcount}}
+    workload=com.yahoo.ycsb.workloads.CoreWorkload
+    readallfields=true
+    readproportion=0.5
+    updateproportion=0
+    scanproportion=0
+    insertproportion=0
+    readmodifywriteproportion=0.5
+    requestdistribution=zipfian
```
dustinblack: You might take a look at the recent work I've done for fio that allows for overriding config defaults via the CR file. Something similar here might provide even more flexibility for the power user.

roles/ycsb-bench/tasks/main.yml
```diff
@@ -67,10 +26,32 @@
...
-              args: ["-c", "/ycsb/bin/ycsb run {{ ycsb.driver }} -s -P /ycsb/workloads/{{ ycsb_workload }} {{ ycsb.options_load }}"]
+              args: ["-c", "/opt/snafu/ycsb-wrapper/ycsb-wrapper.py -r 1 -d {{ ycsb.driver }} -w {{ ycsb_workload }} -x \"{{ ycsb.options_load }}\""]
+              volumeMounts:
+                - name: config-volume
+                  mountPath: "/tmp/ycsb"
             restartPolicy: Never
+            volumes:
+              - name: config-volume
+                configMap:
+                  name: ycsb-test
```
dustinblack: Similar comment about pulling this out to a template file.

roles/load-ycsb/defaults/main.yml
jtaleric: Ah, so this is due to me moving to passing the env vars via the container yaml. If the user does not pass the `elasticsearch` var ansible will throw up. I had to create these defaults... However, the SNAFU code should ignore empty value server/port.

roles/load-ycsb/tasks/main.yml
jtaleric: I thought about this. It might not be a bad idea.

roles/load-ycsb/templates/configmap.yml.j2
jtaleric: Correct, I just got a email from Elko and he mentioned that we should look at changing this to uniform. I should template this out too...

PR #152, opened 2019-07-12
integrate snafu wrapper of pgbench for elastic indexing

dustinblack:
Enhances pgbench to execute via snafu wrapper, allowing indexing to elastic.

### DIFF
--- a/docs/pgbench.md
+++ b/docs/pgbench.md
@@ -69,3 +69,72 @@ Once done creating/editing the resource file, you can run it by:
 ```bash
 # kubectl apply -f <path_to_cr_file>
 ```
+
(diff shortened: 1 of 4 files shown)


dustinblack:
Minikube failed on what looks like an operator problem. Doesn't make a lot of sense to me.

`error: unable to recognize "STDIN": no matches for kind "Benchmark" in version "ripsaw.cloudbulldozer.io/v1alpha1"`

Minishift failed waiting on the pgbench client to be ready.

KNI passed, so per our recent agreement the Mini* failures should not be blockers.

dustinblack:
I think minikube and minishift are just timing out on pulling the updated image from quay. Will rerun again.

PR #153, opened 2019-07-14
updating namespace to my-ripsaw

aakarshg:
Ripsaw can now also be installed through operatorhub.io, in which case
all the resources would be in my-ripsaw namespace. To avoid confusion, updating
all docs to refer to my-ripsaw namespace. Thus the docs/instruction would remain same
irrespective of how ripsaw is deployed.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -46,6 +46,9 @@ in the installation guide.
 Note: The benchmark-operator's code-name is ripsaw, so the names have been
 used interchangeably.
 
+Note: If you're on a vanilla k8s distribution, then you can also deploy Ripsaw through
+      operatorhub.io, please check [ripsaw in operatorhub](https://operatorhub.io/operator/ripsaw) for more details.
+
 First we'll need to clone the operator:
 
 ```bash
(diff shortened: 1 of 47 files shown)


docs/installation.md
```diff
@@ -55,15 +58,17 @@ First we'll need to clone the operator:
 ```
 
 We try to maintain all resources created/required by ripsaw in the namespace
```
jtaleric: Can you add that we did this to support introduction into OperatorHub so people don't think we randomly picked it.

tests/full_test_file_trigger
```diff
@@ -1,8 +1,11 @@
...
-resources/
-tests/
+build/Dockerfile
+deploy/10_service_account.yaml
+deploy/20_role.yaml
+deploy/30_role_binding.yaml
+deploy/40_cluster_role_kubevirt.yaml
+group_vars/all.yml
+resources/operator.yaml
+tests/common.sh
```
dustinblack: Were the directory-based paths broken here? This seems like a strange addition to this commit.

PR #154, opened 2019-07-16
removing pv req for indexing

aakarshg:
Indexing fio documents/logs requires for storage to be created with a vol plugin
that supports `RWX`. This is not possible in all environments, so moving to an indexing
in a similar fashion to uperf.

Commit also does following
1. Cleanup fio-d
2. Remove any files that were related to `RWX` pv


roles/fio-distributed/templates/client.yaml
```diff
@@ -18,8 +18,20 @@ spec:
...
+             export es_index={{elasticsearch.index}}
+{% endif %}
+{% endif %}
+{% if test_user is defined %}
+             export test_user={{test_user}};
+{% endif %}
              for i in `seq 1 {{fiod.samples}}`; do
-               cd /tmp && fio --client=/tmp/host/hosts /tmp/fio/fiojob;
+               mkdir -p /tmp/fiod-{{ uuid }}/$i;
+               python /opt/snafu/fio-wrapper/fio-wrapper.py /tmp/host/hosts /tmp/fio/fiojob -s $i -d /tmp/fiod-{{ uuid }}/$i;
```
jtaleric: This might change with the SNAFU change.

roles/fio-distributed/templates/job.fio.rand
```diff
@@ -10,7 +10,6 @@ time_based=1
 runtime={{fiod.runtime}}
 clocksource=clock_gettime
 ramp_time={{fiod.ramp_time}}
-stonewall
```
dustinblack: Why are we removing stonewall from the jobfiles?

Side note: This only needs to be defined once in the `[global]` section; it is redundant to repeat it in the individual job sections.

aakarshg:
have replies to certain concerns inline.

> **Splitting jobfiles into single jobs instead of using stonewall**
> I'm actually fine with this approach in concept, as it is how I have always run fio in my own wrapper scripts in the past. However, I don't think that it is the correct approach to take a combined jobfile, provided as a ripsaw role template, and then split it via python loop logic in the snafu wrapper script. Doing this confuses the context of the job from the view of an experienced fio user who may look into or even modify our jobfile templates in ripsaw. As an fio user, I would look at the combined jobfile and expect it to run as combined, so I would immediately question why `stonewall` is not being used. I wouldn't be able to resolve that question without digging into the wrapper code in snafu.
> 

The splitting jobs through python script, can be toggled by adding another parameter `no_split` to the cr. I can add this functionality to avoid confusion. That way if the user wants the jobs to not be split, they can still achieve that.


> **The right place to split the jobfiles**
> IMO, what we should be doing instead is splitting the jobfiles in the ripsaw templates. This to me is much simpler than maintaining the `_parse_jobs` and `_build_fio_job` python functions as currently written in the snafu [wrapper](https://github.com/cloud-bulldozer/snafu/blob/0da1bf70b4d13dde31f7e786013a3e3b983ea204/fio-wrapper/fio-wrapper.py). This also provides the added advantage of better user control of the jobs to be run by allowing us to adjust the CR such that the `job` is a list of built-in jobfiles to loop through, and this also better paves the way for the user to provide custom jobfiles.

`simpler` is debatable, because imho it'll make things harder to maintain. Ansible isn't a programming language, so the framework would've less control on running the jobs. My concern is we'll probably hit the same issues we've seen with indexing work and ultimately decide to move even that logic into snafu.

> 
> _User story: As a user, I want to run only the sequential write and random write fio jobs._
> 
> With the current implementation, that ^^ is not possible.
> 

This can be easily addressed by allowing custom job files, which is a bit out of scope of this PR.

> **What snafu should and shouldn't do**
> I believe that the snafu wrapper scripts should:
> 
> 1. Normalize the data output for the CDM
> 2. Provide the logic for uploading the data to the remote store (Elasticsearch or other)
> 
> The logic to split apart the combined fio jobfile I think is an overload of snafu. I think the context of _how_ to run the job should remain within ripsaw. Perhaps that could still be passed to snafu for better logic handling via python when needed (spitballing... maybe a python ansible module would make even more sense?), but a snafu wrapper shouldn't fundamentally change the way a job operates, as that may be counter to the user expectations set within ripsaw.
> 

I think ultimately everything boils down to this `What snafu should and shouldn't do?` . I'm of the opinion that snafu is a component, with input of following
1. benchmark type, in this case `fio`
2. benchmark metadata, in this case `job file`, `hosts`, `sample` and other parameters such as `es` details.
3. certain optional metadata related to the environment i.e. in case of kubernetes run ( not implemented yet ) - `nodes` on which the `pods` landed.

The output of snafu is :
1. structured benchmark data.
2. optionally index to elasticsearch/other_data_store.

If we feel that snafu should only do following:
> 1. Normalize the data output for the CDM
> 2. Provide the logic for uploading the data to the remote store (Elasticsearch or other)

Then we'll need the input to snafu to be:
1. the stdout from the benchmark run. 

The output will still be:
1. structured benchmark data.
2. optionally index to elasticsearch/other_data_store.

So if we move forward with the above, then snafu shouldn't run the benchmark as well and ripsaw would still run benchmark and just pass the output.

I'm okay with either way, but we need to define the interfaces well. So ultimately we need to decide the input/output of the component ( snafu ) and its interfaces with other components ( ripsaw ).

bengland2:
@dustinblack , I won't comment on "snafu" since I'm not familiar with it.
I think @aakarshgopi  proposal to run a separate fio --client per fio job
is much more in line with how I and others (e.g. CBT, pbench-fio) have used
it in the past.    It also means there is less uncertainty on how to
timestamp the data from the fio logs, since we know that the job starts
when the fio --client starts (+ a couple of sec for transmitting job
files).    There will always be some uncertainty in a distributed workload
- this doesn't matter much as long as the fio threads are almost entirely
running in the same time interval (say 95%).     As long as fio --client
starts them at approximately the same time +- a few seconds, they should
stop at approximately the same time also because they all have the same
"runtime" parameter, assuming nodes are time-synced.

 It's important for timestamps of fio results to be reasonably accurate so
that we can correlate with data collected by prometheus/pbench/whatever on
h/w resource utilization, etc.  as Alex Calhoun was saying.      I'm not
hung up on the output format from ripsaw, but some consistency in it would
make it easier to convert benchmark results to elastic search documents and
correlate it to time-series data from other sources (prometheus, pbench,
etc).

dustinblack:
> have replies to certain concerns inline.
> 
> > **Splitting jobfiles into single jobs instead of using stonewall**
> > I'm actually fine with this approach in concept, as it is how I have always run fio in my own wrapper scripts in the past. However, I don't think that it is the correct approach to take a combined jobfile, provided as a ripsaw role template, and then split it via python loop logic in the snafu wrapper script. Doing this confuses the context of the job from the view of an experienced fio user who may look into or even modify our jobfile templates in ripsaw. As an fio user, I would look at the combined jobfile and expect it to run as combined, so I would immediately question why `stonewall` is not being used. I wouldn't be able to resolve that question without digging into the wrapper code in snafu.
> 
> The splitting jobs through python script, can be toggled by adding another parameter `no_split` to the cr. I can add this functionality to avoid confusion. That way if the user wants the jobs to not be split, they can still achieve that.

I am very against this. You are taking a function in the script that I already think is overloaded and adding more logic to it. My argument is that the better approach is to take the logic away from snafu and put it in ripsaw where it is closer to the user context for the benchmark itself.

> 
> > **The right place to split the jobfiles**
> > IMO, what we should be doing instead is splitting the jobfiles in the ripsaw templates. This to me is much simpler than maintaining the `_parse_jobs` and `_build_fio_job` python functions as currently written in the snafu [wrapper](https://github.com/cloud-bulldozer/snafu/blob/0da1bf70b4d13dde31f7e786013a3e3b983ea204/fio-wrapper/fio-wrapper.py). This also provides the added advantage of better user control of the jobs to be run by allowing us to adjust the CR such that the `job` is a list of built-in jobfiles to loop through, and this also better paves the way for the user to provide custom jobfiles.
> 
> `simpler` is debatable, because imho it'll make things harder to maintain. Ansible isn't a programming language, so the framework would've less control on running the jobs. My concern is we'll probably hit the same issues we've seen with indexing work and ultimately decide to move even that logic into snafu.

We are talking very basic jinja2 loop logic, here. Nothing earth-shattering.

You'll always have the point that Ansible is a bad place to be doing logic in the first place. I get that. But I don't think the snafu wrapper is the right place to address this. We're not re-writing the operator in Golang any time soon. However, if we are overloading Ansible too much, we should think about new Ansible modules, I think. (IMO that's overkill by a lot for this particular concern.)

> 
> > _User story: As a user, I want to run only the sequential write and random write fio jobs._
> > With the current implementation, that ^^ is not possible.
> 
> This can be easily addressed by allowing custom job files, which is a bit out of scope of this PR.

I think it will stay out of scope for longer with your current proposed approach. By splitting the jobfiles in ripsaw, we have a much stronger base for getting custom jobfiles into place.

> 
> > **What snafu should and shouldn't do**
> > I believe that the snafu wrapper scripts should:
> > 
> > 1. Normalize the data output for the CDM
> > 2. Provide the logic for uploading the data to the remote store (Elasticsearch or other)
> > 
> > The logic to split apart the combined fio jobfile I think is an overload of snafu. I think the context of _how_ to run the job should remain within ripsaw. Perhaps that could still be passed to snafu for better logic handling via python when needed (spitballing... maybe a python ansible module would make even more sense?), but a snafu wrapper shouldn't fundamentally change the way a job operates, as that may be counter to the user expectations set within ripsaw.
> 
> I think ultimately everything boils down to this `What snafu should and shouldn't do?` . I'm of the opinion that snafu is a component, with input of following
> 
> 1. benchmark type, in this case `fio`
> 2. benchmark metadata, in this case `job file`, `hosts`, `sample` and other parameters such as `es` details.
> 3. certain optional metadata related to the environment i.e. in case of kubernetes run ( not implemented yet ) - `nodes` on which the `pods` landed.
> 
> The output of snafu is :
> 
> 1. structured benchmark data.
> 2. optionally index to elasticsearch/other_data_store.
> 
> If we feel that snafu should only do following:
> 
> > 1. Normalize the data output for the CDM
> > 2. Provide the logic for uploading the data to the remote store (Elasticsearch or other)
> 
> Then we'll need the input to snafu to be:
> 
> 1. the stdout from the benchmark run.
> 
> The output will still be:
> 
> 1. structured benchmark data.
> 2. optionally index to elasticsearch/other_data_store.
> 
> So if we move forward with the above, then snafu shouldn't run the benchmark as well and ripsaw would still run benchmark and just pass the output.
> 
> I'm okay with either way, but we need to define the interfaces well. So ultimately we need to decide the input/output of the component ( snafu ) and its interfaces with other components ( ripsaw ).

This comes back to conversations we've had around your interest in making snafu more consumable outside of ripsaw. I just don't think that's anywhere on our roadmap or within our needs. Stick to the two things I proposed it should do, and leave the snafu scripts as external utilities for ripsaw.

I don't think there is any reason to change fundamentally the way the sanfu scripts work, as the model has proven itself already with uperf and pgbench. It doesn't need to change to a post-processing tool, as you suggest. Using the scripts as wrappers has other advantages.

aakarshg:
I see your concern about modifying how the workload runs by breaking it apart into multiple runs, so we can simply suggest that user should pass a job file with a single job in it, and if they pass a single job then we can index the logs / result jsons as is. However if the user passes a job file with multiple fio jobs, we just skip indexing logs as we can't accurately measure the start time. Do you think this is acceptable @bengland2 , given that most of the times you'd run fio with a single job in the jobfile?

aakarshg:
And just to be clear, i'm okay to modify the code according to what @bengland2 and @dustinblack think is the best.

bengland2:
Proposed goal:  invoke ripsaw, once, and get it to run a series of fio job files and record the fio test result files for that, in such a way that this data can be indexed into elastic and visualized.   That implies we should be able to convert relative timestamps in fio logs to absolute timestamps.   I think Aakarsh's solution of running separate fio --client run per job would have done that, since the stat() st_mtime would then be (approximately) the end time for the job, and since we know the job duration (runtime=), we can compute the start time and rebase the timestamps in the log on that.   Not as accurate as having fio **tell us** the start time, but good enough.  

Why does this conflict with Dustin's idea?   Dustin is saying "The logic to split apart the combined fio jobfile I think is an overload of snafu".   But I'm not asking for that.   If ripsaw runs a series of fio job files as separate fio runs, then the only requirement on snafu is to process a series of runs, rather than just one - it doesn't require any additional knowledge of fio to process a series of runs.   I think there isn't really a conflict here.

dustinblack:
> However if the user passes a job file with multiple fio jobs, we just skip indexing logs as we can't accurately measure the start time, as well as the visualizations wouldn't make much sense as there'd be an overlap of the jobs. Do you think this is acceptable @bengland2 , given that most of the times you'd run fio with a single job in the jobfile?

This is a question that can be deferred since it only affects custom jobfiles and the initial implementation will only support our built-in jobfiles. This will allow us more time to better address the questions of timestamps, stonewall behavior, and overlapping results. We could even phase in the support for custom jobfiles, first supporting only files with single jobs and later adding support for combined files once we better understand the behavior.

dustinblack:
> Proposed goal: invoke ripsaw, once, and get it to run a series of fio job files and record the fio test result files for that, in such a way that this data can be indexed into elastic and visualized.

+1, though I would augment this to "run a series of jobfiles of the user's choosing". I should be able to pass a specific list of jobs I want to run.

> That implies we should be able to convert relative timestamps in fio logs to absolute timestamps. I think Aakarsh's solution of running separate fio --client run per job would have done that, since the stat() st_mtime would then be (approximately) the end time for the job, and since we know the job duration (runtime=), we can compute the start time and rebase the timestamps in the log on that. Not as accurate as having fio **tell us** the start time, but good enough.

My proposal doesn't change this part of the operation; individual fio runs per job will still be orchestrated by the snafu wrapper script, implying that $whatever needs to be done to normalize the timestamp data still applies. That, to me, is within the purview of snafu.

> 
> Why does this conflict with Dustin's idea? Dustin is saying "The logic to split apart the combined fio jobfile I think is an overload of snafu". But I'm not asking for that. If ripsaw runs a series of fio job files as separate fio runs, then the only requirement on snafu is to process a series of runs, rather than just one - it doesn't require any additional knowledge of fio to process a series of runs. I think there isn't really a conflict here.

+1

PR #156, opened 2019-07-19
Implement looping through a list of total number of databases

dustinblack:
based on a pattern supplied in the CR file

This allows us an outer loop that will ramp up the number of databases we are testing against based on a pattern option supplied in the CR file.

### DIFF
--- a/roles/pgbench/tasks/run_workload.yml
+++ b/roles/pgbench/tasks/run_workload.yml
@@ -1,0 +1,35 @@
+---
+- name: Init workload start signal to false
+  command: "redis-cli set pgb_start false;"
+
+- name: Init client ready state to empty
+  command: "redis-cli del pgb_client_ready"
+
+- name: Setup pgbench test job(s)
+  k8s:
+    definition: "{{ lookup('template', 'templates/workload.yml.j2') }}"
+  with_indexed_items: "{{ pgbench.databases }}"
+  when: item.0|int in range(dbnum)
+
+- name: Wait for all clients to be ready
+  shell: "redis-cli --raw llen pgb_client_ready"
+  register: client_ready
+  until: client_ready.stdout|int == dbnum|int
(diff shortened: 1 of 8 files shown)


roles/pgbench/tasks/run_workload.yml
```diff
@@ -0,0 +1,36 @@
...
+- name: Setup pgbench test job(s)
+  k8s:
+    definition: "{{ lookup('template', 'templates/workload.yml.j2') }}"
+  with_indexed_items: "{{ pgbench.databases }}"
+  when: item.0|int in range(dbnum)
+
+- name: Wait for all clients to be ready
+  shell: "redis-cli --raw llen pgb_client_ready"
+  register: client_ready
+  #until: "client_ready.stdout|int == pgbench.databases|length"
```
dry923: Can we nix this line? Doesn't seem needed.

tests/test_crs/valid_pgbench.yaml
```diff
@@ -17,6 +17,7 @@ spec:
       init_cmd_flags: ''
       scaling_factor: 1
       samples: 2
+      num_databases_pattern: 'all'
```
aakarshg: not a blocker but highly suggest that for the testing we can include add_1 or add_2 patterns and test against multiple db if possible. This will increase coverage.

roles/pgbench/tasks/run_workload.yml
dustinblack: Whoops. Testing stuff that didn't get cleaned up.

tests/test_crs/valid_pgbench.yaml
dustinblack: +1

dustinblack: I played with this briefly, but I think it will be better done in a later commit.

PR #157, opened 2019-07-19
Custom job URL for FIO benchmarks

rsevilla87:
Fixes https://github.com/cloud-bulldozer/ripsaw/issues/143

### DIFF
--- a/roles/fio-distributed/tasks/main.yml
+++ b/roles/fio-distributed/tasks/main.yml
@@ -9,9 +9,37 @@
     fio_path: /tmp
   when: fiod.storageclass is not defined
 
+- name: Custom fio job
+  block:
+
+  - name: Download custom fio job
+    get_url:
+      url: "{{ fiod.job_url }}"
+      validate_certs: "{{ fiod.job_validate_certs }}"
+      dest: /tmp/fio.job
+
+  - name: Generate fio test from custom job
+    k8s:
+      definition:
+        apiVersion: v1
+        kind: ConfigMap
+        metadata:
+          name: fio-test
+          namespace: '{{ operator_namespace }}'
+        data:
+          fiojob: "{{ lookup('file', '/tmp/fio.job') }}"
+
+  - name: Delete fio job file
+    file:
+      path: /tmp/fio.job
+      state: absent
+
+  when: fiod.job_url is defined
+
 - name: Generate fio test
   k8s:
     definition: "{{ lookup('template', 'configmap.yml.j2') | from_yaml }}"
+  when: fiod.job_url is not defined
 
 - name: Create PVC(s)
   k8s:
--- a/docs/fio_distributed.md
+++ b/docs/fio_distributed.md
@@ -48,6 +48,29 @@ Additionally, fio distributed will default to numjobs:1, and this current cannot
 
 (*Technical Note*: If you are running kube/openshift on VMs make sure the diskimage or volume is preallocated.)
 
+It's also possible to fetch a custom FIO job from a URL.
+
+```yaml
+apiVersion: ripsaw.cloudbulldozer.io/v1alpha1
+kind: Benchmark
+metadata:
+  name: fio-benchmark
+  namespace: my-ripsaw
+spec:
+  workload:
+    name: "fio_distributed"
+    args:
+      servers: 1
+      samples: 1 # number of times to run the fio test
+      pin: false
+      pin_server: "master-0"
+      storageclass: rook-ceph-block
+      storagesize: 5Gi
+      job_url: https://example.com/job.fio
+      job_validate_certs: true
+```
+
+
 ## Indexing in elasticsearch and visualization through Grafana ( Experimental )
 
 ### Setup of Elasticsearch and Grafana


roles/fio-distributed/tasks/main.yml
```diff
@@ -17,6 +17,30 @@
     fio_path: /tmp
   when: fiod.storageclass is not defined
 
+- name: Download custom fio job
+  get_url:
+    url: "{{ fiod.job_url }}"
+    validate_certs: "{{ fiod.job_validate_certs }}"
+    dest: /tmp/fio.job
+  when: fiod.job_url is defined and fiod.job_url != ""
```
chawlanikhil24: QUERY: if we are checking for fio.job_url is defined  ,then I don't think so , we need check if its an empty string, provided  it is mentioned in the doc that it has to be a working url.

roles/fio-distributed/tasks/main.yml
```diff
@@ -17,6 +17,30 @@
...
+  k8s:
+    definition:
+      apiVersion: v1
+      kind: ConfigMap
+      metadata:
+        name: fio-test
+        namespace: '{{ operator_namespace }}'
+      data:
+        fiojob: "{{ lookup('file', '/tmp/fio.job') }}"
+  when: fiod.job_url is defined and fiod.job_url != ""
```
chawlanikhil24: same as above

roles/fio-distributed/tasks/main.yml
```diff
@@ -27,6 +51,7 @@
         namespace: '{{ operator_namespace }}'
       data:
         fiojob: "{{ lookup('template', 'job.fio.{{fiod.job}}') }}"
+  when: fiod.job_url is not defined or fiod.job_url == ""
```
chawlanikhil24: same

roles/fio-distributed/tasks/main.yml
```diff
@@ -17,6 +17,30 @@
...
+      apiVersion: v1
+      kind: ConfigMap
+      metadata:
+        name: fio-test
+        namespace: '{{ operator_namespace }}'
+      data:
+        fiojob: "{{ lookup('file', '/tmp/fio.job') }}"
+  when: fiod.job_url is defined and fiod.job_url != ""
+
+- name: Delete fio job file
```
chawlanikhil24: Is there a need to delete the job file, since once the container has ended the job , the container has to be deleted, that means all the associated data in it too, unless the data is added onto a persistent volume, which in this case not going to happen.

roles/fio-distributed/tasks/main.yml
rsevilla87: Sure, makes sense, I'll remove it

roles/fio-distributed/tasks/main.yml
rsevilla87: The ansible code is executed in the ripsaw operator itself, this pod is not ephemeral so that we need to remove the job file after.

roles/fio-distributed/tasks/main.yml
```diff
@@ -17,6 +17,30 @@
...
+        namespace: '{{ operator_namespace }}'
+      data:
+        fiojob: "{{ lookup('file', '/tmp/fio.job') }}"
+  when: fiod.job_url is defined
+
+- name: Delete fio job file
+  file:
+    path: /tmp/fio.job
+    state: absent
+
```
aakarshg: Although file module will not fail if the path doesn't exist, we can still use a `when: fiod.job_url is defined` as this would mean a task that's not needed will not be performed like in most cases where we'd not use a custom job URL. Just an optimization we can put the 3 new tasks in a block and have a single when condition, so tasks like this one that don't need to be performed if `job_url` is not defined.

roles/fio-distributed/tasks/main.yml
```diff
@@ -17,6 +17,30 @@
     fio_path: /tmp
   when: fiod.storageclass is not defined
 
+- name: Download custom fio job
+  get_url:
+    url: "{{ fiod.job_url }}"
+    validate_certs: "{{ fiod.job_validate_certs }}"
+    dest: /tmp/fio.job
+  when: fiod.job_url
```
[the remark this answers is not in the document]
rsevilla87: sure, my fault :)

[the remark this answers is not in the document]
aakarshg: nbd, it's all good :) it happens to me a lot of times too.

jtaleric:
There are some conflicts that need resolution. 

```
- name: Generate fio test
  k8s:
<<<<<<< master
    definition:
      apiVersion: v1
      kind: ConfigMap
      metadata:
        name: fio-test
      data:
        fiojob: "{{ lookup('template', 'job.fio.{{fiod.job}}') }}"
  when: fiod.job_url is not defined
=======
    definition: "{{ lookup('template', 'configmap.yml.j2') | from_yaml }}"
>>>>>>> master
```

chawlanikhil24:
Looks good to me.

aakarshg:
Needs a rebase, also putting this on hold until we can come to an agreement on how/if we should run a custom job file with more than one fio job and index it.

jtaleric:
With this patch being so old, I suggest we close and re-visit this later.

PR #161, opened 2019-07-24
improve fio jobfile and allow for more customization in the CR

dustinblack:
There are several things going on in this commit:

1. The `elasticsearch` and `test_user` keys in the CR now have default values. This means that `elasticsearch.port`, `elasticsearch.index`, and `test_user` can all be left undefined or blank, and they will default to `9200`, `ripsaw-fio`, and `ripsaw`, respectively. The `elasticsearch.server` key can be left blank or undefined, and it will default to `''`, which will fail the jinja2 test in the template and skip adding the es environment vars.
2. Environment vars in the client template are now set in a k8s-native way, and the jinja2 logic surrounding the es stuff is simplified.
3. When defining the `storagesize` in the CR, you currently need to use IEC prefixes for base-2 values (KiB, MiB, etc). To make the values for fio consistent with this, the fio jobfile now sets `kb_base=1000`, which means that K or KB will be interpreted as base-10, while Ki or KiB will be base-2, consistent with the IEC standards. The CR examples have been updated to use IEC notation accordingly.
4. All fio jobfile defaults in the configmap have been moved to the `[global]` section. This assists with the implementation of item 5 below.
5. An "expert" section has been added to the CR files. This allows for fine user customization of the jobs, both by introducing overrides to the `[global]` values, which will be inserted at the `[job]` level, as well as per-job parameter insertions based on string matches.
6. Some general default jobfile changes:
 a. Introduced `unit_base='8'` to ensure all data is consistently reported as bytes instead of auto-adjusted between bits and bytes. (note a quirk/bug that the value here has to be quoted as '8' or the fio job will fail)
 b. Changed from `end_fsync=1` to `fsync_on_close=1` for write jobs (now defined in the CR expert section)
 c. Changed `create_on_open=1` to only apply to write jobs (now defined in the CR expert section; it is important that we don't use this parameter for read jobs)
7. Set `imagePullPolicy: Always` for the client container

There are a couple of other cosmetic changes not worth listing.

### DIFF
--- a/resources/crds/ripsaw_v1alpha1_fio_distributed_cr.yaml
+++ b/resources/crds/ripsaw_v1alpha1_fio_distributed_cr.yaml
@@ -1,9 +1,13 @@
 apiVersion: ripsaw.cloudbulldozer.io/v1alpha1
 kind: Benchmark
 metadata:
-  name: example-benchmark
+  name: fio-benchmark
   namespace: my-ripsaw
 spec:
+#  elasticsearch:
+#    server: <es_server>
+#    port: 9200
+#  test_user: ripsaw
   workload:
     name: "fio_distributed"
     args:
@@ -15,13 +19,32 @@ spec:
         - read
         - write
       bs:
-        - 64k
+        - 64Ki
       numjobs:
         - 1
       iodepth: 4
       runtime: 60
       ramp_time: 5
-      filesize: 2g
+      filesize: 2Gi
       log_sample_rate: 1000
       #storageclass: rook-ceph-block
       #storagesize: 5Gi
+#######################################
+#  EXPERT AREA - MODIFY WITH CAUTION  #
+#######################################
+#  global_overrides:
+#    - "key=value"
+  job_params:
+    - jobname_match: "w"
+      params:
+        - "fsync_on_close=1"
+        - "create_on_open=1"
+    - jobname_match: "rw"
+      params:
+        - "rwmixread=50"
+    - jobname_match: "readwrite"
+      params:
+        - "rwmixread=50"
+#    - jobname_match: "<search_string>"
+#      params:
+#        - "key=value"
--- a/roles/fio-distributed/defaults/main.yml
+++ b/roles/fio-distributed/defaults/main.yml
@@ -1,0 +1,6 @@
+---
+es: "{{ elasticsearch|default([]) }}"
+es_server: "{{ es.server|default('') }}"
+es_port: "{{ es.port|default(9200) }}"
+es_index: "ripsaw-fio"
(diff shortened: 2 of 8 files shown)


dry923:
A comment regarding the documentation compared to your PR description. The description here outlines nicely what is required/ignored when using elastic search (item 1 in the PR descriptions) however I cannot seem to find any of those details in the doc file. A little bit of a nit but it would be nice to see that spelled out some there as well (especially since it is commented out in the test cr).

dustinblack:
> A comment regarding the documentation compared to your PR description. The description here outlines nicely what is required/ignored when using elastic search (item 1 in the PR descriptions) however I cannot seem to find any of those details in the doc file. A little bit of a nit but it would be nice to see that spelled out some there as well (especially since it is commented out in the test cr).

I'm actually not sure that 1 in the PR description is 100% accurate now. I've had some trouble getting defaults to behave the way I want them to. I'll take a closer look and improve the documentation accordingly.

resources/crds/ripsaw_v1alpha1_fio_distributed_cr.yaml
```diff
@@ -1,9 +1,14 @@
...
 kind: Benchmark
 metadata:
-  name: example-benchmark
+  name: fio-benchmark
   namespace: my-ripsaw
 spec:
+#  elasticsearch:
+#    server: <es_server>
+#    port: 9200
+#    index: fio
```
jtaleric: I don't like that we provide a way to change the index. I prefer we explictily state we will use `ripsaw-fio-[results|logs]`. This will help with templating..  Just my .02

dustinblack: I'll adjust this. The way I've done it in my latest pgbench commit is to default it in the ripsaw role rather than hard coding "ripsaw-pgbench" in snafu, which technically shouldn't be dependent on ripsaw and so should have index name flexibility. This still allows for an undocumented override of the index name via the CR, but it removes it from being explicitly available as a configurable option.

jtaleric: i am also ok with prefixing with snafu-<workload>-<type>

roles/fio-distributed/defaults/main.yml
```diff
@@ -0,0 +1,6 @@
+---
+es: "{{ elasticsearch|default([]) }}"
+es_server: "{{ es.server|default('') }}"
+es_port: "{{ es.port|default(9200) }}"
+es_index: "ripsaw-pgbench"
```
aakarshg: `ripsaw-fio`

PR #165, opened 2019-07-25
Add uperf 'nthr' param to allow the option to run multi-threaded uper…

jeniferh:
…f instance inside the same pod


jeniferh:
Oh just realized I didn't update the test CRs, those are separate from resources/crds?

jeniferh:
Ah since I added another nthr value to test, it doubled the test matrix so the timeout is just a bit too short, let me increase that.

PR #166, opened 2019-07-25
tweaks for pgbench defaults and env vars

dustinblack:


roles/pgbench/templates/workload.yml.j2
```diff
@@ -14,27 +14,31 @@ spec:
...
-{% endif %}
-{% endif %}
-{% if test_user is defined %}
-             export test_user={{ test_user }};
-{% endif %}
-{% if pgbench.run_time is defined and pgbench.run_time|int > 0 %}
-             export duration={{ pgbench.run_time }};
-
-{% endif %}
+          - "export pgb_vers=$(pgbench --version);
```
jtaleric: Could this be moved to SNAFU?

dustinblack:
Rebased and resolved minor nits.

roles/pgbench/templates/workload.yml.j2
```diff
@@ -14,27 +14,30 @@ spec:
...
+          - name: uuid
+            value: '{{ uuid }}'
+          - name: pgbench_auth
+            value: >-
+                   -h {{ item.1.host }}
+                   -p {% if item.1.port is defined and item.1.port|int > 0 %} {{ item.1.port }} {% else %} {{ db_port }} {% endif %}
+                   -U {{ item.1.user }}
+          - name: test_user
+            value: '{{ test_user }}'
+{% if es_server %}
```
aakarshg: shouldn't this be `es_server|length > 0` ? Because we're by default initializing the var to an empty string?

dustinblack: No; this is valid jinja2 logic that will pass if the var is non-empty.
https://jinja.palletsprojects.com/en/2.10.x/templates/#if

PR #168, opened 2019-07-26
Uperf fix

jtaleric:
Fixing UPerf logic
    
    In this commit we are doing a couple of things:
    
    1) Manage our own CR State.
    - This will allow us to manage the state of our benchmarks. If the
    benchmark has many phases we can capture that.
    2) UPerf Logic no longer needs changes to the CR
    - Previously we would need to change parameters in Uperf to have the
    same CR run. This PR removes that logic, making things easier to run the
    same CR multiple times.


jtaleric:
Changing this to a clean PR.

PR #170, opened 2019-07-26
moving sample loop to snafu

aakarshg:
Follows merge of https://github.com/cloud-bulldozer/snafu/pull/22


jtaleric:
> I see you've already updated the snafu script, so doesn't reusing the `-s` flag break current tests until this PR is merged?

It shouldn't since `-s` used to be for the iteration number, and our test scripts shouldn't have > 1 iteration...

PR #171, opened 2019-07-29
Mange the status of the workloads

jtaleric:
The built-in status from the Ansible Operator doesn't fit the different
cycles a benchmark workflow goes through. For example, our YCSB workload
has 3 phases:

- Starting
- Load
- Run

Other workloads such as smallfile have different stages of run the run
phase.

This PR only addresses UPerf, but will serve as an example going
forward.


dry923:
Like it. LGTM

PR #174, opened 2019-08-05
Pgbench timeout

dustinblack:
Note: This includes the commits for #166 and so is intended to be merged behind that one.

The changes here get rid of the user-supplied timeout in the CR and instead use a series of simple calculations to dynamically determine appropriate timeouts for the load and run phases based on the scale and loop iterations of the test.

### DIFF
--- a/roles/pgbench/tasks/run_workload.yml
+++ b/roles/pgbench/tasks/run_workload.yml
@@ -5,6 +5,46 @@
 - name: Init client ready state to empty
   command: "redis-cli del pgb_client_ready"
 
+- name: Calculate timeouts
+  include_tasks: calculate_timeouts.yml
+
+- name: Pre-pull pgbench container image
+  k8s:
+    definition:
+      kind: Job
+      apiVersion: batch/v1
+      metadata:
+        name: pgb-image-load-{{ item.0 }}
+        namespace: "{{ operator_namespace }}"
+        labels:
+          app: pgb-image-load
+      spec:
+        template:
+          spec:
+            containers:
+              - name: pgb-image-load-{{ item.0 }}
+                image: "{{ pgb_base_image }}:{{ pgb_version }}"
+                command: ['/bin/true']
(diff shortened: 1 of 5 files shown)


roles/pgbench/tasks/run_workload.yml
```diff
@@ -5,6 +5,46 @@
 - name: Init client ready state to empty
   command: "redis-cli del pgb_client_ready"
 
+- name: Calculate timeouts
+  include_tasks: calculate_timeouts.yml
+
+- name: Pre-pull pgbench container image
```
aakarshg: Do we need this? This feels like an additonal step.

dustinblack: Image download time is harder to predict. This is done as a pre-pull so that it can have its own timeout value as an operation separate from starting the pod(s).

dustinblack: Look at it this way -- If I _don't_ separate out the pre-pull, then I need to add the tolerated maximum amount of time for the image pull (currently about 11 minutes x number of databases) to the load_timeout. This permanently expands the load_timeout by this added image pull tolerance time, regardless of whether the image is already present on the k8s nodes. With the submitted design, if the images _are_ already present on the system, then the pre-pull jobs will complete very quickly and we will move on to the pgbench test job(s) task, which then has calculated timeouts relevant to only how long it should take to start/load the pods given that the image is already local to the nodes.

roles/pgbench/tasks/run_workload.yml
```diff
@@ -31,5 +70,5 @@
       - app = pgbench-client
   register: pgbench_pods
   until: "'Succeeded' in (pgbench_pods | json_query('resources[].status.phase'))"
```
jtaleric: I noted this in chat, but this will *stop* the first pod that completes. I have a fix, and we don't need to block for this. Just noting.

PR #176, opened 2019-08-06
Moving some of the CR status to the playbook

jtaleric:
Now all workloads will have a "Starting" status. We need to implement
the Running and Complete for each.

This additionally adds Load,<workload>, Complete for YCSB.

### DIFF
--- a/roles/byowl/tasks/main.yml
+++ b/roles/byowl/tasks/main.yml
@@ -6,3 +6,23 @@
     state: present
     definition: "{{ lookup('template', 'workload.yml') | from_yaml }}"
   when: byowl.kind is not defined
+
+- name: Wait for fiod job to succeed...
+  k8s_facts:
+    kind: Job
+    api_version: batch/v1
+    name: '{{ meta.name }}-byowl'
+    namespace: "{{ operator_namespace }}"
+  register: byowl_jobs
+  until: "byowl_jobs | json_query('resources[].status.succeeded')"
+  retries: 1000
+  delay: 20
+
+- k8s_status:
+    api_version: ripsaw.cloudbulldozer.io/v1alpha1
+    kind: Benchmark
+    name: "{{ meta.name }}"
+    namespace: "{{ operator_namespace }}"
+    status:
+      state: "Complete"
+      complete: true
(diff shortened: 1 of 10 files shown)


roles/load-ycsb/tasks/main.yml
```diff
@@ -1,4 +1,13 @@
 ---
+
+- name: Get current state
```
aakarshg: Don't think its needed here.

jtaleric: yup, you are right.

roles/byowl/tasks/main.yml
```diff
@@ -6,3 +6,23 @@
     state: present
     definition: "{{ lookup('template', 'workload.yml') | from_yaml }}"
   when: byowl.kind is not defined
+
+- name: Wait for fiod job to succeed...
```
dustinblack: s/fiod/byowl/

roles/byowl/tasks/main.yml
```diff
@@ -6,3 +6,23 @@
...
+    kind: Job
+    api_version: batch/v1
+    name: '{{ meta.name }}-byowl'
+    namespace: "{{ operator_namespace }}"
+  register: byowl_jobs
+  until: "byowl_jobs | json_query('resources[].status.succeeded')"
+  retries: 1000
+  delay: 20
+
+- k8s_status:
```
dustinblack: Should this have a `when` condition based on the success of the 'wait for job to succeed' task?

roles/fio-distributed/tasks/main.yml
```diff
@@ -78,3 +78,23 @@
...
+    kind: Job
+    api_version: batch/v1
+    name: '{{ meta.name }}-fio-client'
+    namespace: "{{ operator_namespace }}"
+  register: fiod_jobs
+  until: "fiod_jobs | json_query('resources[].status.succeeded')"
+  retries: 1000
+  delay: 20
+
+- k8s_status:
```
dustinblack: Same as above -- Do we want to set `Complete` to `true` even if the previous task times out?

jtaleric: yes we do.

`Completed` does not necessarily mean `Success`.

roles/byowl/tasks/main.yml
jtaleric: `Completed` does not necessarily mean `Success`.

PR #177, opened 2019-08-09
Doubled existing timeouts

dry923:
To help with the timeout issues we have been seeing recently I doubled all the existing timeouts in the test scripts.


aakarshg:
can we close this out @dry923 seeing that its no longer needed

dry923:
Agreed. Closing

PR #178, opened 2019-08-09
Adding clustername var to roles that index benchmark results

aakarshg:
Follows https://github.com/cloud-bulldozer/snafu/pull/34/files


aakarshg:
we don't need it, I've defaulted it in the group_vars/all.yml. I've added it to the default crs to avoid confusion.

PR #180, opened 2019-08-12
defaulting image pull policy to always

aakarshg:
Although this will increase the time to run the CI test as it'll always pull down the latest image, it will also ensure that no one is running tests with an older version of the benchmark image.

We should be following this up with using a specific tag/version for each of the benchmark images as well, so that once we implement versioning in Ripsaw we can then accordingly have a branch in magazine/snafu with the same number.

### DIFF
--- a/roles/smallfile-bench/templates/workload_job.yml.j2
+++ b/roles/smallfile-bench/templates/workload_job.yml.j2
@@ -14,6 +14,7 @@ spec:
       containers:
         - name: benchmark-server
           image: "quay.io/cloud-bulldozer/smallfile:master"
+          imagePullPolicy: Always
           command: ["/bin/sh", "-c"]
           workingDir: /root/smallfile-master/
           args:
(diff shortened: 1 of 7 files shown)


dry923:
imagePullPoicy should be imagePullPolicy but I agree with the change

roles/smallfile-bench/templates/workload_job.yml.j2
```diff
@@ -14,6 +14,7 @@ spec:
       containers:
         - name: benchmark-server
           image: "quay.io/cloud-bulldozer/smallfile:master"
```
dry923: Should this be latest or master? Obviously changing this to a version number in the long run but atm?

aakarshg: both master and latest shall point to the same based on how i setup the triggers in quay.io, but either way out of scope of this PR

dry923: ack sounds good.

dry923:
LGTM

PR #181, opened 2019-08-12
Added results.xml to be passed on CI completion

dry923:
This will be used to (hopefully) pass more information back to github on completion of a CI run.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -30,6 +30,20 @@ echo "running test suit consisting of ${test_list}"
 
 pass=0
 
+testcount=0
+failcount=0
+
+# Prep the results.xml file
+echo '<?xml version="1.0" encoding="UTF-8"?>
+<testsuites>
+   <testsuite name="CI Results" tests="NUMTESTS" failures="NUMFAILURES">' > results.xml
+
+# Prep the results.markdown file
+echo "Results for "$JOB_NAME > results.markdown
+echo "" >> results.markdown
+echo 'Test | Result' >> results.markdown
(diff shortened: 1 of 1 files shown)


test.sh
```diff
@@ -30,6 +30,18 @@ echo "running test suit consisting of ${test_list}"
...
 
+testcount=0
+failcount=0
+
+# Prep the results.xml file
+echo '<?xml version="1.0" encoding="UTF-8"?>
+<testsuites>
+   <testsuite name="CI Results" tests="NUMTESTS" failures="NUMFAILURES">' > results.xml
+
+# Prep the results.markdown file
```
aakarshg: we should add the job name here as well, so that we know which job the fail/pass is associated with.

dry923: This corresponds to the overall ci test so can only show total failures. The details get saved in the ci system which shows which specific ones failed. Additionally, a markdown of the actual pass/fails gets dropped into the PR as a comment. So hopefully that should cover it.

test.sh
```diff
@@ -30,6 +30,19 @@ echo "running test suit consisting of ${test_list}"
...
+echo "Results for: "$JOB_NAME > results.markdown
```
dry923: @aakarshg This should resolve your concern I believe

aakarshg:
/hold until we add the job definition for minikube and minishift as well so we can see this in action. 

@dry923 lmk when you push the job definitions to update the three jobs when you get time so that we can merge this.

PR #183, opened 2019-08-14
changed dash to underscore for fio wrapper

acalhounRH:
This change is needed because snafu is using invalid syntax for python
directory/file names

### DIFF
--- a/roles/fio-distributed/templates/client.yaml
+++ b/roles/fio-distributed/templates/client.yaml
@@ -36,7 +36,7 @@ spec:
              for file in $(ls /tmp/fio/ | egrep ^fiojob); do
                cat /tmp/fio/$file;
                mkdir -p /tmp/fiod-{{ uuid }}/$file;
-               python /opt/snafu/fio-wrapper/fio-wrapper.py /tmp/host/hosts /tmp/fio/$file -s {{fiod.samples}} -d /tmp/fiod-{{ uuid }}/$file;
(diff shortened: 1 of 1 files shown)


dustinblack:
Obviously this is simple and there's no reason to reject it, so long as the related snafu changes are made. I don't fully understand what this fixes, though, and I'm curious.

acalhounRH:
The reason for this fix is to support "run_snafu.py" and conform to python standards. 

- With the current implementation this structure doesn't cause an error because the module is either self-contained or relies on modules in the same directory. 

- With the implementation of run_snafu.py, run_snafu would be a parent module of all wrapper module, and need to import each sub-module in order to handle that specific benchmark. 

directory example:
```
working_dir/snafu
        run_snafu.py
	__init__.py
	fio-wrapper/
		__init__.py
		fio-wrapper.py
		...
	uperf-wrapper/
	...
```

because of this run_snafu.py will import the wrapper scripts as sub-modules, if using dash "-" with in the directory name and/or file name, snippet below,

```
#run_snafu.py
...
from fio-wrapper import fio-wrapper
from uperf-wrapper import fio-wrapper
...
````   
the python interpreter will fail on a syntax error.

**Comment by acalhounRH** [comment]
Example execution of run_snafu.py
```
bash-4.2$ python snafu/run_snafu.py -t fio -H /tmp/host/hosts -j /tmp/fio/fiojob-write-64Ki-1 -s 1 -d /tmp/test_dir/test_output
  File "snafu/run_snafu.py", line 28, in <module>
    from utils.wrapper_factory import wrapper_factory
  File "/tmp/test_dir/snafu/utils/wrapper_factory.py", line 2
    from backpack-wrapper import backpack-wrapper
                 ^
SyntaxError: invalid syntax
```

PR #187, opened 2019-08-15
Run CI tests in parallel

dry923:
Since we don't really care about the performance results when doing CI testing we should run multiple CI tests at once. This will set it to run 3 at a time. We may need to adjust timeouts up to account for the new load on the system. Additionally, we will need to validate the CI output is usable and not jumbled together.


dry923:
Closing this while I finish cleaning up the job. We re-open when ready.

PR #189, opened 2019-08-19
Run CI tests in parallel.

dry923:
Additionally, increased timeouts due to greater concurrent load.

This will run up to 4 tests concurrently (this can be adjusted up/down as necessary). Logs will be posted to CI as the job completes to avoid output collision with error messages.


dustinblack:
I think some of the changes from this got pulled in with #196? Give this a rebase.

dry923:
Removed the increased timeouts as they do not seem necessary atm. Retesting to be sure.

PR #191, opened 2019-08-19
Control looping and show uuid

aakarshg:
Essentially this commit will facilitate:
1. showing uuid when doing a get benchmarks
2. running multiple benchmarks of even same type
3. avoid unnecessary looping

### DIFF
--- a/roles/ycsb-bench/templates/configmap.yml.j2
+++ b/roles/ycsb-bench/templates/configmap.yml.j2
@@ -2,7 +2,7 @@
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: ycsb-test
+  name: ycsb-test-{{ trunc_uuid }}
   namespace: '{{ operator_namespace }}'
 data:
   tsworkload : |
--- a/tests/common.sh
+++ b/tests/common.sh
@@ -37,6 +37,29 @@ function wait_clean {
   done
 }
 
+# The argument is 'timeout in seconds'
+function get_uuid () {
+  sleep_time=$1
+  sleep $sleep_time
+  counter=0
+  counter_max=6
+  uuid="False"
+  until [ $uuid != "False" ] ; do
+    uuid=$(kubectl -n my-ripsaw get benchmarks -o jsonpath='{.items[0].status.uuid}')
+    if [ -z $uuid ]; then
+      sleep $sleep_time
+      uuid="False"
+    fi
+    counter=$(( counter+1 ))
+    if [ $counter -eq $counter_max ]; then
+      return 1
+    fi
+  done
+  echo ${uuid:0:8}
+  return 0
+}
+
+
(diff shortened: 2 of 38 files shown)


roles/ycsb-bench/templates/configmap.yml.j2
```diff
@@ -3,7 +3,7 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: ycsb-test
-  namespace: '{{ operator_namespace }}'
+  namespace: '{{ operator_namespace }}-{{ uuid }}'
```
jtaleric: bad copy/pasta?

aakarshg: yesss, nice catch will fix it :)

tests/common.sh
```diff
@@ -37,6 +37,28 @@ function wait_clean {
...
 }
 
+# The argument is 'timeout in seconds'
+function get_uuid () {
+  counter=0
+  sleep_time=5
+  counter_max=$(( $1 / sleep_time ))
+  uuid="False"
+  until [ $uuid != "False" ] ; do
+    sleep $sleep_time
```
dry923: If I'm following this right could we move the ```sleep $sleep_time``` into your if statement? Right now we always sleep X seconds before trying the first time. If we move it we might avoid a sleep.

aakarshg: that makes sense. updated it

jtaleric:
Is something slowing down the pulling of the Operator/Redis/Sidecar? 

@dry923 we might want to time these things -- or time all things. Jenkins has a way to present timestamps in the console (or at least I haven seen it done before). This might help us?

dry923:
> Is something slowing down the pulling of the Operator/Redis/Sidecar?
> 
> @dry923 we might want to time these things -- or time all things. Jenkins has a way to present timestamps in the console (or at least I haven seen it done before). This might help us?

@jtaleric Agreed. I enabled timestamps in the console of Jenkins. I was thinking about other options as well (timing individual tests, etc) but we can discuss those later. 

Take a look at the output from the latest run and see if that helps any.

tests/common.sh
```diff
@@ -37,6 +37,29 @@ function wait_clean {
...
+  sleep 5
+  counter=0
+  sleep_time=5
+  counter_max=$(( $1 / sleep_time ))
+  uuid="False"
+  until [ $uuid != "False" ] ; do
+    if [ -z $uuid ]; then
+      uuid=$(kubectl -n my-ripsaw get benchmarks -o jsonpath='{.items[0].status.uuid}')
+      sleep $sleep_time
+      uuid="False"
```
dry923: This code will never succeed. You obtain the uuid then set it to false two lines later.

jtaleric: @aakarshg

jtaleric:
KNI built fine. Good to merge.

PR #192, opened 2019-08-20
fs-drift benchmark

bengland2:
This has been tested with both RBD block storage and with Cephfs storage (in AWS OCP4.1 Ceph cluster).   However, the test scripts have not been through CI, could use some help with that, particularly what storage class to use.   Sorry I couldn't successfully squash this, I could create a fresh branch for the PR if necessary, but if you look at "files changed" tab you can see what I did pretty clearly.   The PR is similar to smallfile-bench by Nikhil.


roles/fs-drift/tasks/main.yml
```diff
@@ -0,0 +1,85 @@
...
+
+- name: Create PVCs
+  k8s:
+    definition:
+      kind: PersistentVolumeClaim
+      apiVersion: v1
+      metadata:
+        name: "fs-drift-claim-{{ item }}"
+        namespace: "{{ operator_namespace }}"
+        annotations:
```
aakarshg: we can put this file in a template and add a check condition to use a specific storage class only when fs_drift.storageclass is defined so that if user doesn't provide a name it'll auto pick the one with annonation default

bengland2:
Aakarsh is right, no need for that file, removing it < 5 min

bengland2:
wow it ignored "do not need this file" commit but it was in the commit log before I pushed it.

bengland2:
ok, I probably shouldn't have reduced that sleep to 20, but at least I have some idea what's causing it now.  Will boost up to 45.

PR #207, opened 2019-08-29
rearrange fio loop order

dustinblack:
After running some tests, it is clear to me that the order of the nested workload loops is inverted from what it should be for best analysis of the data. This change re-orders the loops and updates the related documentation.


roles/fio-distributed/templates/configmap.yml.j2
```diff
@@ -5,9 +5,9 @@ metadata:
   name: fio-test-{{ trunc_uuid }}
   namespace: '{{ operator_namespace }}'
 data:
-{% for job in fiod.jobs %}
-{% for bs in fiod.bs %}
 {% for numjobs in fiod.numjobs %}
+{% for bs in fiod.bs %}
```
aakarshg: this part of change is ideally not needed, because the order in which the files are created is not gonna affect the way we loop to trigger fio.

PR #209, opened 2019-09-03
add pod antiaffinity for fio server pods

dustinblack:


dustinblack:
Some testing confirms that implementing this anti-affinity will cause even distribution of the fio server pods among the schedulable nodes. Without this code, the scheduler will tend to place the server pods in an imbalanced way due to existing pod load across the cluster.

PR #212, opened 2019-09-09
Ansible-ize the CI test scripts

dry923:
Converted the test_*.sh scripts into roles. With the addition of reconciliation we are able to boil down most test scripts to a few lines of unique variable definitions and reuse roles for all the work. The two exceptions being ycsb and pgbench. Since they do not have reconciliation enabled yet they have their own specific roles. Additionally, FSdrift will be added once PR #211 is completed.


dry923:
To fix the CI issues being seen on minishift this will need PR #215

tests/group_vars/all.yml
```diff
@@ -0,0 +1,2 @@
+namespace: my-ripsaw
+ripsaw_dir: /home/dry/working/ripsaw
```
dustinblack: Does this need to be changed?

dry923: It should be doc'd most likely. That dir gets overwritten with the current working directory when run via the test.sh script.

dustinblack: Are you saying there are docs updated somewhere in this PR that I'm not seeing, or are you acknowledging that docs need to be done? Stupid english--

dry923: Acknowledging that docs need to be done for this change.

dry923: @dustinblack please check out the updated docs

jtaleric:
@dry923 we need a rebase it seems?

dry923:
Closing the PR for now. It may be revisited later

PR #213, opened 2019-09-09
Optionally deploy privileged pod for dropping ceph osd caches via url

dustinblack:
So far this automates the deployment of the privileged pod for dropping ceph osd caches. This doesn't yet included automation to issue the curl command to the pod.


dustinblack:
Further progress here is stalled pending https://github.com/axboe/fio/issues/814

tl;dr -- fio fails to execute jobs when either `exec_prerun` or `exec_postrun` are used.

This was intended to be the simple method of automating the cache drop between job runs. Short of this, we would need to modify the snafu scripts to inject the URL call before each run.

roles/fio-distributed/tasks/main.yml
```diff
@@ -27,6 +26,11 @@
   register: resource_state
 
 - block:
+  - name: Include rook-ceph cache drop role
+    include_role:
+      name: ceph-cache-dropper
+    when: fiod.rook_ceph_drop_caches is sameas True | default(False)
```
bengland2: I'm not saying you're wrong, just trying to understand.   where is "sameas" defined?  Never saw it before.   Didn't show up in google.  Why is this necessary vs just 
```
when: fiod.rook_ceph_drop_caches
```
and define this variable's default in defaults/main.yml

dustinblack: We've had some trouble with jinja2 logic not working as defined or expected. The `is sameas True` method looks ridiculous, but it works. I don't remember if I had a problem in this particular case or if I even tested it with the much cleaner `when: fiod.rook_ceph_drop_caches`, but I know we've dealt with this in other code (from which I may have copy pasta'd here).

Here is one reference:
https://www.shellhacks.com/jinja2-check-if-variable-empty-exists-defined-true/

And here's the jinja2 doc reference for `sameas`:
https://jinja.palletsprojects.com/en/2.10.x/templates/#sameas

dustinblack:
Closed in favor of PR237

PR #216, opened 2019-09-11
Replace xargs with parallel

dry923:
To fix the mashed together output from the use of xargs I've swapped to parallel which keeps the output for each test together.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -61,7 +61,7 @@ do
 done
 
 # Run tests in parallel up to $max_concurrent at a time.
-xargs -n 1 -a tests/my_tests -P $max_concurrent ./run_test.sh 
+parallel -n 1 -a tests/my_tests -P $max_concurrent ./run_test.sh 
 
 # Update and close JUnit test results.xml and markdown file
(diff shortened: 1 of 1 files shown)


test.sh
```diff
@@ -61,7 +61,7 @@ do
 done
 
 # Run tests in parallel up to $max_concurrent at a time.
-xargs -n 1 -a tests/my_tests -P $max_concurrent ./run_test.sh 
+parallel -n 1 -a tests/my_tests -P $max_concurrent ./run_test.sh
```
jtaleric: should we provide -j and # of jobs we want? By default it will be # of cores..

dry923: -P should be doing the same. From the man page:
       --jobs N
       -j N
       --max-procs N
       -P N
           Number of jobslots on each machine. Run up to N jobs in parallel.  0 means as many
           as possible. Default is 100% which will run one job per CPU core on each machine.

jtaleric: ah, TIL! Good to go.

aakarshg:
minishift is failing on sysbench but passed on other envs so gonna just merge

PR #217, opened 2019-09-11
There ain't no rest for the wicked!

dustinblack:
Fix a bunch of silly sleeps in the CI scripts.

### DIFF
--- a/tests/common.sh
+++ b/tests/common.sh
@@ -40,21 +40,20 @@ function wait_clean {
 
 # The argument is 'timeout in seconds'
 function get_uuid () {
-  sleep_time=$1
-  sleep $sleep_time
   counter=0
-  counter_max=6
+  sleep_time=5
+  counter_max=$(( $1 / sleep_time ))
   uuid="False"
   until [ $uuid != "False" ] ; do
     uuid=$(kubectl -n my-ripsaw get benchmarks -o jsonpath='{.items[0].status.uuid}')
     if [ -z $uuid ]; then
-      sleep $sleep_time
       uuid="False"
     fi
     counter=$(( counter+1 ))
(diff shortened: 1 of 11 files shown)


dry923:
I'd hold on this atm. A) adding PR #215 will help some also if we go with PR #212 then all of these loops and sleeps are gone anyway

dustinblack:
> I'd hold on this atm. A) adding PR #215 will help some also if we go with PR #212 then all of these loops and sleeps are gone anyway

AFAICT neither of those addresses the sleep annoyances in the CI scripts, but maybe the CI work just isn't done relative to those changes yet?

These sleeps, as they are today, are just outright silly are are slowing down the script executions. Working with the scripts manually in local tests I have problems like waiting 20 seconds for the get_uuid function to complete just because it sleeps before it actually does $thing.

dry923:
> > I'd hold on this atm. A) adding PR #215 will help some also if we go with PR #212 then all of these loops and sleeps are gone anyway
> 
> AFAICT neither of those addresses the sleep annoyances in the CI scripts, but maybe the CI work just isn't done relative to those changes yet?
> 
> These sleeps, as they are today, are just outright silly are are slowing down the script executions. Working with the scripts manually in local tests I have problems like waiting 20 seconds for the get_uuid function to complete just because it sleeps before it actually does $thing.

I don't disagree about replacing the sleeps however PR #212 completely replaces everything in the tests directory. There are no sleeps or random waits except a 2 second one on two tests which I can likely remove at this point. Waiting is done with until loops that look for specific states or information to be present or timeouts to be hit. I think that "should" help with the sleep issue unless I'm missing something completely.

aakarshg:
Unless I'm mistaken, looks like #212 would make this change obsolete, so maybe we can close this ?

PR #218, opened 2019-09-11
Uperf on VMinstances

aakarshg:
This will allow Ripsaw to orchestrate Uperf on VMs.
please note: I'm not gonna add a CI until kubevirt/cnv stabilizes - rn we're hitting oom and all sorts of problems while the job itself would take more than the current test suite run time as for some reason custom container-disk-images only seem to work on kubevirt without any issue

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,18 +5,18 @@ a performance baseline of Kubernetes cluster on your provider.
 
 ## Workloads status
 
-| Workload                       | Use                    | Status in Operator | Reconciliation usage       |
-| ------------------------------ | ---------------------- | ------------------ | -------------------------- |
-| [UPerf](docs/uperf.md)         | Network Performance    | Working            |  Used, default : 30second  |
-| [Iperf3](docs/iperf3.md)       | Network Performance    | Working            |  Used, default : 30second  |
-| [fio](docs/fio_distributed.md) | Storage IO             | Working            |  Used, default : 30second  |
-| [Sysbench](docs/sysbench.md)   | System Performance     | Working            |  Used, default : 30second  |
-| [YCSB](docs/ycsb.md)           | Database Performance   | Working            |  Not used                  |
-| [Byowl](docs/byowl.md)         | User defined workload  | Working            |  Used, default : 30second  |
-| [Pgbench](docs/pgbench.md)     | Postgres Performance   | Working            |  Not used                  |
-| [Smallfile](docs/smallfile.md) | Storage IO Performance | Working            |  Used, default : 30second  |
-| [fs-drift](docs/fs-drift.md)   | Storage IO Longevity   | Working            |  Not used                  |
-| [hammerdb](docs/hammerdb.md)   | Database Performance   | Working            |  Not used                  |
+| Workload                       | Use                    | Status in Operator | Reconciliation usage       | VM support (kubevirt) |
+| ------------------------------ | ---------------------- | ------------------ | -------------------------- | --------------------- |
+| [UPerf](docs/uperf.md)         | Network Performance    | Working            |  Used, default : 30second  | Preview               |
+| [Iperf3](docs/iperf3.md)       | Network Performance    | Working            |  Used, default : 30second  | Not Supported         |
+| [fio](docs/fio_distributed.md) | Storage IO             | Working            |  Used, default : 30second  | Not Supported         |
+| [Sysbench](docs/sysbench.md)   | System Performance     | Working            |  Used, default : 30second  | Not Supported         |
+| [YCSB](docs/ycsb.md)           | Database Performance   | Working            |  Not used                  | Not Supported         |
+| [Byowl](docs/byowl.md)         | User defined workload  | Working            |  Used, default : 30second  | Not Supported         |
+| [Pgbench](docs/pgbench.md)     | Postgres Performance   | Working            |  Not used                  | Not Supported         |
+| [Smallfile](docs/smallfile.md) | Storage IO Performance | Working            |  Used, default : 30second  | Not Supported         |
+| [fs-drift](docs/fs-drift.md)   | Storage IO Longevity   | Working            |  Not used                  | Not Supported         |
+| [hammerdb](docs/hammerdb.md)   | Database Performance   | Working            |  Not used                  | Not Supported         |
 
 
 ### Reconciliation
--- a/docs/uperf.md
+++ b/docs/uperf.md
@@ -20,6 +20,7 @@ spec:
       serviceip: false
       hostnetwork: false
       pin: false
+      kind: pod
       pin_server: "node-0"
       pin_client: "node-1"
       multus:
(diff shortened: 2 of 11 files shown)


aakarshg:
NOTE: NOT WORKING, this is just for review reasons. have to test some more.

jeniferh:
Testing uperf I am still seeing the client's volumecontainerdisk get OOMKilled, although not every time. Can you tell me what Limits you get for that volume in 'oc describe virt-launcher-uperf-client-xxx'?

for ex. I see:
```
Containers:
  volumecontainerdisk:
    Container ID:  cri-o://ae7b13ad49e0cc8ba5a482e85a3a95f4c29cff76f6f922b89804d03a0322fceb
    Image:         kubevirt/fedora-cloud-container-disk-demo:latest
    Image ID:      docker.io/kubevirt/fedora-cloud-container-disk-demo@sha256:374f3d742d7dc9b30c64d27cd0b4b1031abf4df014080ced24a4f56fa7ebe1e4
    Port:          <none>
    Host Port:     <none>
    Command:
      /usr/bin/container-disk
    Args:
      --copy-path
      /var/run/kubevirt-ephemeral-disks/container-disk-data/855597bf-d65e-11e9-9104-3cfdfec4f600/disk_0
    State:          Terminated
      Reason:       OOMKilled
      Exit Code:    137
      Started:      Fri, 13 Sep 2019 19:41:52 +0000
      Finished:     Fri, 13 Sep 2019 19:42:14 +0000
    Ready:          False
    Restart Count:  0
    Limits:
      cpu:     100m
      memory:  20M
    Requests:
      cpu:        10m
      memory:     1M
```

I don't have a storageclass defined by default, is that needed?

jeniferh:
Interesting, watching the client console I wonder if downloading the updates causes the OOM?

```
[    6.044215] cloud-init[626]: Cloud-init v. 17.1 running 'modules:config' at Fri, 13 Sep 2019 20:17:01 +0000. Up 5.89 seconds.
[  OK  ] Started Apply the settings specified in cloud-config.
         Starting Execute cloud user/final scripts...

Fedora 29 (Cloud Edition)
Kernel 4.18.16-300.fc29.x86_64 on an x86_64 (ttyS0)

uperf-client-10 login: [   11.801561] cloud-init[652]: Fedora Modular 29 - x86_64                      308 kB/s | 1.5 MB     00:04
[   16.190321] cloud-init[652]: Fedora Modular 29 - x86_64 - Updates            502 kB/s | 2.0 MB     00:04
uperf-client-10 login: [   49.861184] cloud-init[652]: Fedora 29 - x86_64 - Updates                    900 kB/s |  29 MB     00:33
```

aakarshg:
@jeniferh please review

roles/uperf-bench/templates/server_vm.yml.j2
```diff
@@ -0,0 +1,68 @@
...
+      - disk:
+          bus: virtio
+        name: containerdisk
+      - disk:
+          bus: virtio
+        name: cloudinitdisk
+    interfaces:
+      - name: default
+        {{ uperf.server_vm.network.frontend }}: {}
+{% if uperf.server_vm.network_type_frontend is sameas bridge %}
```
jeniferh: can remove this check

jeniferh:
Actually, thinking about masquerade mode a bit more.. I am not sure 'resources[].status.interfaces[0].ipAddress' is going to work since the vmi shows the pod ip at first but then at some point it is updated with the dummy ip (ex. 10.0.2.2/24). So 'resources[].status.podIP' might be better in general.

docs/uperf.md
```diff
@@ -20,6 +20,7 @@ spec:
       serviceip: false
       hostnetwork: false
       pin: false
+      kind: pod
```
jtaleric: Could we just default to pod, so we could leave it out of the definition?

docs/uperf.md
```diff
@@ -81,7 +82,79 @@ Once done creating/editing the resource file, you can run it by:
...
+
+You must've configured your k8s cluster with [Kubevirt](https://kubevirt.io) preferably v0.20.3.
+
+
+### changes to cr file
+
+```yaml
+kind: vm
+server_vm:
+  dedicatedcpuplacement: false
```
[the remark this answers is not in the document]
jtaleric: I would add this as a note in the doc. /cc @aakarshg

roles/fio-distributed/templates/server_vm.yaml
```diff
@@ -0,0 +1,62 @@
...
+      - name: test-volume
+        # This makes it a disk
+        disk: {}
+        serial: CVLY623300HK250D
+{% endif %}
+    machine:
+      type: ""
+    resources:
+      requests:
+        memory: 1024M
```
jtaleric: Shouldn't this be a template value from the CR?

docs/uperf.md
```diff
@@ -81,7 +82,79 @@ Once done creating/editing the resource file, you can run it by:
...
 ```
 
-# Storing results into Elasticsearch
+## Running Uperf in VMs through kubevirt/cnv [Preview]
+Note: this is currently in preview mode.
+
+
+### Pre-requisites
+
+You must've configured your k8s cluster with [Kubevirt](https://kubevirt.io) preferably v0.20.3.
```
dry923: Ok i'm gonna be that guy.... must have instead of "must've" :-)

roles/fio-distributed/templates/server_vm.yaml
jtaleric: Should this be static? If there is a issue, we literally have to change the code in the operator, it might be better to allow the user to change this.

roles/uperf-bench/templates/server_vm.yml.j2
```diff
@@ -0,0 +1,66 @@
...
+        #cloud-config
+        password: ripsaw
+        chpasswd: { expire: False }
+        runcmd:
+{% if uperf.server_vm.network.multiqueue.enabled %}
+          - ethtool -L eth0 combined {{ uperf.server_vm.network.multiqueue.queues }}
+{% endif %}
+          - yum install -y yum-plugin-copr
+          - yum copr enable ndokos/pbench -y
+          - yum install -y pbench-uperf redis git
```
jtaleric: Ugh. I wonder if there is a way we can do this from an outside script? maybe curl/wget a set of instructions on what to install? My concern is if something changes with pbench, we have to re-spin a new operator each time. 

As much as this doesn't sound like a problem, it became a problem with Browbeat. Every time someone would try the PBench work with Browbeat something was broken with the install of PBench... Just something we need to keep an eye on.

jtaleric:
Additionally, I do not have a CNV env to kick the tires on this.

docs/uperf.md
aakarshg: This is just first iteration, maybe with subsequent pr we can do that?

roles/uperf-bench/templates/server_vm.yml.j2
aakarshg: We can look at using custom container-disk-images, but so far I've ran out of luck with it. maybe for subsequent iterations when there's enough adoption we could do that. But either way, given that it's iteration 0, i'd not try to worry about it rn.

jeniferh:
I've been testing on my kubevirt cluster and bridge/masq/vcpus all work fine, however I can't get multiqueue to work. The VM image doesn't appear to have ethtool installed to set the combined queues, but the client VM actually gets to an Error state with a strange complaint about applying queues to the disk.. trying to figure out where that is coming from: 
`{"component":"virt-launcher","level":"error","msg":"unsupported configuration: queues attribute in disk driver element is only supported by virtio-blk","pos":"virDomainDiskDefValidate:5428","subcomponent":"libvirt","thread":"68","timestamp":"2019-10-08T18:05:57.567000Z"}`

jeniferh:
Ok, on the multiqueue issue, turns out it is based on cpu requests:
`The  "" is invalid: spec.template.spec.domain.devices.networkInterfaceMultiqueue: MultiQueue for network interfaces can't be used without specifying CPU requests or limits.`
By default kubevirt will give the VM 100m cpu requests so we were not seeing that error however that will not work for multiqueue since ethtool returns "Cannot set device channel parameters: Invalid argument". Once you do set cpu requests there is no need for the ethtool step, the combined queues in the guest should be updated to that value automatically.  We would want to calculate cpu requests to be equal to total vcpus.

I haven't been able to reproduce the Client VM Error state outside of ripsaw so not sure yet if that is a different issue..

jeniferh:
The client failure is not a ripsaw problem, Kubevirt has a netmultiqueue bug: https://github.com/kubevirt/kubevirt/issues/2798

But ripsaw will need to set cpu requests (=total vcpus) so that netmultiqueue will eventually work properly.

README.md
```diff
@@ -5,18 +5,18 @@ a performance baseline of Kubernetes cluster on your provider.
...
-| [Iperf3](docs/iperf3.md)       | Network Performance    | Working            |  Used, default : 30second  |
-| [fio](docs/fio_distributed.md) | Storage IO             | Working            |  Used, default : 30second  |
-| [Sysbench](docs/sysbench.md)   | System Performance     | Working            |  Used, default : 30second  |
-| [YCSB](docs/ycsb.md)           | Database Performance   | Working            |  Not used                  |
-| [Byowl](docs/byowl.md)         | User defined workload  | Working            |  Used, default : 30second  |
-| [Pgbench](docs/pgbench.md)     | Postgres Performance   | Working            |  Not used                  |
-| [Smallfile](docs/smallfile.md) | Storage IO Performance | Working            |  Used, default : 30second  |
-| [fs-drift](docs/fs-drift.md)   | Storage IO Longevity   | Working            |  Not used                  |
-| [hammerdb](docs/hammerdb.md)   | Database Performance   | Working            |  Not used                  |
+| Workload                       | Use                    | Status in Operator | Reconciliation usage       | VM support (kubevirt) |
```
jtaleric: s/VM/Kubevirt ?

jtaleric:
Small nit -- not enough to block.

jeniferh:
Hah, ok forget about the whole cpu request thing.. now its based on vcpus: https://github.com/kubevirt/kubevirt/pull/2783

Will re test once the next Kubevirt release is out..

aakarshg:
> Hah, ok forget about the whole cpu request thing.. now its based on vcpus: [kubevirt/kubevirt#2783](https://github.com/kubevirt/kubevirt/pull/2783)
> 
> Will re test once the next Kubevirt release is out..

still waiting on newer kubevirt release https://github.com/kubevirt/kubevirt/releases ( last release was 22 days back )

jeniferh:
Now with Kubevirt v0.23.0 I can successfully run w/ multiqueue enabled, both with bridge and masquerade mode. The num queues is now automatically set based on the number of vcpus, so you can remove the ethtool step, in fact it complains because the VM doesn't have that package:

`[    8.289247] cloud-init[723]: /var/lib/cloud/instance/scripts/runcmd: line 2: ethtool: command not found`

but if you install it you can see it already has queues set to num vcpus, ex:
```
[fedora@uperf-client-10 ~]$ ethtool -L eth0
no channel parameters changed.
current values: rx 0 tx 0 other 0 combined 5
```

aakarshg:
> Now with Kubevirt v0.23.0 I can successfully run w/ multiqueue enabled, both with bridge and masquerade mode. The num queues is now automatically set based on the number of vcpus, so you can remove the ethtool step, in fact it complains because the VM doesn't have that package:
> 
> `[ 8.289247] cloud-init[723]: /var/lib/cloud/instance/scripts/runcmd: line 2: ethtool: command not found`
> 
> but if you install it you can see it already has queues set to num vcpus, ex:
> 
> ```
> [fedora@uperf-client-10 ~]$ ethtool -L eth0
> no channel parameters changed.
> current values: rx 0 tx 0 other 0 combined 5
> ```

As mentioned in slack channel, for the image we're using by default ( provided by kubevirt ) we probably won't need to do that, but for a custom container disk image, we'll have to so edited documents and mentioned that.

jeniferh:
LGTM !

roles/uperf-bench/templates/configmap_script.yml.j2
```diff
@@ -0,0 +1,43 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: uperf-run-script-{{item.status.interfaces[0].ipAddress}}-{{ trunc_uuid }}
+  namespace: '{{ operator_namespace }}'
+data:
+  run_script.sh : |
+    export h={{item.status.interfaces[0].ipAddress}}
```
jtaleric: Add `export kind=[container/vm]` so we can tell the difference between. We would need a SNAFU change too.

aakarshg: i'm using the resourcetype argument over environment variable for this.

dry923:
Is it possible to also run CI on the vm test? The test cr only contains the pod kind so any changes to the vm would never get vetted through CI. Otherwise looks good.

PR #222, opened 2019-09-17
move to using podman

aakarshg:


aakarshg:
Done @jtaleric give it a nod and merge please

PR #224, opened 2019-09-17
use run_snafu.py for fs-drift

bengland2:
do not merge or test this PR without snafu PR 51.
It simply calls run_snafu.  I removed some cruft that was no longer needed.

### DIFF
--- a/resources/crds/ripsaw_v1alpha1_fs_drift_cr.yaml
+++ b/resources/crds/ripsaw_v1alpha1_fs_drift_cr.yaml
@@ -14,8 +14,7 @@ spec:
   workload:
     name: fs-drift
     args:
-      # increase samples to 3 to get % deviation in results
-      # NOT IMPLEMENTED YET
+      # increase samples to 3 to measure variance in results
       samples: 1
       # specify as many pods as you like
       worker_pods: 2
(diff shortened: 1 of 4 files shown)


bengland2:
left out the ES env. vars in workload.yml.j2, putting that in, DNM yet.   Need a way to test with AWS.

bengland2:
This PR is ready to merge now.   This PR has been tested in AWS, but we haven't seen it deliver ES docs yet from AWS obviously, but as with smallfile we do see the environment variables being passed through from the CR to run_snafu.py.   The ES document generation was tested using a local docker image separately.

bengland2:
@aakarsh if you think it's ok, can you mark this OK to test?  I'll run the CI and resolve any problems.

bengland2:
do not merge, there are problems that minishift CI is reporting.  Wondering why minikube passed it, more later.    I don't like the test script, it doesn't tell you what went wrong, fixing that.

bengland2:
Should be ready for merge now.   tests/test_crs/valid_fs_drift.yaml wasn't including the environment variables or the samples parameter, I fixed workload.yml.j2 so it doesn't require the user to supply these (unless they want ES or multiple samples).   minikube CI passed on my laptop.

bengland2:
@aakarsh whaddya think?

tests/test_fs_drift.sh
```diff
@@ -5,8 +5,8 @@ source tests/common.sh
 
 function finish {
   echo "Cleaning up fs-drift"
-  kubectl delete -f tests/test_crs/valid_fs_drift.yaml
-  delete_operator
+  #kubectl delete -f tests/test_crs/valid_fs_drift.yaml
+  #delete_operator
```
aakarshg: what's the reason behind this change @bengland2 , also ccing @dry923 for thoughts.

resources/crds/ripsaw_v1alpha1_fs_drift_cr.yaml
```diff
@@ -4,11 +4,17 @@ metadata:
   name: example-benchmark
   namespace: my-ripsaw
 spec:
+  test_user: homer_simpson
+  # to separate this test run from everyone else's
+  clustername: aws-2009-09-10
+  # where elastic search is running
+  es_server: ec2-18-220-65-133.us-east-2.compute.amazonaws.com
+  es_port: 19020
+  es_index: ripsaw-smallfile
```
aakarshg: this needs to be updated too, just noticed this.

bengland2:
oh wait I just saw this, fixing now.

PR #225, opened 2019-09-18
Multus support

jtaleric:
Adding Multus Support for macvlan tested.

User must create port per pod, multi-pod not supported.

### DIFF
--- a/docs/uperf.md
+++ b/docs/uperf.md
@@ -22,6 +22,8 @@ spec:
       pin: false
       pin_server: "node-0"
       pin_client: "node-1"
+      multus:
+        enabled: false
       samples: 1
       pair: 1
       test_types:
@@ -53,7 +55,11 @@ $ oc adm policy add-scc-to-user privileged -z benchmark-operator
 
 `pin_client` what node to pin the client pod to.
 
+`multus[1]` Configure our pods to use multus.
+
 `samples` how many times to run the tests. For example
+
+[1] https://github.com/intel/multus-cni/tree/master/examples
 
 ```yaml
       samples: 3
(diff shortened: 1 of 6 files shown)


docs/uperf.md
```diff
@@ -53,6 +55,8 @@ $ oc adm policy add-scc-to-user privileged -z benchmark-operator
 
 `pin_client` what node to pin the client pod to.
 
+`multus` Configure our pods to use multus.
```
dry923: Minor nit but can we throw a link to a multus doc page or something in case someone has no idea what it is?

docs/uperf.md
```diff
@@ -53,8 +55,12 @@ $ oc adm policy add-scc-to-user privileged -z benchmark-operator
 
 `pin_client` what node to pin the client pod to.
 
+`multus[1]` Configure our pods to use multus.
```
aakarshg: Not be nitpicky, but can we add experimental tag or something to this, we are not gonna test it in CI as of now, so would be better if we could probably add that.

jtaleric:
I disagree about the experimental tag, as it has been tested... Just because there is no *CI* doesn't constitute experimental. If having a Multus CI is a gap, we need to get it filled.

aakarshg:
> I disagree about the experimental tag, as it has been tested... Just because there is no _CI_ doesn't constitute experimental. If having a Multus CI is a gap, we need to get it filled.

Tested on a specific environment(s), so the confidence is low. And also because of lack of CI, any subsequent prs could cause a failure which would go unnoticed until someone would raise an issue after running with multus. Having an experimental tag would help the user to understand that if it doesn't run as expected, it wouldnt give people an excuse to assume Ripsaw is broken. 

Again not a blocker, so gonna approve it.

jtaleric:
> > I disagree about the experimental tag, as it has been tested... Just because there is no _CI_ doesn't constitute experimental. If having a Multus CI is a gap, we need to get it filled.
> 
> Tested on a specific environment(s), so the confidence is low. And also because of lack of CI, any subsequent prs could cause a failure which would go unnoticed until someone would raise an issue after running with multus. Having an experimental tag would help the user to understand that if it doesn't run as expected, it wouldnt give people an excuse to assume Ripsaw is broken.

Ripsaw doesn't interact with Multus persay. If the user provides the Multus annotations, Ripsaw will simply query to see if there are IPs allocated to it (via the annotations).

If Ripsaw created networks and did things with Multus -- ie deployed XYZ-cni, I would agree, however that isn't what this PR does. 

> 
> Again not a blocker, so gonna approve it.

PR #231, opened 2019-09-24
smallfile_cli.py is in /usr/local/bin, run the script directly

vasukulkarni:
smallfile_cli.py is in /usr/local/bin, when run with python the PATH variable doesn't include /usr/local/bin which causes the job to fail. Invoke the
cli directly instead.

Fixes: https://github.com/cloud-bulldozer/ripsaw/issues/230

Signed-off-by: Vasu Kulkarni <vasu@redhat.com>


jtaleric:
Thanks @vasukulkarni!  Great find! We need to work with @bengland2 on having proper exit code status.

PR #237, opened 2019-10-21
Adds functionality to trigger (via snafu) dropping of Ceph OSD caches

dustinblack:
A sample file is included in `resources` to allow the user to manually
deply a privileged pod in the rook-ceph namespace with the capabilities
of listening on an HTTP port for a request to drop caches and then
translating that to a set of direct ceph commands. The fio CR file is
updated to accpet the options to trigger the cache drop, and some minimal
changes have been made to the fio role to pass the option to snafu.
Documentation has been updated to explain the requirements.


dustinblack:
> Overall happy with this, one question is the `rook_ceph_drop_cache_pod.yaml ` should this YAML also be in bohica? I am fine with this change either way, just curious what others think.

IMO we are already causing ourselves ridiculous amounts of pain with all of the inter-dependent code segregation, and my preference is to stop proliferating that behavior. That said, I don't care enough to argue against if it is deemed that the YAML isn't a ripsaw resource and should otherwise be placed in $other_project.

dustinblack:
> > > Overall happy with this, one question is the `rook_ceph_drop_cache_pod.yaml ` should this YAML also be in bohica? I am fine with this change either way, just curious what others think.
> > 
> > 
> > IMO we are already causing ourselves ridiculous amounts of pain with all of the inter-dependent code segregation, and my preference is to stop proliferating that behavior. That said, I don't care enough to argue against if it is deemed that the YAML isn't a ripsaw resource and should otherwise be placed in $other_project.
> 
> Is the issue that we have Ripsaw / SNAFU and now Bohica? Is your argument we should just have Ripsaw / SNAFU ?

This discussion has gone deeper in-channel, but to close the loop here, I don't mind the move of the yaml to snafu, and I would prefer it go there rather than bohica.

dustinblack:
Looping back on this PR since @aakarshg brought it up in conversation. The work was already done here to exclude the cache dropping privileged pod spin-up from ripsaw, replacing that automation with documentation. The open question is about whether the example yaml provided to the user for deploying that pod manually be included here in the ripsaw role or pointed to elsewhere in snafu or bohica. It seems silly that this PR has been sitting for weeks untouched on this point, so let's just make a decision and move forward.

IMO, the yaml should stay here.

docs/fio_distributed.md
```diff
@@ -173,6 +177,29 @@ jobfile configmap. These options will therefore override the global values for a
...
+[resources/rook_ceph_drop_cache_pod.yaml](../resources/rook_ceph_drop_cache_pod.yaml) file, which will
+deploy a pod with the correct permissions and tools, as well as running a simple HTTP listener to trigger
+the cache drop by URL.
+**You must deploy this privileged pod in order for the drop caches requests in the workload to function.**
+
+```bash
+kubectl apply -f resources/rook_ceph_drop_cache_pod.yaml
+```
+
+*Note: If Ceph is in a namespace other than `rook-ceph` you will need to modify the provided YAML accordingly.*
```
jtaleric: Which YAML do we update for the different ns? Looking at the provided YAML it is hard set to `rook-ceph`? If the YAML needs updating, the user running this work will have to rebuild the operator. Versus if you template the namespace and allow the user to override from the CR.

PR #243, opened 2019-11-04
move away from deprecated apis

### LINKED ISSUE #226: Move away from deprecated API groups
Kubernetes 1.16 was released recently and the apis that were deprecated include `deployments` in `extensions/v1beta1` [1], which is currently being used for pgbench test script [2].

We should look at replacing it with `apps/v1` api.

[1] https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/
[2] https://github.com/cloud-bulldozer/ripsaw/blob/master/tests/test_pgbench.sh#L42

aakarshg:
fixes #226

### DIFF
--- a/tests/test_pgbench.sh
+++ b/tests/test_pgbench.sh
@@ -39,7 +39,7 @@ data:
   PGDATA: /var/lib/postgresql/data/pgdata
 EOF
 cat << EOF | kubectl apply -f -
-apiVersion: extensions/v1beta1
+apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: postgres


jtaleric:
@aakarshg Need to figure out why things are failing before we review?

aakarshg:
CI is not using kube 1.16 ;) @jtaleric

jtaleric:
> CI is not using kube 1.16 ;) @jtaleric

OK, I suggest you add a check versus breaking functionality. if version == 1.16 (use new API) else (old api)

rsevilla87:
tests/mssql.yaml and  tests/test_ycsb.sh also have references to apps/v1beta1. This API is also deprecated in k8s-1.16 according to https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.15.md#deprecations-and-removals.

bengland2:
CI should now be using kube 1.16.

bengland2:
whine alert ;-) why don't logs of offending workload pods get dumped into jenkins log also?  It seemingly only dumps pod logs for things that have nothing to do with the problem.  So all the evidence is gone by the time you go and look?   Ideally you shouldn't have to run the test on your laptop to figure it out, this is time consuming.

jtaleric:
> whine alert ;-) why don't logs of offending workload pods get dumped into jenkins log also? It seemingly only dumps pod logs for things that have nothing to do with the problem. So all the evidence is gone by the time you go and look? Ideally you shouldn't have to run the test on your laptop to figure it out, this is time consuming.

Please if we plan on to warehouse the log output, lets not use jenkins, lets push to a public source, that way non RH people can see the output!

aakarshg:
closing in favor of #278

Issue #169, opened 2019-07-26
[YCSB] make the load step optional


aakarshg:
Following merge of #149, ycsb load phase is not optional. It can be useful while rerunning same ycsb workloads, for testing/debugging purposes.

PR #247, opened 2019-11-11
Adding functionality to selectively run/load ycsb

### LINKED ISSUE #182: only initialization or only run workload flags
for database workloads, like pgbench, sysbench or YCSB, it will be a good idea to allow a cr to only run the initialization part or just the actual workload.
The reason is that loading of the database can take long time, especially on large databases or on many small databases, and you don't want to go through this process on every run.

aakarshg:
Fixes #182, #169

### DIFF
--- a/docs/ycsb.md
+++ b/docs/ycsb.md
@@ -50,15 +50,28 @@ Note: We currently support running either workloade or workloadd( and not in the
 Reason: Similar to reason provided by [YCSB documentation](https://github.com/brianfrankcooper/YCSB/wiki/Core-Workloads#running-the-workloads), we'd need to drop the database
         after running either of the workloadd/workloade. This would mean operator to have knowledge of interaction with the said DB, which was outside the scope of this operator.
 
-`options_run` these are the options passed to ycsb binary while running the workload, this is where the target database and options are passed.
-Please read YCSB documentation on the necessary args required for the particular database. And note, the url or API needs to be accessible from the ycsb pod.
-
 The following options are optional:
 
-`loaded` This can be optionally set to true, if you have already loaded the database.
+`load` This can be optionally set to false, if you'd like to just run the workload and have already loaded previously.
 
-`options_load` This needs to be set if `loaded` is not defined or set to false, and like in the case of `options_run` this needs to be configured properly,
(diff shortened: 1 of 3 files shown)


playbook.yml
```diff
@@ -70,14 +70,14 @@
           name: "load-ycsb"
         vars:
           ycsb: "{{ workload.args }}"
-        when: workload.name == "ycsb" and workload.args.workloads | length > 0
+        when: workload.name == "ycsb" and workload.args.workloads | length > 0 and load_ycsb
```
dry923: does this really need to be a variable defined in group_vars/all.yml? Feels like it could just be 
```workload.name == "ycsb" and workload.args.workloads | length >  0 and workload.args.load | default('True')```

Same for the run_ycsb var below

File: docs/ycsb.md [hunk 2]
```diff
@@ -50,16 +50,29 @@ Note: We currently support running either workloade or workloadd( and not in the
...
 
-`options_run` these are the options passed to ycsb binary while running the workload, this is where the target database and options are passed.
-Please read YCSB documentation on the necessary args required for the particular database. And note, the url or API needs to be accessible from the ycsb pod.
-
 The following options are optional:
 
-`loaded` This can be optionally set to true, if you have already loaded the database.
+`load` This can be optionally set to false, if you'd like to just run the workload and have already loaded previously.
+
+`options_load` This needs to be set if `load` is not defined or set to True, these are the options passed to ycsb binary while running the workload, this is where the target database and options are passed.
```
dry923: Awkward wording...maybe: "...This needs to be set if `load` is set to true (the default)..."

docs/ycsb.md
```diff
@@ -50,16 +50,29 @@ Note: We currently support running either workloade or workloadd( and not in the
...
+Please read YCSB documentation on the necessary args required for the particular database. And note, the url or API needs to be accessible from the ycsb pod.
+
+`run` This can be optionally set to false, if you'd like to just load the database.
 
-`options_load` This needs to be set if `loaded` is not defined or set to false, and like in the case of `options_run` this needs to be configured properly,
+`options_run` needs to be set if `run` isn't set or set to True, and like in the case of `options_load` this needs to be configured properly,
```
dry923: Same wording comment as above

PR #249, opened 2019-11-18
so run_snafu.py can work in CI

bengland2:
This makes the 2 CRs conform to the snafu README.md documentation for how to point to elastic search.

### DIFF
--- a/resources/crds/ripsaw_v1alpha1_fs_drift_cr.yaml
+++ b/resources/crds/ripsaw_v1alpha1_fs_drift_cr.yaml
@@ -9,7 +9,7 @@ spec:
   clustername: aws-2009-09-10
   # where elastic search is running
   elasticsearch:
-    server: marquez.perf.lab.eng.rdu2.redhat.com
+    server: my.elasticsearch.server
     port: 9200
   workload:
     name: fs-drift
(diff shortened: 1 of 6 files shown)


dry923:
@bengland2 is it possible to instead update the templates to accept elasticsearch.server etc like the other benchmarks instead? That would also make it consistent with what the metadata collection expects as well.

bengland2:
@dry923 ok, but then we should update the snafu documentation to reflect that.   @acalhounRH @aakarshg any opinions?  Also, what happens to es_index?   Does it make sense to have something like this example:
```
elasticsearch:
  host: marquez
  port: 9200
  index: ripsaw-fs-drift-results
  test_user: homerSimpson
  clustername: springfield
```
instead of what's documented now:
```
es_host: marquez
es_port: 9200
es_index: ripsaw-fs-drift-results
test_user: homerSimpson
clustername: springfield
```
The advantage of the elasticsearch: format is that if someone doesn't want elasticsearch they then know what to leave out.

bengland2:
This PR depends on snafu PR 99, merge this when you merge that one or things will break.  I tested both of them together for smallfile and fs-drift, and it does run successfully with tests/test_{fs_drift,smallfile}.sh but not yet tested for fio_distributed.   Please tell me if this is what you want, while I try to test fio_distributed this way.  Do we want to post CI results to an ES server?

PR #251, opened 2019-11-25
Add client_node and server_node metadata

mohit-sheth:
This will export client_node and server_node for indexing when the 'pin' is set to true. If it is set to false it will export the string "UNSET" @jtaleric

### DIFF
--- a/roles/uperf-bench/templates/workload.yml.j2
+++ b/roles/uperf-bench/templates/workload.yml.j2
@@ -37,7 +37,7 @@ spec:
               topologyKey: kubernetes.io/hostname
       containers:
       - name: benchmark
-        image: "quay.io/cloud-bulldozer/uperf:latest"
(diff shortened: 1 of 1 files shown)


roles/uperf-bench/templates/workload.yml.j2
```diff
@@ -61,6 +61,13 @@ spec:
 {% endif %}
 {% if test_user is defined %}
              export test_user={{test_user}};
+{% endif %}
+{% if uperf.pin is sameas true %}
```
aakarshg: This is fine for now, but moving forward we should be able to extract the node of the client and server through redis somehow.

jtaleric: How do you propose we do this? As this would have to be queried from kube and provided to SNAFU via Redis.... There is not a *clean* way to provide this unless the user pins. Happy to hear what you think of.

PR #253, opened 2019-11-27
metadata docs update

dry923:
Minor updates to metadata documentation


dry923:
> LGTM - Should we include the backpack_role as a resource for people to use?

We can, I can drop the yaml into the resources folder if thats the preference vs having it doc'd.

PR #255, opened 2019-12-04
Add UUID to the label

### LINKED ISSUE #256: Added the UUID label to uperf service template
To match with the server.j2.yml - https://github.com/cloud-bulldozer/ripsaw/blob/d24383b53d27b0106b4f5b2114c96aa9929f0a9b/roles/uperf-bench/templates/service.yml.j2#L8 needs to be replaced with `app : uperf-bench-server-{{item}}-{{ trunc_uuid }}`

mohit-sheth:
This change will sync up with the change in https://github.com/cloud-bulldozer/ripsaw/blob/d24383b53d27b0106b4f5b2114c96aa9929f0a9b/roles/uperf-bench/templates/server.yml.j2#L8
@aakarshg 

Closes #256


aakarshg:
ok to test

PR #259, opened 2019-12-12
Pass number of clients and redis host to smallfile jobs

rsevilla87:
Required to fix https://github.com/cloud-bulldozer/snafu/issues/98
These variables are required to the smallfile snafu wrapper, so that this will be able to synchronize smallfile status using redis channel subscriptions

### DIFF
--- a/roles/smallfile-bench/tasks/main.yml
+++ b/roles/smallfile-bench/tasks/main.yml
@@ -7,7 +7,8 @@
     namespace: '{{ operator_namespace }}'
   register: resource_state
 
-- k8s_status:
+- name: Update current state
+  k8s_status:
     api_version: ripsaw.cloudbulldozer.io/v1alpha1
     kind: Benchmark
     name: "{{ meta.name }}"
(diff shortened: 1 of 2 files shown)


bengland2:
One thing I would love this PR to do is handle the case where multiple operations and samples are being run in smallfile - could we synchronize start of each of these with redis?   I don't think this version of it does that, but redis could do that for us if we had a different redis key for each sample and op type, right?

aakarshg:
So if I understand it correctly @bengland2 snafu will invoke smallfile with a list of operations, so if we wanted to synchroniz start of the atomic operations across pods, that logic will need to be inside smallfile ? The only way we can have the logic stay in smallfile would be if smallfile is being invoked once per operation, which doesn't seem to be the case rn?

bengland2:
It finally sunk in - this change passes the number of clients so that the snafu wrapper for smallfile knows how many clients to wait for before each op.     A separate commit to snafu will use that information to invoke the redis subscriber to actually do the wait.   So this commit by itself is a no-op, because it just sets extra environment variables but run_snafu.py is not interpreting those yet.   Thanks Raul for getting this going.

rsevilla87:
Snafu commit https://github.com/cloud-bulldozer/snafu/pull/102

PR #260, opened 2019-12-12
allow container image location and account to be overridden by user

bengland2:
This lets any user run the CI using their own account, since
they do not have privs to use the CI account.


aakarshg:
Hey @bengland2 this is a very valuable addition, can you also please update the docs to let users know how they can run their tests before submitting a pr?

bengland2:
sure thing, as soon as I know it works.   But the problem is that I just found out that each ci_test.sh in each wrapper embeds the account name and container image location as well, so we need to make all of them use an environment variable also :-(   Not hard to do, though.   Will talk to Russ about it when he's available.

aakarshg:
@bengland2 I think you're confusing between snafu's test scripts and ripsaw's test scripts. This PR *should* work fine for ripsaw testing, and anyone who makes change to ripsaw's code can now run the test script :) For snafu, it's a completely different discussion.

bengland2:
@aakarsh right you are, I forgot to separate snafu from ripsaw.   The fixes to snafu ci/*/ci_test.sh are a separate commit to snafu to use those 2 RIPSAW_CI environment variables, I guess.   We'll see what Russ thinks.   But it looks like the change above did not break the CI.

bengland2:
So this isn't quite enough.    When using my own account, I had to make the benchmark-operator "public" in quay.io settings or else it couldn't load the image into minikube.    What do you suggest we do here @dry @aakarshg?  

Also, every wrapper's ci_test.sh has to be changed to use RIPSAW_CI_IMAGE_{LOCATION,ACCOUNT} environment variables.    Do you think it's reasonable for common.sh to construct the prefix of the URL for the wrappers, and then they just append their wrapper name to the prefix to get the URL?   I'll provide a commit that does this. 

At present the smallfile wrapper runs successfully up to the point where ci/check_es.py runs, and then it fails with "no module named 'elasticsearch'".

bengland2:
never mind that last paragraph, forgot to install python3-elasticsearch when I reinstalled my laptop with F31.

aakarshg:
> So this isn't quite enough. When using my own account, I had to make the benchmark-operator "public" in quay.io settings or else it couldn't load the image into minikube. What do you suggest we do here @dry @aakarshg?
> 
This is default behavior of quay.io, so just adding a note to make the images public will be enough :) 

> Also, every wrapper's ci_test.sh has to be changed to use RIPSAW_CI_IMAGE_{LOCATION,ACCOUNT} environment variables. Do you think it's reasonable for common.sh to construct the prefix of the URL for the wrappers, and then they just append their wrapper name to the prefix to get the URL? I'll provide a commit that does this.
> 

This is very much reasonable. 


Thanks for the work put into this, @bengland2 this is nice work.

jtaleric:
> > So this isn't quite enough. When using my own account, I had to make the benchmark-operator "public" in quay.io settings or else it couldn't load the image into minikube. What do you suggest we do here @dry @aakarshg?
> 
> This is default behavior of quay.io, so just adding a note to make the images public will be enough :) which you've rightfully done so at https://github.com/cloud-bulldozer/snafu/pull/104/files#diff-04c6e90faac2675aa89e2176d2eec7d8R158
> 
> > Also, every wrapper's ci_test.sh has to be changed to use RIPSAW_CI_IMAGE_{LOCATION,ACCOUNT} environment variables. Do you think it's reasonable for common.sh to construct the prefix of the URL for the wrappers, and then they just append their wrapper name to the prefix to get the URL? I'll provide a commit that does this.
> 
> This is very much reasonable.
> 
> Thanks for the work put into this, @bengland2 this is nice work.

+1 Here!

Still need to add a small document update to let people know this var exists and what it is used for.

bengland2:
the error in the logs is a python syntax error in run_snafu.py, but the line of code referenced does not show up in master branch now, I think aakarsh fixed it (not sure how it got in).  So I'm re-running...

bengland2:
I tried to reproduce the kni_jjb failure in an AWS OpenShift cluster, but this did not work - smallfile passed, for example.  So I really don't understand why run_snafu.py was obviously not updated in the kni_jjb case but was updated in the minikube_jjb test case.  Does test.sh remove old code and re-clone?  Checking kni_jjb result log.

bengland2:
The minikube log was tiny, a few hundred lines.   The kni log was > 840 KB in size, and it hung my browser trying to look at it.   Can anyone extract the kni_jjb console log to a file and get it to me?    It's really hard to see how this PR caused KNI to break but not minikube.

PR #262, opened 2019-12-15
Add service selector to uperf svc jobs

rsevilla87:
Uperf service targeted benchmark is missing pod selector, this PR includes one based on the pod's type label.

### DIFF
--- a/roles/uperf-bench/templates/service.yml.j2
+++ b/roles/uperf-bench/templates/service.yml.j2
@@ -8,6 +8,8 @@ metadata:
     app: uperf-bench-server-{{item}}-{{ trunc_uuid }}
     type: uperf-bench-server-{{ trunc_uuid }}
 spec:
+  selector:
+    type: uperf-bench-server-{{ trunc_uuid }}
   ports:
   - name: uperf
     port: 20000


rsevilla87:
> this affects uperf, but your patch name indicates it is for smallfile

Typo... already fixed

aakarshg:
This is a good fix, thanks for catching it @rsevilla87. Oddly, looking at https://kubernetes.io/docs/concepts/services-networking/service/#environment-variables when I ran uperf pod to svc without this change, I was still able to see the env variables correctly configured so I still don't know why the env variables are automatically set in all pods of the namespace.

aakarshg:
@mohit-sheth might be worth your time to list down your experience with pod to svc for uperf, I know we're discussing some issues you're having offline. Hopefully this will fix it?

rsevilla87:
@aakarshg The env variables are always configured in all pods of the namespace regardless of the service selector. Those variables by any process can be used to resolve the service hostname, ports, etc.
Without this commit uperf tests over k8s services wouldn't work properly, as there wouldn't any pod targeted by it. 
https://kubernetes.io/docs/concepts/services-networking/service/#defining-a-service

jtaleric:
@mohit-sheth @aakarshg Please review and approve for merge.

PR #263, opened 2019-12-17
Enable I/O tests to benchmark hostPath volumes

rsevilla87:
Doing I/O benchmarks directly on top of containers implies making those I/O operations on the container filesystem, usually OverlayFS. This PR enables some ripsaw benchmarks to perform those tests on hostPaths  devices in order to avoid the OverlayFS layer.
Accessing to hostPath volumes require running pod as privileged, so the role has been also granted usage to the hostaccess SCC

### DIFF
--- a/tests/common.sh
+++ b/tests/common.sh
@@ -31,6 +31,7 @@ function populate_test_list {
 }
 
 function wait_clean {
+  kubectl delete benchmark --all -n my-ripsaw
   kubectl delete all --all -n my-ripsaw
   for i in {1..30}; do
     if [ `kubectl get pods --namespace my-ripsaw | grep bench | wc -l` -ge 1 ]; then
@@ -110,9 +111,7 @@ function pod_count () {
 function apply_operator {
   operator_requirements
   kubectl apply -f resources/operator.yaml
-  ripsaw_pod=$(get_pod 'name=benchmark-operator' 300)
-  kubectl wait --for=condition=Initialized "pods/$ripsaw_pod" --namespace my-ripsaw --timeout=60s
-  kubectl wait --for=condition=Ready "pods/$ripsaw_pod" --namespace my-ripsaw --timeout=300s
+  kubectl wait --for=condition=available "deployment/benchmark-operator" -n my-ripsaw --timeout=300s
 }
 
 function delete_operator {
--- a/tests/test_fiod.sh
+++ b/tests/test_fiod.sh
@@ -3,36 +3,28 @@ set -xeEo pipefail
 
 source tests/common.sh
 
-function finish {
-  if [ $? -eq 1 ] && [ $ERRORED != "true" ]
-  then
-    error
-  fi
-
-  echo "Cleaning up Fio"
-  kubectl delete -f tests/test_crs/valid_fiod.yaml
-  delete_operator
-}
-
 trap error ERR
-trap finish EXIT
+trap wait_clean EXIT
 
 function functional_test_fio {
-  figlet $(basename $0)
   apply_operator
-  kubectl apply -f tests/test_crs/valid_fiod.yaml
+  test_name=$1
+  cr=$2
+  echo "Performing: ${test_name}"
+  kubectl apply -f ${cr}
   uuid=$(get_uuid 20)
-  
   wait_for_backpack $uuid
-
   pod_count "app=fio-benchmark-$uuid" 2 300
   fio_pod=$(get_pod "app=fiod-client-$uuid" 300)
-  wait_for "kubectl wait --for=condition=Initialized pods/$fio_pod --namespace my-ripsaw --timeout=200s" "200s" $fio_pod
-  wait_for "kubectl wait --for=condition=complete -l app=fiod-client-$uuid jobs --namespace my-ripsaw --timeout=500s" "500s" $fio_pod
-  sleep 30
+  wait_for "kubectl wait --for=condition=Initialized pods/$fio_pod -n my-ripsaw --timeout=200s" "200s" $fio_pod
+  wait_for "kubectl wait --for=condition=complete -l app=fiod-client-$uuid jobs -n my-ripsaw --timeout=500s" "500s" $fio_pod
   # ensuring the run has actually happened
-  kubectl logs "$fio_pod" --namespace my-ripsaw | grep "fio has successfully finished sample"
-  echo "Fio distributed test: Success"
+  kubectl logs "$fio_pod" -n my-ripsaw
+  kubectl logs "$fio_pod" -n my-ripsaw | grep "fio has successfully finished sample"
+  echo "${test_name} test: Success"
+  wait_clean
 }
 
-functional_test_fio
+figlet $(basename $0)
+functional_test_fio "Fio distributed" tests/test_crs/valid_fiod.yaml
+functional_test_fio "Fio hostpath distributed" tests/test_crs/valid_fiod_hostpath.yaml
--- a/tests/test_fs_drift.sh
+++ b/tests/test_fs_drift.sh
@@ -3,51 +3,38 @@ set -xeo pipefail
 
 source tests/common.sh
 
-function finish {
-  echo "Cleaning up fs-drift"
-  kubectl delete -f tests/test_crs/valid_fs_drift.yaml
-  delete_operator
-}
-
-trap finish EXIT
+trap error ERR
+trap wait_clean EXIT
 
 function functional_test_fs_drift {
-  figlet $(basename $0)
   apply_operator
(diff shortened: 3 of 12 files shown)


rsevilla87:
@aakarshg PTAL

aakarshg:
PTAL @ekuric @acalhounRH @dustinblack @bengland2 @jtaleric .

aakarshg:
ok to test

bengland2:
@rsevilla87   It could be much simpler.   Instead of defining hostVolume, which impacts the entire thing, just redefine cephVolume.   Then you *do not need* all those extra CI scripts.  If you are running CI, the storageclass will not be defined anyway, so it's ok to override it.  In hindsight I shouldn't have called it "cephVolume" because it really isn't anything to do with Ceph, it's just a filesystem somewhere that the benchmark can use.   Perhaps the error was to use that name, but please, please let's not do it this way.

tests/test_fs_drift.sh
```diff
@@ -33,21 +31,17 @@ function functional_test_fs_drift {
...
-    echo "fs-drift test: Success" 
-  else 
-    echo fs-drift test FAILURE 
-    exit 1 
-  fi
+  wait_for "kubectl wait --for=condition=Initialized pods/$fsdrift_pod -n my-ripsaw --timeout=200s" "200s" $fsdrift_pod
+  wait_for "kubectl wait --for=condition=complete -l app=fs-drift-benchmark-$uuid jobs -n my-ripsaw --timeout=100s" "200s" $fsdrift_pod
+  # Print logs and check status
+  kubectl logs "$fsdrift_pod" -n my-ripsaw
+  kubectl logs "$fsdrift_pod" -n my-ripsaw | grep "RUN STATUS"
```
bengland2: do we need to check status of this grep command and fail the test if grep does not find this?  Like:
```
kubectl logs "$fsdrift_pod" -n my-ripsaw | grep "RUN STATUS" || exit $NOTOK
```

rsevilla87: grep already returns 1 if there isn't any match, in addition all test scripts have a trap to capture ERROR signals I think it wouldn't be necessary.

rsevilla87:
This time looks fine @bengland2, what do you think?. I don't know why ycsb tests failed, it hasn't any relationship with this changes though.

bengland2:
we shouldn't let unrelated CI problems block it!   I'm having that problem as well with my own PRs, see ripsaw PR 269.   Looking at this PR code now that you've updated it...

tests/test_fiod.sh
```diff
@@ -4,35 +4,33 @@ set -xeEo pipefail
...
 function finish {
-  if [ $? -eq 1 ] && [ $ERRORED != "true" ]
-  then
-    error
-  fi
-
-  echo "Cleaning up Fio"
-  kubectl delete -f tests/test_crs/valid_fiod.yaml
-  delete_operator
+  kubectl delete all --all -n my-ripsaw
```
bengland2: why should this be specific to test_fiod?  and why "delete all --all" instead of just "delete namespace my-ripsaw"?   Just curious, I'm not going to argue about it.

tests/test_smallfile.sh
```diff
@@ -4,27 +4,20 @@ set -xeEo pipefail
...
 function finish {
-  if [ $? -eq 1 ] && [ $ERRORED != "true" ]
-  then
-    error
-  fi
-
-  echo "Cleaning up Smallfile"
-  kubectl delete -f tests/test_crs/valid_smallfile.yaml
-  delete_operator
+  kubectl delete all --all -n my-ripsaw
```
bengland2: same comment as with test_fiod.   Why was the delete_operator function a bad idea?

tests/test_fiod.sh
rsevilla87: If we delete the namespace here. In case of error, the CI would fail in the next retry, since the namespace is deleted/created outside out form the retry loop.
Look here:
https://github.com/cloud-bulldozer/ripsaw/blob/master/run_test.sh#L19-L56

tests/test_smallfile.sh
rsevilla87: You're right. It's easy to fix, so let's include the code change here.

tests/test_fs_drift.sh
bengland2: yeah I see that now, the set -xe... on line 1,  this comment is resolved.

tests/test_fiod.sh
bengland2: true, I'm ok with this as long as it deletes things in the right order, and finalizers should enforce that anyway.   Never mind ;-)

tests/common.sh
```diff
@@ -219,6 +217,7 @@ function error {
   kubectl -n my-ripsaw logs --tail=40 -l name=benchmark-operator -c benchmark-operator
   echo "Ansible sidecar Logs"
   kubectl -n my-ripsaw logs -l name=benchmark-operator -c ansible
+  kubectl -n my-ripsaw delete benchmark --all
```
dry923: We shouldn't have to do this here. It should be done when being cleaned up when it hits the finish function.

tests/test_fiod.sh
dry923: I'm making some changes to the namespace create/delete per https://github.com/cloud-bulldozer/ripsaw/pull/271 which adds a namespace delete to wait clean and creation one of the operator functions. I'd also say given that we should just use wait clean here and not just the straight delete all command.

tests/test_fiod.sh
```diff
@@ -4,35 +4,33 @@ set -xeEo pipefail
...
-  sleep 30
+  wait_for "kubectl wait --for=condition=Initialized pods/$fio_pod -n my-ripsaw --timeout=200s" "200s" $fio_pod
+  wait_for "kubectl wait --for=condition=complete -l app=fiod-client-$uuid jobs -n my-ripsaw --timeout=500s" "500s" $fio_pod
   # ensuring the run has actually happened
-  kubectl logs "$fio_pod" --namespace my-ripsaw | grep "fio has successfully finished sample"
-  echo "Fio distributed test: Success"
+  kubectl logs "$fio_pod" -n my-ripsaw
+  kubectl logs "$fio_pod" -n my-ripsaw | grep "fio has successfully finished sample"
+  echo "${test_name} test: Success"
+  kubectl delete -f ${cr}
```
dry923: This is redundant with wait_clean

tests/test_fs_drift.sh
```diff
@@ -4,21 +4,21 @@ set -xeo pipefail
 source tests/common.sh
 
 function finish {
-  echo "Cleaning up fs-drift"
-  kubectl delete -f tests/test_crs/valid_fs_drift.yaml
-  delete_operator
+  kubectl delete all --all -n my-ripsaw
```
dry923: Same comment as above re wait_clean. Also, lets keep the echo in.

tests/test_smallfile.sh
dry923: Same comment re wait_clean as before. Also, why are you removing the error if statement?

tests/test_smallfile.sh
```diff
@@ -41,11 +35,15 @@ function functional_test_smallfile {
...
-  sleep 30
-  # ensuring the run has actually happened
-  kubectl logs "$smallfile_pod" --namespace my-ripsaw | grep "RUN STATUS"
-  echo "Smallfile test: Success"
+  wait_for "kubectl wait --for=condition=complete -l app=smallfile-benchmark-$uuid jobs --namespace my-ripsaw --timeout=200s" "200s" $smallfile_pod
+  # Print logs and check status
+  kubectl logs "$smallfile_pod" -n my-ripsaw
+  kubectl logs "$smallfile_pod" -n my-ripsaw | grep "RUN STATUS"
+  echo "${test_name} test: Success"
+  kubectl delete -f ${cr}
```
dry923: Redundant with the below.

bengland2:
@dry923 @rsevilla87  IMHO it would be a good long-term goal is to minimize the size of */ci_test.sh scripts by moving code into ci/common.sh wherever possible,  But it doesn't all need to be dealt with in this PR.   At this point, whatever helps us drain the PR swamp at this point and reach consensus and merge is fine with me ;-)

tests/test_fiod.sh
rsevilla87: wait_clean does not remove Benchmark CRs

bengland2:
0 CI fails!!!!  Have I died and gone to heaven?

rsevilla87:
Do we need anything else to get this one merged? :)

PR #266, opened 2020-01-06
Fix to allow the exec_prerun parameters in the fio jobfile to work co…

dustinblack:
…rrectly. Probably also fixes other unknown bugs.


dustinblack:
When `exec_prerun` is used, the server outputs a text file in the `$pwd` where the `fio --server` command was executed. I tested, and the `--aux-path` parameter doesn't seem to affect this, so the only option is to first change to a directory where we have write access.

PR #270, opened 2020-01-14
Increasing the time to wait for backpack to complete

dry923:
The wait_for_backpack function would sometimes timeout due to the amount of time it would take to pull down the image. Particularly when run on multiple node clusters.

### DIFF
--- a/tests/common.sh
+++ b/tests/common.sh
@@ -227,7 +227,8 @@ function wait_for_backpack() {
   
   uuid=$1
   count=0
-  while [[ $count -lt 30 ]]
+  max_count=60
+  while [[ $count -lt $max_count ]]
   do
     if [[ `kubectl -n my-ripsaw get daemonsets backpack-$uuid` ]]
     then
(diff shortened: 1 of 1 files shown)


rsevilla87:
With this PR, I'm wondering the next. Do we need to run backpack in every workload test? Most of CI time is spent on it. Maybe we can create an specific job for backpack and get the rid of it from the remaining tests (i.e. collect_metadata: false)
With the above we could increase parallelism on the tests and optimize CI time.

@jtaleric @aakarshg

aakarshg:
> With this PR, I'm wondering the next. Do we need to run backpack in every workload test? Most of CI time is spent on it. Maybe we can create an specific job for backpack and get the rid of it from the remaining tests (i.e. collect_metadata: false)
> With the above we could increase parallelism on the tests and optimize CI time.
> 
> @jtaleric @aakarshg

The main reason for adding collect_metadata to the crs is because of the interaction of it with various workloads, we're hitting issues with ycsb alone a while back, because it was not using the reconciliation.

PR #271, opened 2020-01-14
Adding namespace deletion/creation to requirements

dry923:
The namespace was not getting cleaned up during ci tests. This would sometimes lead to things being stuck in pending states. This will now delete the namespace in wait_clean and also ensure its created in the operator_requirements.

### DIFF
--- a/run_test.sh
+++ b/run_test.sh
@@ -10,8 +10,6 @@ figlet $ci_test
 cd $ci_dir
 source tests/common.sh
 
-kubectl apply -f resources/namespace.yaml
-
 count=0
 start_time=`date`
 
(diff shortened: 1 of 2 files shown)


bengland2:
what about sql_server namespace cleanup for hammerdb?   Are there any other benchmark-specific namespaces?   We should at least document what is expected for cleanup?

dry923:
> what about sql_server namespace cleanup for hammerdb? Are there any other benchmark-specific namespaces? We should at least document what is expected for cleanup?

That's a fair point but I think that would live outside this change. It should be getting cleaned up as part of the test script in my opinion.

PR #273, opened 2020-01-16
Test multiple clients in parallel

rsevilla87:
Required for https://github.com/cloud-bulldozer/snafu/pull/102. @aakarshg


aakarshg:
ok to test

bengland2:
thank you!   multi-client smallfile is the only thing that makes sense for measuring throughput.   1 client is just not that useful  except from a latency perspective.

PR #274, opened 2020-01-17
Add pod anti-affinity rules to IO jobs

rsevilla87:
Ensure IO clients spread among different nodes

### DIFF
--- a/roles/fs-drift/templates/workload_job.yml.j2
+++ b/roles/fs-drift/templates/workload_job.yml.j2
@@ -12,6 +12,18 @@ spec:
       labels:
         app: fs-drift-benchmark-{{ trunc_uuid }}
     spec:
+      affinity:
+        podAntiAffinity:
+          preferredDuringSchedulingIgnoredDuringExecution:
+          - weight: 100
+            podAffinityTerm:
+              labelSelector:
+                matchExpressions:
+                - key: app
+                  operator: In
+                  values:
+                  - fs-drift-benchmark-{{ trunc_uuid }}
+              topologyKey: "kubernetes.io/hostname"
       restartPolicy: Never
       containers:
         - name: benchmark-server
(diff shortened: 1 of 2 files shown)


bengland2:
@rsevilla87  why is this necessary?  It may well be a good idea but wouldn't the benchmark run without anti-affinity?

aakarshg:
> @rsevilla87 why is this necessary? It may well be a good idea but wouldn't the benchmark run without anti-affinity?

That's interesting opinion too, if we can make it optional that'll be great so folks can either use it if they want or not use it as well.

rsevilla87:
So you're proposing to execute jobs with anti-affinity in a configurable way. I can't see the use case where one use wants to run several IO clients from the same node. Possibly I'm missing something here.

pd: FIO servers already run with pod anti-affinity

rsevilla87:
> if this is going to spread workload generator pods more evenly across physical hosts, then I'm all for it.

Indeed. This is the goal of this PR :)

aakarshg:
> So you're proposing to execute jobs with anti-affinity in a configurable way. I can't see the use case where one use wants to run several IO clients from the same node. Possibly I'm missing something here.
> 
> pd: FIO servers already run with pod anti-affinity

Its not the same node, rather same sets of node.. so for example during our scale test, we only wanted the pods to land on worker nodes ( not infra/master/workload) so in that case we had to rebuild the image essentially, but maybe that's out of the scope of this PR as I mentioned earlier. 

For now, if we can just add a small README update about how we use antipodaffinity to spread the pods that'll be great :) I'll open an issue for the nodeselector label work.

rsevilla87:
Exactly, nodeSelector is a different thing here. The scope of this PR is to evenly spread all workload pods across all nodes matched with the nodeSelector (which is currently static as you mentioned). I'll  add another commit to this PR to update the docs explaining this behavior.

bengland2:
@rsevilla87 @dry923  it's hammer time - all of them passed except hammerdb

PR #275, opened 2020-01-28
Support StatefulSet v1 spec in test_ycsb.sh wrapper

mffiedler:
StatefulSet no longer tolerates the apps/v1beta1 API group.   It has been v1 since 1.9, so this change should be safe.


aakarshg:
Fails on minikube :/

PR #276, opened 2020-01-30
Updating CRD api to version v1

dry923:
Updating the CRD api to v1. This will require changes to the crd file as well as changes to cr's using older api versions for kinds deployment and stateful sets.

Additional changes may need to be made on a benchmark to benchmark basis to account for the tighter restrictions around cr validation.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -1,18 +1,17 @@
 ## Installation
-This guide uses minishift version v1.33.0+ as the local Kubernetes cluster
-and quay.io for the public registry. We also test on minikube version v0.35+.
+This guide uses minikube version v1.5.2+ as the local Kubernetes cluster
+and quay.io for the public registry. We also test on crc version v1.7.0+.
 `kubectl` is used but can be substituted with `oc`.
 
 ### Supported versions
-* [OKD](https://www.okd.io/)
-  * Experimental: 3.11
 * [OpenShift® Container Platform](https://www.openshift.com/products/container-platform/)
-  * Fully supported: 4.0
-  * Experimental: 3.11
+  * Fully supported: 4.3
 * [kubernetes](https://kubernetes.io/)
-  * Experimental: 1.11-1.13
+  * Fully supported: 1.16.2
 
-Note: Experimental tag refers to some workloads that might be functioning
+Note: 
+* Experimental tag refers to some workloads that might be functioning
+* To use versions of Openshift and kubernetes prior to 4.3 and 1.16.2 respectively, please use version 0.0.2 of ripsaw
 
 ### Requirements
 <!---
--- a/resources/crds/ripsaw_v1alpha1_ripsaw_crd.yaml
+++ b/resources/crds/ripsaw_v1alpha1_ripsaw_crd.yaml
@@ -1,4 +1,4 @@
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   creationTimestamp: null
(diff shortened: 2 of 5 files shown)


tests/test_ycsb.sh
```diff
@@ -40,12 +40,15 @@ spec:
    role: mongo
 EOF
 cat << EOF | kubectl apply -f -
-apiVersion: apps/v1beta1
+apiVersion: apps/v1
```
aakarshg: this is a duplicate of https://github.com/cloud-bulldozer/ripsaw/pull/275.

dry923: yes, however I was updating them all and it fails if its not updated properly. So if we merge the other one first thats fine. If not then we'll catch them all in one shot.

dry923:
@bengland2 re fs_drift and smallfile; can we move the test_user and clustername variables into the workload definition section of the cr? Also, can es_index go into the elasticsearch object? The workload section is wildcarded and allows any variables to be set, however the top level spec section gets validated now and we would have to define them in the crd if they cannot be moved.
i.e. this:
```
spec:
  test_user: homer_simpson
  # to separate this test run from everyone else's
  clustername: test_ci
  # where elastic search is running
  elasticsearch:
    server: "marquez.perf.lab.eng.rdu2.redhat.com"
    port: 9200
  es_index: ripsaw-fs-drift
  metadata_collection: true
  workload:
    name: fs-drift
```
to this:
```
spec:
   # where elastic search is running
  elasticsearch:
    server: "marquez.perf.lab.eng.rdu2.redhat.com"
    port: 9200
    index: ripsaw-fs-drift
  metadata_collection: true
  workload:
    name: fs-drift
    test_user: homer_simpson
    # to separate this test run from everyone else's
    clustername: test_ci
```

bengland2:
@dry923  So I don't see why the clustername needs to be part of the workload definition,  the workload to me seems to have nothing to do with which cluster it ran on, which I think of as part of the configuration definition.  Why are we doing this?

aakarshg:
> @dry923 So I don't see why the clustername needs to be part of the workload definition, the workload to me seems to have nothing to do with which cluster it ran on, which I think of as part of the configuration definition. Why are we doing this?

We need an identifier to associate the results with which cluster this ran on..

bengland2:
Yes we do, but wasn't it already there?  Just in a different layer of YAML?  

Let's update a documentation file to describe what this format is if we haven't already.  I started that with PR 104, I think.  Make it really clear to benchmark developers what the format has to be.

aakarshg:
> Yes we do, but wasn't it already there? Just in a different layer of YAML?
> 
> Let's update a documentation file to describe what this format is if we haven't already. I started that with PR 104, I think. Make it really clear to benchmark developers what the format has to be.

Agreed @dry923 can you please update the documentation in this PR so that the others can follow with the strict layering ?

dry923:
> > Yes we do, but wasn't it already there? Just in a different layer of YAML?
> > Let's update a documentation file to describe what this format is if we haven't already. I started that with PR 104, I think. Make it really clear to benchmark developers what the format has to be.
> 
> Agreed @dry923 can you please update the documentation in this PR so that the others can follow with the strict layering ?

@aakarshg Definitely. There is going to be a nice chunk of doc updates coming. I put this up now so we can figure out some of the details

dry923:
Minikube and kni will need to be updated to a more recent version of kubernetes. The api that is currently installed does not have a v1 for the crd

dry923:
This is pending an api update to minikube and kni still

resources/crds/ripsaw_v1alpha1_ripsaw_crd.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
```
aakarshg, line 1: We'll need to track from which version of k8s and associated ocp is this valid

dry923: The v1 update was include in 1.16 of kubernetes which is 4.3 of OCP I believe

aakarshg: can you please add this to documentation? its very important that we point out that if they're using older k8s version then they can use v0.0.2 of ripsaw

docs/installation.md
```diff
@@ -7,10 +7,9 @@ and quay.io for the public registry. We also test on minikube version v0.35+.
 * [OKD](https://www.okd.io/)
   * Experimental: 3.11
 * [OpenShift® Container Platform](https://www.openshift.com/products/container-platform/)
-  * Fully supported: 4.0
-  * Experimental: 3.11
+  * Fully supported: 4.3
 * [kubernetes](https://kubernetes.io/)
-  * Experimental: 1.11-1.13
+  * Fully supported: 1.16.2
```
dry923, line 12: @aakarshg Updated install docs to specify version

docs/installation.md
```diff
@@ -7,12 +7,13 @@ and quay.io for the public registry. We also test on minikube version v0.35+.
 * [OKD](https://www.okd.io/)
   * Experimental: 3.11
```
aakarshg, line 8: This will also be need to be updated or removed altogether as the current master wont work..

dry923: Updated. Also modified the intro paragraph which discusses minikube/shift/etc versions

PR #277, opened 2020-01-31
Adding cleanup of ripsaw crd

dry923:
Making sure we cleanup the ripsaw crd during wait_clean


dry923:
So looking at the CI and thinking about it for a minute this isn't going to work this simply. The CI runs multiple tests at once so it is removing the crd at times when it shouldn't. It also appears that it is in the resource cleanup job which is run at the start of every run so this PR is not needed. Closing this.

PR #278, opened 2020-02-03
Updating Deployment to apps/v1

dry923:
Updating to apps/v1 and adding the now required selector


dry923:
NOTE: This does not update the mssql Deployment since that will be taken care of with https://github.com/cloud-bulldozer/ripsaw/pull/267 once it is merged

PR #279, opened 2020-02-03
Adding a check to the namespace deletion

dry923:
There are times that we run a wait_clean after a wait_clean. When the namespace is already deleted this can throw an error. This adds a check for the existence of the namespace first.

### DIFF
--- a/tests/common.sh
+++ b/tests/common.sh
@@ -40,7 +40,9 @@ function wait_clean {
       break
     fi
   done
-  kubectl delete namespace my-ripsaw
+  if [[ `kubectl get namespace my-ripsaw` ]]; then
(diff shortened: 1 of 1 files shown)


aakarshg:
@rsevilla87 ptal :)

rsevilla87:
Yeah, this is one of the changes we needed :)
LGTM

rsevilla87:
> LGTM but do we want to add `--wait=true` ?

I think by default is true

```
$ kubectl delete -h | grep wait=
      --wait=true: If true, wait for resources to be gone before returning. This waits for finalizers.
```

jtaleric:
Ack. Again. LGTM.

PR #280, opened 2020-02-04
Changing trap wait_clean to trap finish to fix snafu ci error

dry923:
Snafu ci expects there to be a trap finish function that it can remove to allow the existing pods to run longer so it can get the uuid from them.

### DIFF
--- a/tests/test_fiod.sh
+++ b/tests/test_fiod.sh
@@ -3,8 +3,18 @@ set -xeEo pipefail
 
 source tests/common.sh
 
+function finish {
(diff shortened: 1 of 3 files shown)


rsevilla87:
I'm agree to have this one as a temporally solution for CI problems on Snafu. Though as I discussed with @aakarshg we should start working on get the rid of ripsaw dependencies in Snafu ci.

aakarshg:
> I'm agree to have this one as a temporally solution for CI problems on Snafu. Though as I discussed with @aakarshg we should start working on get the rid of ripsaw dependencies in Snafu ci.

Agreed lets hope this works at least

PR #281, opened 2020-02-06
Backpack will now properly fail the ci test if it does not launch

dry923:
Previously if backpack failed to launch during a ci test it would continue anyway. This will now have it fail and exit properly.

### DIFF
--- a/tests/common.sh
+++ b/tests/common.sh
@@ -240,13 +240,16 @@ function wait_for_backpack() {
       if [[ $desired -eq $ready ]]
       then
         echo "Backpack complete. Starting benchmark"
-        count=$max_count
+        break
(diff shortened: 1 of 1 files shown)


tests/common.sh
```diff
@@ -240,13 +240,15 @@ function wait_for_backpack() {
...
-        count=$max_count
+        break
       fi
     fi
+    count=$((count + 1))
     if [[ $count -ne $max_count ]]
     then
       sleep 5
-      count=$((count + 1))
+    else
```
jtaleric: Want to provide a log message with the exit?

dry923: Good call. Added.

PR #282, opened 2020-02-06
Add continue/abort functionality if a failure occurs

mohit-sheth:
@jtaleric @aakarshg 
If a failure occurs, this commit will give us the ability to toggle between continuing with the pending tests or aborting them. We can also toggle between collecting/not_collecting the status code from redis. This PR is in conjunction with snafu PR 136 (https://github.com/cloud-bulldozer/snafu/pull/136)

### DIFF
--- a/docs/uperf.md
+++ b/docs/uperf.md
@@ -36,6 +36,8 @@ spec:
       nthrs:
         - 1
       runtime: 30
+      continue_on_failure: "true"
+      redis_collection: "true"
 ```
 
 `serviceip` will place the uperf server behind a K8s [Service](https://kubernetes.io/docs/concepts/services-networking/service/)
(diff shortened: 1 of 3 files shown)


dry923:
Can these be values that are defaulted true (or false) and only overridden in a cr if a user desires? This way its implicitly defined in the cr to always continue on failure and will not need to be added unless a user wants to change the way it runs. Similar to the way metadata_collection is defaulted to false but can be set to true, etc

mohit-sheth:
> Can these be values that are defaulted true (or false) and only overridden in a cr if a user desires? This way its implicitly defined in the cr to always continue on failure and will not need to be added unless a user wants to change the way it runs. Similar to the way metadata_collection is defaulted to false but can be set to true, etc

yeah, can do that by initializing with the default value in vars/ @dry923

mohit-sheth:
closing in favor of https://github.com/cloud-bulldozer/ripsaw/pull/289

PR #285, opened 2020-02-19
Adding option to run backpack as an initcontainer

dry923:
This PR adds the default ability to run backpack as init containers instead of as a daemon set. 

There are a few notable differences between the DaemonSet and targeted options:

- First, and most important, the DaemonSet will run on ALL nodes of the cluster.
This means that if you have 200 nodes it will collect data from all the nodes
even if your relevant pod(s) are only running on a subset of nodes. You will end
up with more data than you need and slow down the start of your workload.

- If running targeted and with a service account (more on that below) the
entire workload will run as that service account. That is because service accounts
are done at a level above the container definition and can only be applied once.

- When run as a daemonset the backpack pods will not complete/terminate 
until you delete your benchmark. This is done to allow additional collections to 
be done as an ad-hoc basis as well.

- When running as a targeted init container the benchmark metadata status will be
set to "Collecting" however it will never be marked as completed as that would require
additional logic in the workloads which is out of scope.

- Finally, running as a targeted init container requires no additions to the workload
template. This may be preferable in certain situations.

Additionally, I have only added the backpack collection to workload/client template files. This may need adjustment. Also, as ycsb does not have a container template I enabled the classic backpack style for its testing.

### DIFF
--- a/docs/metadata.md
+++ b/docs/metadata.md
@@ -2,6 +2,9 @@
 
 How To:
 * [How it Works](#how-it-works)
+* [Targeted Init Containers](#targeted-init-containers)
+* [DaemonSet](#daemonset)
+* [Differences between targeted and daemonset](#Differences-between-targeted-and-daemonset)
 * [Enable Collection](#enable-collection)
 * [Additional Options](#additional-options)
 * [Running Additional Collections](#running-additional-collections)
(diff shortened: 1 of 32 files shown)


dry923:
Updated to properly wait for server pods to be running before clients. With this I also added/fixed the reconciliation for iperf and uperf

playbook.yml
```diff
@@ -43,6 +43,16 @@
       
       - include_role:
           name: backpack
+        when: not metadata_targeted | default('true') | bool
```
aakarshg: So if I can grok this ansible check correctly, this would mean the role would be included by default? Wouldn't we want the default to be false ?

jtaleric: I am also not tracking this... `not metadata_targeted` would make me think false, but if it isn't set, it defaults to `true`?

docs/metadata.md
```diff
@@ -11,16 +14,80 @@ How To:
 The metadata collection is done through the use of [Stockpile](https://github.com/cloud-bulldozer/stockpile), [Backpack](https://github.com/cloud-bulldozer/backpack) and [Scribe](https://github.com/cloud-bulldozer/scribe).
 The data is then uploaded to a defined Elasticsearch instance.
 
-When launching your benchmark, if enabled, the metadata collection container (backpack)
-will launch as a DaemonSet on all nodes in the cluster. It will then run stockpile
-to gather data and then push that information up to Elasticsearch.
+There are two ways to launch metadata collection from a workload.
+
+# Targeted Init Containers
```
aakarshg: Can you please update docs to reflect that ycsb doesn't support this ? And maybe also add a note to ycsb docs?

playbook.yml
dry923: so we only want to run the role when metadata_targeted is false. the init container is outside of the backpack role. that help some?

docs/metadata.md
dry923: ack. good call

dry923: updated

PR #286, opened 2020-02-19
Adding stalebot

dry923:
Adding configuration file for stalebot - https://github.com/probot/stale

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,22 @@
+# Settings for Pulls:
+pulls:
+  # Number of days of inactivity before an Issue or Pull Request becomes stale
+  daysUntilStale: 15
+
+  # Number of days of inactivity before an Issue or Pull Request with the stale label is closed.
+  daysUntilClose: 10
+
+  # Label to use when marking as stale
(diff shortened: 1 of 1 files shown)


.github/stale.yml
```diff
@@ -0,0 +1,9 @@
+# Number of days of inactivity before an Issue or Pull Request becomes stale
+daysUntilStale: 15
+
+# Number of days of inactivity before an Issue or Pull Request with the stale label is closed.
+# Set to false to disable. If disabled, issues still need to be closed manually, but will remain marked as stale.
+daysUntilClose: 10
```
aakarshg: I'm not entirely sure that it's a good idea to auto close PRs/Issues in 10 days.

dry923: It's 10 days after they get marked as stale. So it would be 25 days total from last touch. Would 14 make you happier?

jtaleric: I think 10 is MORE than enough time personally. Is there a way for admins to override?

dry923: There's an option to ignore certain labels. We could use that as a way to override.

Issue #228, opened 2019-09-20
minor nit: CI complains because 2 benchmark-operator pods in existence


bengland2:
also this error occurs in CI, is most likely caused when previous incarnation of benchmark-operator is still terminating while the other one is running.   This doesn't seem to stop it from working, but should be easy to fix.

```
+++ kubectl get pods -l name=benchmark-operator --namespace my-ripsaw -o name
+++ cut -d/ -f2
++ pod_name='benchmark-operator-7b7b6d9786-5gw7b
benchmark-operator-7b7b6d9786-6n878'
++ '[' -z benchmark-operator-7b7b6d9786-5gw7b benchmark-operator-7b7b6d9786-6n878 ']'
tests/common.sh: line 73: [: benchmark-operator-7b7b6d9786-5gw7b: binary operator expected
++ counter=5
++ '[' 5 -eq 60 ']'
++ '[' benchmark-operator-7b7b6d9786-5gw7b benchmark-operator-7b7b6d9786-6n878 '!=' False ']'
tests/common.sh: line 70: [: too many arguments
++ sleep 5
+++ kubectl get pods -l name=benchmark-operator --namespace my-ripsaw -o name
+++ cut -d/ -f2
++ pod_name='benchmark-operator-7b7b6d9786-5gw7b
benchmark-operator-7b7b6d9786-6n878'
++ '[' -z benchmark-operator-7b7b6d9786-5gw7b benchmark-operator-7b7b6d9786-6n878 ']'
tests/common.sh: line 73: [: benchmark-operator-7b7b6d9786-5gw7b: binary operator expected
++ counter=6
++ '[' 6 -eq 60 ']'
++ '[' benchmark-operator-7b7b6d9786-5gw7b benchmark-operator-7b7b6d9786-6n878 '!=' False ']'
tests/common.sh: line 70: [: too many arguments
```

aakarshg:
@dry923 want to take this on?

PR #287, opened 2020-02-19
handle case where old pods are still being terminated

bengland2:
for issue #228 
filter out pods in Terminating state
-z test should have quoted string as argument
so a list of > 1 pod will not cause syntax error

tested with fs-drift and smallfile, but how will CI react?   
Seemed like it made 2 passes when only 1 should have been necessary.

### DIFF
--- a/tests/common.sh
+++ b/tests/common.sh
@@ -76,8 +76,8 @@ function get_pod () {
   pod_name="False"
   until [ $pod_name != "False" ] ; do
     sleep $sleep_time
-    pod_name=$(kubectl get pods -l $1 --namespace ${3:-my-ripsaw} -o name | cut -d/ -f2)
-    if [ -z $pod_name ]; then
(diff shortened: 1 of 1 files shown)


tests/common.sh
```diff
@@ -76,8 +76,8 @@ function get_pod () {
   pod_name="False"
   until [ $pod_name != "False" ] ; do
     sleep $sleep_time
-    pod_name=$(kubectl get pods -l $1 --namespace ${3:-my-ripsaw} -o name | cut -d/ -f2)
-    if [ -z $pod_name ]; then
+    pod_name="$(kubectl get pods -l $1 --namespace ${3:-my-ripsaw} -o name | grep -vi Terminating | cut -d/ -f2)"
```
rsevilla87: This expression won't work as expected since the pod status is not shown when passing the flag `-o name` to kubectl.

Suggestion: using go-templates might be more robust with something like:
```
kubectl get pods -l $1 --namespace ${3:-my-ripsaw} -o go-template --template='{{range .items }}{{if eq .status.phase "Running"}}{{ .metadata.name }}{{" "}}{{end}}{{end}}'
```

PR #293, opened 2020-03-10
Moving the ServiceIP Capture

### LINKED ISSUE #292: ServiceIP resource capture is missing
ServiceIP is a resource that needs to be captured outside of any of the UPerf blocks.

jtaleric:
Fixes #292

### DIFF
--- a/roles/uperf-bench/tasks/main.yml
+++ b/roles/uperf-bench/tasks/main.yml
@@ -35,8 +35,18 @@
       - name = benchmark-operator
   register: bo
 
+- name: Capture ServiceIP
+  k8s_facts:
+    kind: Service
+    api_version: v1
+    namespace: '{{ operator_namespace }}'
+    label_selectors:
+      - type = uperf-bench-server-{{ trunc_uuid }}
+  register: serviceip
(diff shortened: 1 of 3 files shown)


roles/uperf-bench/tasks/main.yml
```diff
@@ -35,8 +35,18 @@
...
+- name: Capture ServiceIP
+  k8s_facts:
+    kind: Service
+    api_version: v1
+    namespace: '{{ operator_namespace }}'
+    label_selectors:
+      - type = uperf-bench-server-{{ trunc_uuid }}
+  register: serviceip
+  when: uperf.serviceip
+
```
dry923: Wont this fail if it runs before the uperf-bench-server is created like on the first iteration through?

jtaleric: The server doesn't need anything with the serviceip creation... I don't think it should matter (looking through the code)..

dry923: Right. The question is really if k8s_facts returns nothing or an error if it doesnt find anything. I would assume nothing but  would want to be sure.

PR #297, opened 2020-03-24
Updating ci test scripts to run tests related to the change

dry923:
This has ci now only run tests related to the change. However, if some critical files are changed it will run all the tests.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -16,26 +16,20 @@ max_concurrent=3
 eStatus=0
 
 git_diff_files="$(git diff remotes/origin/master --name-only)"
-for file in ${git_diff_files}
-do
-  echo "$file" >> tests/git_diff
-done
 
-check_all_tests=$(check_full_trigger)
-if [[ "$check_all_tests" != "True" ]]
-then
-  echo "checking which tests need to be run"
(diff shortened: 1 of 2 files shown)


test.sh
```diff
@@ -16,19 +16,13 @@ max_concurrent=3
...
-do
-  echo "$file" >> tests/git_diff
-done
 
-check_all_tests=$(check_full_trigger)
-if [[ "$check_all_tests" != "True" ]]
-then
-  echo "checking which tests need to be run"
-  populate_test_list "${git_diff_files}"
+if [[ `echo "${git_diff_files}" | grep -cv /` -gt 0 || `echo ${git_diff_files} | grep -E "(build/|deploy/|group_vars/|resources/|/common.sh|/uuid)"` ]]; then
```
aakarshg: should we also add the [run_test.sh](https://github.com/cloud-bulldozer/ripsaw/blob/master/run_test.sh) to the list?

dry923: The first part of the if checks if there are any changes in the root directory (ie run_test.sh, the playbook, etc, etc) so it should catch that and trigger a full run.

PR #302, opened 2020-04-07
Add block to pre-fill cluster if fiod.prefill is true

acalhounRH:
Added logic to fio-distributed to pre-fill the cluster before running the user provided test. 

This PR un-blocks the perf-sclae team with testing OCS.

@aakarshg @bengland2 @jtaleric @dry923 @dustinblack

### DIFF
--- a/docs/fio_distributed.md
+++ b/docs/fio_distributed.md
@@ -29,6 +29,7 @@ spec:
   workload:
     name: "fio_distributed"
     args:
+      prefill: true
       samples: 3
       servers: 3
       pin_server: ''
@@ -168,7 +169,8 @@ The workload loops are nested as such from the CR options:
 - **rook_ceph_drop_caches**: (optional) If set to `True`, the Rook-Ceph OSD caches will be dropped prior to each sample
 - **rook_ceph_drop_cache_pod_ip**: (optional) The IP address of the pod hosting the Rook-Ceph cache drop URL -- See [cache drop pod instructions](#dropping-rook-ceph-osd-caches) below
 > Technical Note: If you are running kube/openshift on VMs make sure the diskimage or volume is preallocated.
-
+- **prefill**: (Optional) boolean to enable/disable prefill SDS <br />
+--  prefill requirement stems from Ceph RBD thin-provisioning - just creating the RBD volume doesn't mean that there is space allocated to read and write out there. For example, reads to an uninitialized volume don't even talk to the Ceph OSDs, they just return immediately with zeroes in the client.
 #### EXPERT: spec.global_overrides
 The `key=value` combinations provided in the list here will be appended to the `[global]` section of the fio
 jobfile configmap. These options will therefore override the global values for all workloads in the loop.
--- a/roles/fio-distributed/templates/client.yaml
+++ b/roles/fio-distributed/templates/client.yaml
@@ -38,6 +38,14 @@ spec:
         command: ["/bin/sh", "-c"]
         args:
           - "cat /tmp/host/hosts;
+{% if fiod.prefill is defined and fiod.prefill is sameas true %}
+             echo ***************Prefill*****************;
+             cat /tmp/fio/fiojob-prefill;
+             mkdir -p /tmp/fiod-{{ uuid }}/fiojob-prefill;
+             fio --client=/tmp/host/hosts /tmp/fio/fiojob-prefill --output-format=json;
+             sleep 300;
+             echo ***********End of Prefill*************;
+{% endif %}
 {% for numjobs in fiod.numjobs %}
 {% for bs in fiod.bs %}
 {% for job in fiod.jobs %}
--- a/roles/fio-distributed/templates/configmap.yml.j2
+++ b/roles/fio-distributed/templates/configmap.yml.j2
@@ -5,6 +5,26 @@ metadata:
   name: fio-test-{{ trunc_uuid }}
   namespace: '{{ operator_namespace }}'
 data:
+{% if fiod.prefill is defined and fiod.prefill is sameas true %}
+  fiojob-prefill: |
+    [global]
+    directory={{fio_path}}
+    filename_format=f.\$jobnum.\$filenum
+    clocksource=clock_gettime
+    kb_base=1000
+    unit_base=8
+    ioengine=libaio
+    size={{fiod.filesize}}
+    bs=4096KiB
+    iodepth=1
+    direct=1
+    numjobs=1
+    
+    [write]
+    rw=write
+    create_on_open=1
+    fsync_on_close=1
+{% endif %}
 {% for numjobs in fiod.numjobs %}
 {% for bs in fiod.bs %}
 {% for job in fiod.jobs %}
@@ -20,7 +40,7 @@ data:
     log_hist_msec={{fiod.log_sample_rate}}
     clocksource=clock_gettime
     kb_base=1000
-    unit_base='8'
+    unit_base=8
     ioengine=libaio
     size={{fiod.filesize}}
     bs={{bs}}
--- a/tests/test_crs/valid_fiod.yaml
+++ b/tests/test_crs/valid_fiod.yaml
@@ -12,6 +12,7 @@ spec:
   workload:
     name: "fio_distributed"
     args:
+      prefill: true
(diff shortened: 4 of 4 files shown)


roles/fio-distributed/templates/client.yaml
```diff
@@ -38,6 +38,9 @@ spec:
         command: ["/bin/sh", "-c"]
         args:
           - "cat /tmp/host/hosts;
+{% if fiod.prefill is sameas true %}
+             cat /tmp/fio/fiojob-prefill; mkdir -p /tmp/fiod-{{ uuid }}/fiojob-prefill; python /opt/snafu/run_snafu.py -t fio -H /tmp/host/hosts -j /tmp/fio/fiojob-prefill -s 1 -d /tmp/fiod-{{ uuid }}/fiojob-prefill;
```
aakarshg, line 42: if we're not interested in pre-fill results maybe we dont need to invoke snafu ? otherwise it'll skew results as these documents would still have same identifiers

jtaleric: @aakarshg why would you want to invoke fio a different way? Seems awfully counterproductive.

aakarshg: concern is around how do we throw away these prefill documents ? since we're using the container, we'll need to figure out a way to not index these documents, there's 2 ways - either we disable indexing these documents, by unsetting es vars and then resetting them or we can just invoke fio directly. IMHO the right way to do this would be to add another flag that identifies documents generated by prefill job, so that when we do aggregation or comparison, we can filter by that identifier and moving forward if we're even interested in those documents we can still do that.

jtaleric: It was my understanding based on the job below, there would be no results created?

bengland2:
So I don't agree with @jtaleric  on this one.   I think this prefill is a fundamental feature of fio benchmark (and yes I should have paid attention to it sooner).   It has precedent in how we did openstack testing, and baremetal testing (with CBT).     Ceph RBD is a thin-provisioning system so it by definition will  behave differently the first time the volume is used.    The fio benchmark is no use to anyone unless variance in samples can be reduced so we know that the results are accurate.    We will have enough trouble with it in the public clouds as it is (ask Sagy Volkov in TM).  Alex's proposal allows you to avoid doing the prefill if you want to, but I think the default should probably be to prefill.

Joe there's nothing wrong with having an init pod feature, but using that to solve this problem is a customization option that means every single ripsaw fio user will have to rediscover it for themselves.   We *want the benchmark itself* to know how to do it right, automatically.

Also databases have been dealing with the need to build the database before testing it for some time, how did they manage?

roles/fio-distributed/templates/client.yaml
aakarshg: the job below that was newly added is exactly the same as the normal one, so i'm not sure what you mean by there'll be no results created

jtaleric:
> So I don't agree with @jtaleric on this one. I think this prefill is a fundamental feature of fio benchmark (and yes I should have paid attention to it sooner). It has precedent in how we did openstack testing, and baremetal testing (with CBT). Ceph RBD is a thin-provisioning system so it by definition will behave differently the first time the volume is used. The fio benchmark is no use to anyone unless variance in samples can be reduced so we know that the results are accurate. We will have enough trouble with it in the public clouds as it is (ask Sagy Volkov in TM). Alex's proposal allows you to avoid doing the prefill if you want to, but I think the default should probably be to prefill.
> 
> Joe there's nothing wrong with having an init pod feature, but using that to solve this problem is a customization option that means every single ripsaw fio user will have to rediscover it for themselves. We _want the benchmark itself_ to know how to do it right, automatically.
> 
> Also databases have been dealing with the need to build the database before testing it for some time, how did they manage?

Why won't the init container fix this? You could also have the init container _not_ fill similar to this solution. 

The point of my comment is we want to have a *generic* solution, not something specific to just fiod. That is all...

roles/fio-distributed/templates/client.yaml
jtaleric: ack - i thought based on our discussion with @acalhounRH the fill job would be ignored. I wonder if we should not pass the `directory` ?

acalhounRH:
Latest change switched snafu with fio in order to prevent prefill performance results from skewing results. tested with prefill enabled and disabled with no issues observed.

dry923:
Can we update the cr files as well as add in some commentary in the documentation to explain the change (whether it's a temporary fix or the long term one)?

bengland2:
@acalhounRH you could mention in the docs that prefill requirement stems from Ceph RBD thin-provisioning - just creating the RBD volume doesn't mean that there is space allocated to read and write out there.  For example, reads to an uninitialized volume don't even talk to the Ceph OSDs, they just return immediately with zeroes in the client.

acalhounRH:
Updated to include changes to documentation and CI vailid CR file. Current structure does not support invalid test so only included the happy path.

tests/test_crs/valid_fiod.yaml
```diff
@@ -28,6 +30,7 @@ spec:
       read_ramp_time: 1
       filesize: 10MiB
       log_sample_rate: 1000
+      storagesize: 16Mi
```
rsevilla87, line 33: I think we don't need this variable

roles/fio-distributed/templates/client.yaml
```diff
@@ -38,6 +38,14 @@ spec:
         command: ["/bin/sh", "-c"]
         args:
           - "cat /tmp/host/hosts;
+{% if fiod.rook_ceph.prefill is sameas true %}
```
rsevilla87, line 41: Modify conditional for something like this to avoid breaking when the variable is not set
```
{% if prefill is defined and prefill is sameas true %}
```

roles/fio-distributed/templates/configmap.yml.j2
```diff
@@ -5,6 +5,26 @@ metadata:
   name: fio-test-{{ trunc_uuid }}
   namespace: '{{ operator_namespace }}'
 data:
+{% if fiod.rook_ceph.prefill is sameas true %}
```
rsevilla87, line 8: Modify conditional for something like this to avoid breaking when the variable is not set
```
{% if prefill is defined and prefill is sameas true %}
```

acalhounRH: ack. will make change

roles/fio-distributed/templates/client.yaml
acalhounRH: ack. will make change

tests/test_crs/valid_fiod.yaml
```diff
@@ -12,6 +12,8 @@ spec:
   workload:
     name: "fio_distributed"
     args:
+      rook_ceph:
```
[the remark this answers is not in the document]
acalhounRH: ack. will make change

tests/test_crs/valid_fiod.yaml
acalhounRH: the issue here is that we do not want fio to write to all of the PVC, by setting a larger storagesize(PVC size) and keeping filesize to 10MiB, we ensure that its not 100% full and future write operations will not fail.

rsevilla87: Thanks for the explanation @acalhounRH, that makes complete sense

docs/fio_distributed.md
```diff
@@ -168,7 +169,8 @@ The workload loops are nested as such from the CR options:
 - **rook_ceph_drop_caches**: (optional) If set to `True`, the Rook-Ceph OSD caches will be dropped prior to each sample
 - **rook_ceph_drop_cache_pod_ip**: (optional) The IP address of the pod hosting the Rook-Ceph cache drop URL -- See [cache drop pod instructions](#dropping-rook-ceph-osd-caches) below
 > Technical Note: If you are running kube/openshift on VMs make sure the diskimage or volume is preallocated.
-
+- **rook_ceph.prefill**: (Optional) boolean to enable/disable prefill SDS <br />
```
aakarshg, line 172: this is just prefill right ?

PR #303, opened 2020-04-08
Fixing potential infinite loop when logging errors

dry923:
There is a case where we may error out during one of the log messages below which will cause us to re-error. Moving the ERRORED state change above and adding an if statement should resolve this.

### DIFF
--- a/tests/common.sh
+++ b/tests/common.sh
@@ -225,12 +225,18 @@ function wait_for() {
 }
 
 function error {
+  if [[ $ERRORED == "true" ]]
+  then
+    exit
+  fi
+
+  ERRORED=true
+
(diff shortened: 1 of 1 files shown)


tests/common.sh
```diff
@@ -225,12 +225,18 @@ function wait_for() {
 }
 
 function error {
+  if [[ $ERRORED == "true" ]]
+  then
+    exit
```
aakarshg, line 230: so we get logs on first attempt alone if im reading it right

dry923: Correct. one of the potential issues was if it tries to dump the logs and breaks. the trap would send it back to here and it would just go and go. This way we only try and dump the logs once and then bugger out.

PR #316, opened 2020-04-22
support elasticsearch indexing with ssl

### LINKED ISSUE #312: Connecting to a Elasticsearch with TLS self-signed certificates do not work
Hi,

Currently ripsaw is not capable to connect to an Elasticsearch server deployed using TLS self-signed certificates.

2020-04-15T12:29:18Z - WARNING  - MainProcess - run_snafu: Elasticsearch connection caused an exception :Root certificates are missing for certificate validation. Either pass them in using the ca_certs parameter or install certifi to use it automatically.

Regards.

aakarshg:
this commit is a follow up to the snafu commit
cloud-bulldozer/snafu#169, to enable indexing to elasticsearch that
is setup with ssl. also cleaned up elasticsearch related documents,
and put them in a single new document to allow for easier readability.

TODO: update backpack to also support indexing to es with ssl.

Closes #312


roles/hammerdb/templates/db_workload.yml.j2
```diff
@@ -17,9 +17,12 @@ spec:
...
-          - "/usr/local/bin/uid_entrypoint; 
+        args:
+          - "/usr/local/bin/uid_entrypoint;
              {% if elasticsearch is defined %}
+             {% if elasticsearch.cert_verify is defined %}
+             export es_verify_cert={{ elasticsearch.cert_verify }};
+             {% endif %}
              {% if elasticsearch.server is defined %}
              export es={{elasticsearch.server}};
              export es_port={{elasticsearch.port}};
```
rsevilla87, line 28: Why don't we set these environment variables in the containers.env section?. This will make code clearer.
I think the same for the remaining exported variables

roles/uperf-bench/templates/workload.yml.j2
```diff
@@ -56,6 +56,9 @@ spec:
 {% endif %}
 {% endif %}
 {% if elasticsearch is defined %}
+{% if elasticsearch.cert_verify is defined %}
+             export es_verify_cert={{ elasticsearch.cert_verify }};
+{% endif %}
 {% if elasticsearch.server is defined %}
              export es={{elasticsearch.server}};
              export es_port={{elasticsearch.port}};
```
rsevilla87, line 64: Same as above.

roles/hammerdb/templates/db_workload.yml.j2
jtaleric: I think this is a fine suggestion. This is mostly historical, *why* we export vs the pod def.

PR #318, opened 2020-04-23
Adding logic to CI testing to check ES for workload results

dry923:
Added check_es python script that will check an ES instance for a given uuid and index. This is used to help validate that the tests ran and indexed properly. This was only added to workloads that utilize snafu.
Additionally, updated the backpack CI test to run as a privileged container with a service account. This way we can ensure we check as many of the functions of the pod as possible. However, I did have to add in a sleep after the workload was complete as ES seemed to take some time to process the data and make it available to check via the check_es script.

### DIFF
--- a/resources/backpack_role.yaml
+++ b/resources/backpack_role.yaml
@@ -107,6 +107,14 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - coordination.k8s.io
+  resources:
+  - leases
+  verbs:
+  - get
+  - list
+  - watch
 ---
 apiVersion: v1
 kind: ServiceAccount
--- a/tests/check_es.py
+++ b/tests/check_es.py
@@ -1,0 +1,62 @@
+#!/usr/bin/env python
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+import argparse
+import elasticsearch
+import sys
+
+def _check_index(server,port,uuid,index,es_ssl):
+    _es_connection_string = str(server) + ':' + str(port)
+
+    if es_ssl == "true":
+        import urllib3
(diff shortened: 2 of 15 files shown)


dry923:
Note the backpack test will fail until https://github.com/cloud-bulldozer/scribe/pull/26 is merged

resources/backpack_role.yaml
```diff
@@ -107,6 +107,14 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - coordination.k8s.io
```
aakarshg, line 111: curious why is this needed ?

tests/test_backpack.sh
```diff
@@ -21,23 +21,22 @@ function functional_test_backpack {
...
-      echo "Found stockpile.json on "$pod
-    else
-      echo "Could not find stockpile.json on "$pod
-      echo "Backpack test: Fail"
-      exit 1
-    fi
-  done
-
-  echo "Backpack test: Success"
+  indexes="cpu_vulnerabilities-metadata cpuinfo-metadata dmidecode-metadata k8s_configmaps-metadata k8s_namespaces-metadata k8s_nodes-metadata k8s_pods-metadata lspci-metadata meminfo-metadata sysctl-metadata"
```
aakarshg, line 32: this would mean everytime we add a new scribe module, we'll have to update this... im okay with it, but shouldn't backpack be tested outside of ripsaw? we are still testing its integration with ripsaw ( we've to wait for benchmark pods to come up, which in turn waits on backpack to finish )

resources/backpack_role.yaml
dry923: It was not successfully running the node collection due to an api restriction. Adding the coordination api group resolved it.

tests/test_backpack.sh
dry923: We already have a test backpack script/cr in place. This was just adding to the robustness of the test. Additionally, this would ensure that errors related to the backpack role/permissions/etc get caught.

tests/check_es.py
```diff
@@ -0,0 +1,48 @@
...
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+import argparse
+import elasticsearch
+import sys
+
+def _check_index(server,port,uuid,index):
```
jtaleric, line 18: /me waits for someone to mention TLS ;)

dry923:
Updated to include ssl checking for ES. @aakarshg can you confirm that looks correct to you?

aakarshg:
> Updated to include ssl checking for ES. @aakarshg can you confirm that looks correct to you?

this LGTM

PR #320, opened 2020-04-23
Adding a short sleep to each CI test run to avoid command collision

dry923:
Adding a small (<10 second) random sleep to the start of each workload CI test. This is to help avoid potential command collision. 

This also may resolve https://github.com/cloud-bulldozer/ripsaw/issues/319


aakarshg:
merging this to test 318, looks harmless enough and can revert if needed.

PR #321, opened 2020-04-24
Fio yaml enhancements

bengland2:
This has been tested with a 112-OSD Ceph cluster.   It allows for really quick tests to debug workload parameters, etc.  by making  the 300-second prefill sleep configurable (default 0).   But it can handle really long tests efficiently by leveraging Alex's prefill functionality to do the full file write only once with maximum I/O size, and then the actual write test is done on a time-limited basis.   Some fixes were needed to support each test type independently.   We need this for OCS QE.

### DIFF
--- a/roles/fio-distributed/templates/configmap.yml.j2
+++ b/roles/fio-distributed/templates/configmap.yml.j2
@@ -57,7 +57,7 @@ data:
 {% endif %}
 {% if job_params|default([])|length %}
 {% for match in job_params %}
-{% if match.jobname_match in job %}
+{% if match.jobname_match == job %}
 {% for param in match.params %}
     {{ param }}
 {% endfor %}
--- a/resources/crds/ripsaw_v1alpha1_fio_distributed_cr.yaml
+++ b/resources/crds/ripsaw_v1alpha1_fio_distributed_cr.yaml
@@ -12,23 +12,41 @@ spec:
   workload:
     name: "fio_distributed"
     args:
+      # if true, do large sequential write to preallocate volume before using
+      prefill: true
+      # number of times each test
       samples: 3
+      # number of fio pods generating workload
       servers: 3
+      # put all fio pods on this server
       pin_server: ''
+      # test types, see fio documentation
       jobs:
         - write
         - read
+        - randwrite
+        - randread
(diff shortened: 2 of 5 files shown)


roles/fio-distributed/templates/configmap.yml.j2
```diff
@@ -57,7 +57,7 @@ data:
 {% endif %}
 {% if job_params|default([])|length %}
 {% for match in job_params %}
-{% if match.jobname_match in job %}
+{% if match.jobname_match == job %}
```
rsevilla87, line 60: As far as I know, this conditional does not support regular expressions. I've searched about how to implement regex conditionals in jinja but I'm not able to find a way yet.

rsevilla87: @bengland2 I figured out a way to perform regex conditionals in Ansible jinja templates via  the `regex_search` filter. i.e

```jinja2
{% if job | regex_search(match.jobname_match) %}
{% endif %}
```

bengland2: I think we want == exact match because what was happening before is that the "w" in the original yaml would match both write,randwrite,readwrite,... , and that leads to fio exiting if you try to define different parameters for the other job types.    Have to be careful using pattern matching.

rsevilla87: Yeah, I know the intention of using == instead of in. I was suggesting to use the `regex_search` filter since I noticed that you're configuring some jobname_match with regular expressions such as:
     - jobname_match: "^rw$"  of      - jobname_match: "^write$"

bengland2:
Raul pointed out a problem with the PR, the YAML files were still using regex but the code in the template file had been changed to not use regex, I've updated the YAML files to not use regex.   I think this is much simpler to understand for people and more extensible (the user can customize whatever they want with fio job parameters for each test type, but we can set up defaults that are sane).   Will redo CI test with that, it should pass because I'm running it on an OCS cluster now.

roles/fio-distributed/templates/configmap.yml.j2
bengland2: you were right, I updated PR, resolving this.

PR #322, opened 2020-04-24
Adding optional image location for workloads

dry923:
This adds an optional workload argument image which takes the image location. This is to make debugging and testing changes easier by allowing you to pass a different image location without having to rebuild the operator image.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,33 @@ un-block it. The benchmarks mentioned above that state `Used` for Reconciliation
 These two bencharmks are written in a way that doesn't allow for reconciliation to be implemented. To take
 advantage of the reconciliation loop, these two benchmarks need to be rewritten.
 
+## Optional workload images
+Optional locations for workload images can now be added easily without the need to rebuild the operator.
+To do so in the workload args section of the CR add image: [location]
+
+NOTE: This is not a required arguement. If omitted it will default to the currently verified workload image.
+Additionally, this is not enabled for YCSB
+
+For Example:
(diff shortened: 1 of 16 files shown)


jtaleric:
> roles/fio-distributed/templates/.client.yaml.swp Is also added, probably not intentional. Can you also please add small documentation on how one can use it? re: update the cr with the vars?

Seems like a good opportunity to add to our gitignore.

dry923:
> roles/fio-distributed/templates/.client.yaml.swp Is also added, probably not intentional. Can you also please add small documentation on how one can use it? re: update the cr with the vars?

Good call. Removed the swp file and added *.swp to gitignore. Additionally, I added a note in the README on the optional arguement.

README.md
```diff
@@ -33,6 +33,31 @@ un-block it. The benchmarks mentioned above that state `Used` for Reconciliation
 These two bencharmks are written in a way that doesn't allow for reconciliation to be implemented. To take
 advantage of the reconciliation loop, these two benchmarks need to be rewritten.
 
+## Optional workload images
+Optional locations for workload images can now be added easily without the need to rebuild the operator.
+To do so in the workload args section of the CR add image: [location]
+NOTE: This is not a required arguement. If omitted it will default to the currently verified workload image.
```
aakarshg, line 39: This doesn't work for ycsb, maybe we should add that to note as well?

dry923: Noted in the Readme

PR #323, opened 2020-04-28
Adding Cerberus integration

dry923:
This allows the user to pass a cerberus_url variable to ripsaw via the CR file. If provided it will set the benchmark status of Cerberus to connected. If Cerberus reports a failure during the lifecycle of the workload the benchamrk status will be set to Error and it will not proceed further. It will NOT kill any running pods or make any modifications to the cluster. It will just halt any further progress. 

Additionally, once an error state has been reached it will not be reset by ripsaw. The workload will need to either be restarted or you will need to set the state back to its previous value by hand.

Cerberus: https://github.com/openshift-scale/cerberus

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,10 @@ spec:
 ## Metadata Collection
 [Metadata Collection](docs/metadata.md)
 
+## Cerberus Integration
+[Cerberus Integration](docs/cerberus.md)
+
 ## Community
-Key Members(slack_usernames): aakarsh, dblack or rook
+Key Members(slack_usernames): aakarsh, dry923, rsevilla or rook
 * [**#sig-scalability on Kubernetes Slack**](https://kubernetes.slack.com)
 * [**#forum-kni-perfscale on CoreOS Slack**](https://coreos.slack.com)
--- a/docs/cerberus.md
+++ b/docs/cerberus.md
@@ -1,0 +1,60 @@
+# Cerberus Integration
+
+How To:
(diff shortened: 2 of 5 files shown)


chaitanyaenr:
On a side note, we might want to stop the benchmark run early in plow when it's in error state: https://github.com/cloud-bulldozer/plow/blob/master/workloads/network-perf/run_pod_network_test_fromgit.sh#L74. cc: @mohit-sheth.

dry923:
Example output of a benchmark that had Cerberus find an error in the cluster

```
$ kubectl -n my-ripsaw get benchmarks.ripsaw.cloudbulldozer.io
NAME              TYPE    STATE   METADATA STATE   CERBERUS    UUID                                   AGE
byowl-benchmark   byowl   Error   Complete         Connected   11faec65-4009-58e8-ac36-0233f0fc822d   10m
```

README.md
```diff
@@ -69,7 +69,10 @@ spec:
 ## Metadata Collection
 [Metadata Collection](docs/metadata.md)
 
+## Cerberus Integration
+[Cerberus Integration](docs/cerberus.md)
+
 ## Community
-Key Members(slack_usernames): aakarsh, dblack or rook
+Key Members(slack_usernames): aakarsh, dry923, rsevilla or rook
```
aakarshg, line 76: ouch this hurts... @dustinblack has to approve this ;)

jtaleric: lol

PR #328, opened 2020-04-30
Allow privileged for fio server pods

rsevilla87:
Allow privileged fio server pods so we can drop caches using the fio parameter exec_prerun as mentioned at the updated documentation.

### DIFF
--- a/docs/fio_distributed.md
+++ b/docs/fio_distributed.md
@@ -26,6 +26,7 @@ spec:
     port: 9200
   clustername: myk8scluster
   test_user: my_test_user_name
+  privileged: false
   workload:
     name: "fio_distributed"
     args:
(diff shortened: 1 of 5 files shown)


jtaleric:
While this might work for Ben if -

a) The user runs FIO across all the nodes w/ OSDs
b) Are not working with distributed storage like Ceph. 

Overall, I think it is a good change..

rsevilla87:
> While this might work for Ben if -
> 
> a) The user runs FIO across all the nodes w/ OSDs
> b) Are not working with distributed storage like Ceph.
> 
> Overall, I think it is a good change..

Yeah, I'm aware of that. It does not resolve Ben's issue completely because Fio severs usually don't run on OSD nodes.

jtaleric:
> > While this might work for Ben if -
> > a) The user runs FIO across all the nodes w/ OSDs
> > b) Are not working with distributed storage like Ceph.
> > Overall, I think it is a good change..
> 
> Yeah, I'm aware of that. It does not resolve Ben's issue completely because Fio severs usually don't run on OSD nodes.

Right, you might be aware of that, however, people might *think* this does more than appears. :)

bengland2:
I'm not pursuing this because I think it's better to have a separate cache dropping pod (or daemonset for kernel cache) that is privileged and does only that, and then invoke it via a URL.

PR #331, opened 2020-05-04
allow fio servers to avoid nodes with a label

bengland2:
Do not merge yet, this is more of an idea at this point, but I just tried fio in a cluster with 31 workers an d 16 OCS nodes, ripsaw fio deployed 30 pods across 30 nodes including the OCS nodes, which is a real problem for performance.    We want it to use the other workers.   I structured this change so that it is not OCS specific - it will place fio-server pods on any worker **other than** the nodes with the specified label.

### DIFF
--- a/roles/fio-distributed/templates/servers.yaml
+++ b/roles/fio-distributed/templates/servers.yaml
@@ -12,6 +12,14 @@ spec:
         app: fio-benchmark-{{ trunc_uuid }}
     spec:
       affinity:
+{% if fiod.avoidlabel is defined %}
+        nodeAffinity:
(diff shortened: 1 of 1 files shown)


aakarshg:
Can you please add this to documentation as well @bengland2

bengland2:
@rsevilla87 @aakarshg So I found a similar feature already in existence, is this change necessary?   

oc adm cordon  --selector=cluster.ocs.openshift.io/openshift-storage

It works.    So I'm going to close this but the code will still be in here in case we change our mind.

rsevilla87:
Good to know it worked @bengland2

PR #332, opened 2020-05-05
Variable standarization

rsevilla87:
This PR aims to standardize the way to pass certain variables to the workloads.


bengland2:
totally agree with this one, smallfile still requires es-index, there is no reason for this because it should just default to ripsaw-smallfile, I thought it was just smallfile that had the problem.    Make life simpler for people that use ripsaw and snafu.

roles/fio-distributed/templates/client.yaml
```diff
@@ -51,7 +47,7 @@ spec:
 {% for numjobs in fiod.numjobs %}
 {% for bs in fiod.bs %}
 {% for job in fiod.jobs %}
-             cat /tmp/fio/fiojob-{{job}}-{{bs}}-{{numjobs}}; mkdir -p /tmp/fiod-{{ uuid }}/fiojob-{{job}}-{{bs}}-{{numjobs}}; python /opt/snafu/run_snafu.py -t fio -H /tmp/host/hosts -j /tmp/fio/fiojob-{{job}}-{{bs}}-{{numjobs}} -s {{fiod.samples}} -d /tmp/fiod-{{ uuid }}/fiojob-{{job}}-{{bs}}-{{numjobs}};
+             cat /tmp/fio/fiojob-{{job}}-{{bs}}-{{numjobs}}; mkdir -p /tmp/fiod-${uuid}/fiojob-{{job}}-{{bs}}-{{numjobs}}; python /opt/snafu/run_snafu.py -t fio -H /tmp/host/hosts -j /tmp/fio/fiojob-{{job}}-{{bs}}-{{numjobs}} -s {{fiod.samples}} -d /tmp/fiod-${uuid}/fiojob-{{job}}-{{bs}}-{{numjobs}};
```
dry923, line 50: Why the change from {{ uuid }} to ${uuid}? It only seems to be done in fio and uperf

rsevilla87:
@bengland2 I have had to increase the fio test timeout since it was timing out in the CI.

resources/crds/ripsaw_v1alpha1_ripsaw_crd.yaml
```diff
@@ -29,6 +29,8 @@ spec:
                     type: string
                   port: 
                     type: integer
+                  es_index:
```
dry923, line 32: Should we rename this to just index as es is redundant since its in the elasticsearch object?

rsevilla87: I named it `es_index` since index is the name of a function in Jinja and will crash.

roles/load-ycsb/tasks/main.yml
```diff
@@ -35,9 +35,9 @@
               - name: es_port
                 value: "{{elasticsearch.port | default()}}"
               - name: es_index
-                value: "{{ es_index }}"
+                value: "{{ elasticsearch.es_index | default('ripsaw-ycsb') }}"
               - name: user
-                value: "{{test_user | default()}}"
+                value: "{{test_user | default('ripsaw') }}"
```
dry923, line 40: "{{test_user | default('ripsaw')}}" or "{{ test_user | default('ripsaw') }}" for consistency

roles/load-ycsb/tasks/main.yml
```diff
@@ -48,7 +48,7 @@
                 value: "{{uuid}}"
               imagePullPolicy: Always
               command: ["/bin/sh", "-c"]
-              args: ["python /opt/snafu/run_snafu.py --tool ycsb -r 1 -l -d {{ ycsb.driver }} -u {{ uuid }} --user {{ test_user | default(\"ripsaw\") }} -w {{ ycsb_workload }} -x \"{{ ycsb.options_load }}\""]
+              args: ["python /opt/snafu/run_snafu.py --tool ycsb -r 1 -l -d {{ ycsb.driver }} -u {{ uuid }} --user {{ test_user|default('ripsaw') }} -w {{ ycsb_workload }} -x \"{{ ycsb.options_load }}\""]
```
dry923, line 51: Why are we re-defaulting test_user when we do it in line 40 above?

roles/smallfile-bench/templates/workload_job.yml.j2
```diff
@@ -40,24 +40,21 @@ spec:
...
-            value: "{{ es_index }}"
-{% endif %}
+            value: "{{ elasticsearch.es_index | default("ripsaw-smallfile") }}"
           - name: redis_host
             value: "{{ bo.resources[0].status.podIP }}"
           - name: clients
             value: "{{ smallfile.clients }}"
           args:
             - python /tmp/smallfile/subscriber {{bo.resources[0].status.podIP}};
-              export TMPDIR=/tmp
```
dry923, line 60: Are we removing the TMPDIR definition outright? In fio it was re-defined elsewhere but here it just seems to be a remove line

roles/uperf-bench/templates/workload.yml.j2
```diff
@@ -85,7 +84,7 @@ spec:
 {% for nthr in uperf.nthrs %}
                  cat /tmp/uperf-test/uperf-{{test}}-{{proto}}-{{size}}-{{nthr}};
                  for i in `seq 1 {{uperf.samples}}`; do
-                   python /opt/snafu/run_snafu.py --tool uperf -w /tmp/uperf-test/uperf-{{test}}-{{proto}}-{{size}}-{{nthr}} -r $i --resourcetype {{resource_kind}} -u {{ uuid }} --user {{test_user | default("ripsaw")}};
+                   python /opt/snafu/run_snafu.py --tool uperf -w /tmp/uperf-test/uperf-{{test}}-{{proto}}-{{size}}-{{nthr}} -r $i --resourcetype {{resource_kind}} -u {{uuid}} --user {{test_user|default("ripsaw")}};
```
dry923, line 87: Same comment re test_user being re-defaulted again

tests/test_fiod.sh
```diff
@@ -27,7 +27,7 @@ function functional_test_fio {
   wait_for "kubectl -n my-ripsaw wait --for=condition=Initialized -l app=fio-benchmark-$uuid pods --timeout=300s" "300s"
   fio_pod=$(get_pod "app=fiod-client-$uuid" 300)
   wait_for "kubectl wait --for=condition=Initialized pods/$fio_pod -n my-ripsaw --timeout=500s" "500s" $fio_pod
-  wait_for "kubectl wait --for=condition=complete -l app=fiod-client-$uuid jobs -n my-ripsaw --timeout=500s" "500s" $fio_pod
+  wait_for "kubectl wait --for=condition=complete -l app=fiod-client-$uuid jobs -n my-ripsaw --timeout=900s" "900s" $fio_pod
```
dry923, line 30: Any idea why this PR would cause a delay here? The 500s timeout seemed to work prior. My concern going to 900s is that it would be done potentially twice in this test and then repeated up to 3 times. Any way to scale this down some?

roles/smallfile-bench/templates/workload_job.yml.j2
bengland2: @dry923 , good question.  He probably removed it because TMPDIR isn't referenced anywhere in this file, but subprocesses can reference it too.      in this line in smallfile code, TMPDIR is referenced, but if it's not defined smallfile just uses /var/tmp/, which does always exist, so not a fatal error.   And it doesn't use it for files it accesses during the test, just logging and thread-ready synch. files, so I don't think it matters.

rsevilla87: Exactly, that's the reason :). I checked smallfile's source and verified it. TMPDIR is back again as env var. I don't want to break any consistency of the workloads, just reorganize and clean-up.

resources/crds/ripsaw_v1alpha1_ripsaw_crd.yaml
rsevilla87: I changed it to index_name

rsevilla87: I eventually changed it for `index_name`

dry923:
@rsevilla87 can you revert the change to the fio timeouts now that https://github.com/cloud-bulldozer/ripsaw/pull/334 has been merged?

tests/test_crs/valid_fiod.yaml
```diff
@@ -7,8 +7,9 @@ spec:
   elasticsearch:
     server: "marquez.perf.lab.eng.rdu2.redhat.com"
     port: 9200
+    index_name: ripsaw-fio
   metadata:
-    collection: true
```
dry923, line 11: collection should be set to true still

tests/test_crs/valid_smallfile.yaml
```diff
@@ -11,7 +11,7 @@ spec:
   elasticsearch:
     server: "marquez.perf.lab.eng.rdu2.redhat.com"
     port: 9200
-  es_index: ripsaw-smallfile
+  index_name: ripsaw-smallfile
```
dry923, line 14: needs indentation under elasticsearch

rsevilla87: my fault, I forgot to revert this change after running tests locally :)

tests/test_crs/valid_fiod.yaml
rsevilla87: ouch!, I forgot to revert this change after running test locally :)

dry923:
@bengland2 @jtaleric Any concerns with this or are we good to merge?

roles/fs-drift/templates/workload_job.yml.j2
```diff
@@ -42,29 +42,28 @@ spec:
...
           - name: clustername
             value: "{{ clustername }}"
-{% if elasticsearch.server is defined %}
           - name: es
-            value: "{{ elasticsearch.server }}"
+            value: "{{ elasticsearch.server | default() }}"
           - name: es_port
-            value: "{{ elasticsearch.port }}"
+            value: "{{ elasticsearch.port | default() }}"
           - name: es_index
```
bengland2, line 52: shouldn't this default as well?   You know you're in ripsaw at this point so why not default to ripsaw-fs-drift ?

roles/fs-drift/templates/workload_job.yml.j2
```diff
@@ -42,29 +42,28 @@ spec:
...
             value: "{{ uuid }}"
           - name: test_user
-            value: "{{ test_user }}"
+            value: "{{ test_user | default("ripsaw") }}"
           - name: clustername
             value: "{{ clustername }}"
-{% if elasticsearch.server is defined %}
           - name: es
-            value: "{{ elasticsearch.server }}"
+            value: "{{ elasticsearch.server | default() }}"
```
bengland2, line 49: this doesn't make sense, if elasticsearch.server is defined, then why would you need to default it?   I think you meant "if elasticsearch is defined?

roles/smallfile-bench/templates/workload_job.yml.j2
```diff
@@ -40,24 +40,23 @@ spec:
...
           - name: clustername
             value: "{{ clustername }}"
-{% if elasticsearch.server is defined %}
           - name: es
-            value: "{{ elasticsearch.server }}"
+            value: "{{ elasticsearch.server | default() }}"
           - name: es_port
-            value: "{{ elasticsearch.port }}"
+            value: "{{ elasticsearch.port | default() }}"
           - name: es_index
```
bengland2, line 50: why do you need to have es_index if value is removed?   Why not just let es_index default to ripsaw-smallfile?  We know we are in ripsaw at this point.   This is one of most annoying parts of smallfile CR at this point, would like to get rid of it.

roles/smallfile-bench/templates/workload_job.yml.j2
```diff
@@ -40,24 +40,23 @@ spec:
...
             value: "{{ uuid }}"
           - name: test_user
-            value: "{{ test_user }}"
+            value: "{{ test_user | default("ripsaw") }}"
           - name: clustername
             value: "{{ clustername }}"
-{% if elasticsearch.server is defined %}
           - name: es
-            value: "{{ elasticsearch.server }}"
+            value: "{{ elasticsearch.server | default() }}"
```
bengland2, line 47: same comment as with fs-drift - did you mean "if elasticsearch is defined"?   because it doesn't make sense to have a default for elasticsearch.server if the onlly way you can get to it is if the user defines elasticsearch.server in their CR.

roles/fio-distributed/templates/client.yaml
```diff
@@ -20,21 +20,17 @@ spec:
...
           - name: es
-            value: "{{ es_server }}"
+            value: "{{ elasticsearch.server | default() }}"
           - name: es_port
-            value: "{{ es_port }}"
+            value: "{{ elasticsearch.port | default() }}"
           - name: es_index
-            value: "{{ es_index }}"
-{% endif %}
+            value: "{{ elasticsearch.index_name | default("ripsaw-fio") }}"
```
bengland2, line 33: can we do this for smallfile and fs-drift also?   Thanks!

roles/fs-drift/templates/workload_job.yml.j2
rsevilla87: I don't understand. It's already set to "ripsaw-fs-drift"

roles/smallfile-bench/templates/workload_job.yml.j2
rsevilla87: I don't understand. It's already set to "ripsaw-smallfile"

roles/fs-drift/templates/workload_job.yml.j2
bengland2: this default could be computed from "ripsaw-{{ spec.workload.name }}" once, by ripsaw, not each benchmark.

rsevilla87: We can't do that now since that the workload name doesn't mach with the index name, we would have rename the workloads to do so. I think that should be done in a different PR.

bengland2: didn't know that, ok then I'll withdraw that suggestion.

roles/smallfile-bench/templates/workload_job.yml.j2
bengland2: withdrawing this comment, it is really another case of the same comment up above that you answered.

roles/fio-distributed/templates/client.yaml
bengland2: you answered this question before, am withdrawing this comment.

roles/fs-drift/templates/workload_job.yml.j2
bengland2: @rsevilla87 what about this one?   This is my last comment - the one below is just another case of it.

rsevilla87: I've made several updates around this, take a look when you get time.

rsevilla87:
Let's wait for https://github.com/cloud-bulldozer/ripsaw/pull/330 to be merged before this one.

dry923:
@rsevilla87 can you give this guy a rebase please

rsevilla87:
@dry923 done!

PR #334, opened 2020-05-08
shorten both of these tests up so CI doesn't timeout

bengland2:
cut down on number of I/O sizes and numjobs values for valid_fiod
cut down on samples and number of I/O sizes for valid_fiod_hostpath
tested with minikube - I was able to reproduce the timeout before this fix.  With this fix I tested it twice and no timeout occurred.

### DIFF
--- a/tests/test_crs/valid_fiod_hostpath.yaml
+++ b/tests/test_crs/valid_fiod_hostpath.yaml
@@ -14,7 +14,7 @@ spec:
   workload:
     name: "fio_distributed"
     args:
-      samples: 2
+      samples: 1
       servers: 2
       prefill: true
       jobs:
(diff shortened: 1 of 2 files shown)


tests/test_crs/valid_fiod_hostpath.yaml
```diff
@@ -25,7 +25,6 @@ spec:
         - randrw
       bs:
         - 4KiB
-        - 64KiB
       numjobs:
         - 1
```
dry923, line 29: Do we still want numjobs 1 and 2 here or just 2 like in the other fio cr?

bengland2: @dry923  I deliberatelly left numjobs 1 and 2 here for better test coverage because we're not covering it in the other yaml.

dry923: ack. Sounds good to me.

PR #335, opened 2020-05-08
Adding optional label parameters to metadata daemonset. 

dry923:
This will allow the daemonset to be targeted on nodes that contain a specific label name=value.

You can now add a list of labels in name/value pairs like below. This is an implicit OR so the node will have to match any of the labels provided.

```
  metadata:
    collection: true
    targeted: false
    label:
      - [ 'foo', 'bar' ]
      - [ 'awesome', 'sauce' ]
```

### DIFF
--- a/roles/backpack/templates/backpack.yml
+++ b/roles/backpack/templates/backpack.yml
@@ -20,6 +20,19 @@ spec:
       tolerations:
       - key: node-role.kubernetes.io/master
         effect: NoSchedule
+{% if metadata.label is defined %}
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+{% for meta_name, meta_value in metadata.label %}
+            - matchExpressions:
+              - key: {{ meta_name }}
+                operator: In
+                values:
+                - {{ meta_value }}
+{% endfor %}
+{% endif %}
       containers:
       - name: backpack
         image: {{ metadata_image | default('quay.io/cloud-bulldozer/backpack:latest') }}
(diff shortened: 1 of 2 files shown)


bengland2:
I still have concerns about including enough nodes in the metadata collection by using OR, but if other people think it should be an AND and not an OR, I won't stand in the way, I probably only need 1 label to do what I need anyway, and I can create a new label to specify exactly the nodes I want to collect on.

dry923:
@bengland2 when nodeSelector takes a list it makes it an AND. I'm open to suggestions on how to create an "or" functionality but I'm not seeing it straight away

roles/backpack/templates/backpack.yml
```diff
@@ -20,6 +20,12 @@ spec:
       tolerations:
       - key: node-role.kubernetes.io/master
         effect: NoSchedule
+{% if metadata.label is defined %}
+      nodeSelector:
```
bengland2, line 24: you are right, NodeSelector list implies an AND, but there is this other way maybe?   [From K8S docs](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/):

"If you specify multiple nodeSelectorTerms associated with nodeAffinity types, then the pod can be scheduled onto a node if one of the nodeSelectorTerms can be satisfied."    That sounds like an OR.   But I don't know how to do it.  If it becomes a big issue I'll do the work to find out how, but for now I think I can make use of this the way it is.

dry923: Ack. Let me take a look and see if its an  easy implementation

dry923:
@bengland2 updated to be an OR functionality. Good find on the doc :-)

bengland2:
@dry923 , thanks, I tested it and it worked for me with 1 label, will try it with 2 soon.

PR #336, opened 2020-05-12
Refresh ES index before checking for documents

rsevilla87:
There are situations where check_es.py script is not able to find documents on ES. This PR updates that script to refresh the index prior to query it.
In addition, it also standardizes some of the CI code and simplifies the wait_clean function.

### DIFF
--- a/tests/test_backpack.sh
+++ b/tests/test_backpack.sh
@@ -18,7 +18,7 @@ trap error ERR
 trap finish EXIT
 
 function functional_test_backpack {
-  figlet $(basename $0)
+  wait_clean
   apply_operator
   kubectl apply -f tests/test_crs/valid_backpack.yaml
   long_uuid=$(get_uuid 20)
@@ -34,9 +34,10 @@ function functional_test_backpack {
   then
     echo "Backpack test: Success"
   else
+    echo "Failed to find data in ES"
     exit 1
-    echo "Faled to find data in ES"
   fi
 }
 
+figlet $(basename $0)
 functional_test_backpack
--- a/tests/check_es.py
+++ b/tests/check_es.py
@@ -28,6 +28,7 @@ def _check_index(server,port,uuid,index,es_ssl):
                                                  ssl_context=ssl_ctx, use_ssl=True)
     else:
         es = elasticsearch.Elasticsearch([_es_connection_string], send_get_body_as='POST')
+    es.indices.refresh(index=index)
     results = es.search(index=index, body={'query': {'term': {'uuid.keyword': uuid}}}, size=1)
     if results['hits']['total']['value'] > 0:
         return 0
--- a/tests/common.sh
+++ b/tests/common.sh
@@ -46,18 +46,7 @@ function populate_test_list {
 }
 
 function wait_clean {
-  kubectl delete benchmark --all -n my-ripsaw
-  kubectl delete all --all -n my-ripsaw
-  for i in {1..30}; do
-    if [ `kubectl get pods --namespace my-ripsaw | grep bench | wc -l` -ge 1 ]; then
-      sleep 5
-    else
-      break
-    fi
-  done
-  if [[ `kubectl get namespace my-ripsaw` ]]; then
-    kubectl delete namespace my-ripsaw
-  fi
+  kubectl delete namespace my-ripsaw --ignore-not-found
 }
 
 # The argument is 'timeout in seconds'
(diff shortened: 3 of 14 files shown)


dry923:
@rsevilla87 I imagine this will break in the get_uuid function since it does not know how to handle multiple benchmark tests at once.
I'm fine with re-using the operator but we should still clean up the old benchmark pods/etc prior to running in my opinion.

dry923:
> @rsevilla87 I imagine this will break in the get_uuid function since it does not know how to handle multiple benchmark tests at once.
> I'm fine with re-using the operator but we should still clean up the old benchmark pods/etc prior to running in my opinion.

Disregard this. I completely missed the delete benchmark line in place of wait_clean. Apparently, reviewing PRs should be done after I have some coffee. :-)

aakarshg:
@dry923 do you have any concerns of this potentially breaking snafu-ci ? I took a look at the pr and it looks unlikely.

dry923:
> @dry923 do you have any concerns of this potentially breaking snafu-ci ? I took a look at the pr and it looks unlikely.

@aakarshg I took another look at the snafu-ci code and I don't think it should effect it.

rsevilla87:
> > @dry923 do you have any concerns of this potentially breaking snafu-ci ? I took a look at the pr and it looks unlikely.
> 
> @aakarshg I took another look at the snafu-ci code and I don't think it should effect it.

Opening a fake PR in Snafu with Depens-on to verify it works :)

dry923:
@rsevilla87 Can you add in a wait for the benchmark cleanup.? I was watching the snafu pr run and I noticed a uperf server being spun up while the other was still in terminating.

aakarshg:
cloud-bulldozer/snafu#182 (comment) shows that the snafu ci is also working fine.

rsevilla87:
> @rsevilla87 Can you add in a wait for the benchmark cleanup.? I was watching the snafu pr run and I noticed a uperf server being spun up while the other was still in terminating.

I think that shouldn't be problematic, CI uses uuid to get the status of the pods, that uuid is collected from benchmark metadata and the old one has been already deleted before creating the new one.

dry923:
> > @rsevilla87 Can you add in a wait for the benchmark cleanup.? I was watching the snafu pr run and I noticed a uperf server being spun up while the other was still in terminating.
> 
> I think that shouldn't be problematic, CI uses uuid to get the status of the pods, that uuid is collected from benchmark metadata and the old one has been already deleted before creating the new one.

It actually caused the uperf server pod to fail to start a few times (I believe it was after the service ip test). Once the old pod was terminated the new one kicked on fine. This is the error that was seen:

```
[root@perf-sm5039-4-2 ~]# oc logs -n my-ripsaw uperf-server-0-e027717c
3.** TCP: Cannot bind to port 8270 Address already in use
Error starting slave
```

dry923:
@rsevilla87 It looks good. My only thought is we should probably standardize on the figlet and apply_operator call being outside the functional_test_[workload] function and instead be like you have it in test_fiod.sh and test_fs_drift.sh. Thoughts?

dry923:
> @rsevilla87 It looks good. My only thought is we should probably standardize on the figlet and apply_operator call being outside the functional_test_[workload] function and instead be like you have it in test_fiod.sh and test_fs_drift.sh. Thoughts?

@rsevilla87 looks like you missed ycsb when applying the above

dry923:
@rsevilla87 i did some testing in https://github.com/cloud-bulldozer/ripsaw/pull/357 and it looks like adding 
```
es.indices.refresh(index=index)
```
to the check_es.py script seems to work the most reliably (ignore the backpack failures that is something else I need to investigate). Thoughts?

rsevilla87:
For some reason, uperf benchmark gets stuck and don't create clients when running the third uperf benchmark, looks like a bug in the operator itself. So I'm changing the goal for this PR since in my opinion it's not worth to waste more time in debugging these things.

dry923:
I like the change of focus for this one. It's good for me. @aakarshg you good w/ the change?

tests/test_backpack.sh
```diff
@@ -18,7 +18,7 @@ trap error ERR
 trap finish EXIT
 
 function functional_test_backpack {
-  figlet $(basename $0)
```
aakarshg, line 21: any reason we're removing the figlet in a few tests?

aakarshg:
> I like the change of focus for this one. It's good for me. @aakarshg you good w/ the change?

yes good with the change, @rsevilla87 can this be merged now?

rsevilla87:
> > I like the change of focus for this one. It's good for me. @aakarshg you good w/ the change?
> 
> yes good with the change, @rsevilla87 can this be merged now?

I think so. This change is not very aggressive and shouldn't break anything.

PR #337, opened 2020-05-14
migrate  to ansible-operator:v0.17.0

rsevilla87:
This new version is based in ubi8 and uses python-3.6 and ansible-2.9.6
k8s_status module is no longer available and has been renamed to  operator_sdk.util.k8s_status since now is provided by an Ansible collection.

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -1,10 +1,18 @@
-FROM quay.io/operator-framework/ansible-operator:v0.8.1
+FROM quay.io/operator-framework/ansible-operator:v0.17.0
+
+COPY requirements.yml ${HOME}/requirements.yml
+COPY image_resources/centos8-appstream.repo /etc/yum.repos.d/centos8-appstream.repo
+RUN ansible-galaxy collection install -r ${HOME}/requirements.yml \
+ && chmod -R ug+rwx ${HOME}/.ansible
+
+USER root
+RUN dnf install -y --nodocs redis --enablerepo=centos8-appstream && dnf clean all
+
+# Temporally workaround to fix https://github.com/operator-framework/operator-sdk/issues/2648
+RUN ln -s ${HOME}/.ansible /.ansible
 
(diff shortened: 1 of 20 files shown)


roles/fio-distributed/tasks/main.yml
```diff
@@ -119,7 +119,7 @@
   - name: Create IP list
     set_fact:
       fio_hosts: |
-         {% for ip, node in pod_details.iteritems() %}
+         {% for ip, node in pod_details.items() %}
```
aakarshg, line 122: nice catch, as iirc there's no iteritems() in python3, as items() itself does the same job

aakarshg: just to keep in mind though is that, if ansible's using python2 then this might become expensive at a large scale ( which is unlikely to happen ) haven't seen someone run fio with >100s of pods yet

aakarshg:
> pgbench and backpack seem to be flaky, but otherwise this PR looks good to me. nice work @rsevilla87.

the backpack test still seems a bit flaky, do you think we should increase the timeout for the test and see if we still hit it ? I'm concerned something might be affecting it, as the same behavior is not seen on a newer pr https://github.com/cloud-bulldozer/ripsaw/pull/349

aakarshg:
> Results for crc_jjb
> 
> Test	Result	Retries	Duration (HH:MM:SS)
> test_backpack	Pass	0	00:04:22
> test_byowl	Pass	0	00:02:00
> test_fiod	Pass	0	00:10:53
> test_fs_drift	Pass	1	00:10:04
> test_hammerdb	Pass	0	00:04:57
> test_iperf3	Pass	0	00:03:35
> test_pgbench	Pass	0	00:02:46
> test_smallfile	Pass	1	00:12:45
> test_sysbench	Pass	0	00:01:56
> test_uperf	Pass	0	00:18:43
> test_ycsb	Pass	0	00:02:22

dont see any more backpack tests failing so this pr is good to go

PR #339, opened 2020-05-14
Update playbook to chose role to include by workload.name

dry923:
This PR hopes to simplify the playbook.yml greatly by condensing down all the includes into a single include: workload.name. 

Highlights:
- Updated playbook.yml include the variable workload.name. 
- Updated all workloads to use the workload_args variable instead of a workload specific one.
- Condensed ycsb down to a single role and enabled the reconciliation loop for it.
- Minor test cr changes


roles/ycsb/tasks/main.yml
```diff
@@ -0,0 +1,208 @@
...
+    kind: Benchmark
+    name: "{{ meta.name }}"
+    namespace: "{{ operator_namespace }}"
+  register: resource_state
+
+- block:
+
+  - name: setting ycsb_workload var
+    set_fact:
+      ycsb_workload: "workloada"
```
aakarshg, line 32: I suggest that we move away from using the variable ycsb_workload in the load phase, as we're overwriting this variable again in when state is "Running Load" and setting it to the same value workloada, and finally in run phase we're iterating over various workloads.

dry923: That's fair. I updated the load side variable to be ycsb_workload_load

roles/ycsb/tasks/main.yml
```diff
@@ -0,0 +1,208 @@
...
+ 
+  - name: Update resource state
+    k8s_status:
+      api_version: ripsaw.cloudbulldozer.io/v1alpha1
+      kind: Benchmark
+      name: "{{ meta.name }}"
+      namespace: "{{ operator_namespace }}"
+      status:
+        state: "Complete"
+    when: workload_args.workloads is undefined
```
dry923, line 87: @jtaleric @aakarshg Not sure if this makes sense or it should just error here. In the case that no workload was given (i.e. only a load) then it just stops saying complete. If preferable I can remove this and it can just error due to the lact of workload list.

aakarshg: do you think we can call this state as "Load Complete" ? if its undefined, that'd make super sense.

aakarshg:
@dry923 can you please rebase and fix the merge conflicts.

dry923:
> @dry923 can you please rebase and fix the merge conflicts.

fixed

roles/ycsb/tasks/main.yml
dry923: @aakarshg My concern with something other than "Complete" is that it isn't consistent with other workloads when they are "done." Additionally, since it won't be flagged as complete it will continue to fall into the workload loop every time and not just skip over it. Thoughts?

roles/fs-drift/tasks/main.yml
```diff
@@ -30,11 +30,11 @@
 
   - set_fact:
       fs_drift_path: /tmp
-    when: fs_drift.worker_pods > 0 and fs_drift.storageclass is not defined
+    when: workload_args.worker_pods|default('1') > 0 and workload_args.storageclass is not defined
```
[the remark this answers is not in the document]
dry923: @rsevilla87 I tested this without worker_pods being defined and it did not have any issue.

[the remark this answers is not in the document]
rsevilla87: You're right, it doesn't fail in this version of the ansible-operator since it's using python2.7

```
Python 2.7.5 (default, Oct 30 2018, 23:45:53) 
[GCC 4.8.5 20150623 (Red Hat 4.8.5-36)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> '1' > 0
```

and python3
```
Python 3.8.2 (default, Feb 28 2020, 00:00:00) 
[GCC 10.0.1 20200216 (Red Hat 10.0.1-0.8)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> '1' > 0
  File "<stdin>", line 1, in <module>
TypeError: '>' not supported between instances of 'str' and 'int'
```

So I believe it will fail as soon as we migrate to a new version of the ansible-operator based in python3.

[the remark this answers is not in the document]
dry923: Ahhhhh ok gotcha. I updated to cast them as int's.

dry923:
@aakarshg I updated this to pull in your ycsb pinning changes. Please confirm when you are reviewing.

PR #340, opened 2020-05-15
Change reference to container images to SHA based reference for disco…

jeanchlopez:
…nnected environment compatibility

Let me know if this is the correct way to proceed

### DIFF
--- a/docs/metadata.md
+++ b/docs/metadata.md
@@ -25,7 +25,7 @@ the workload template with an init container section that looks like:
 {% if metadata.collection|default(false) is sameas true and metadata.targeted|default(true) is sameas true %}
       initContainers:
       - name: backpack-{{ trunc_uuid }}
-        image: quay.io/cloud-bulldozer/backpack:latest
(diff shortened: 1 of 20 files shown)


dry923:
@jeanchlopez you can currently reference an image by sha using the workload image option that was added recently. Take a read of this https://github.com/cloud-bulldozer/ripsaw/blob/master/README.md#optional-workload-images

aakarshg:
> @jeanchlopez you can currently reference an image by sha using the workload image option that was added recently. Take a read of this https://github.com/cloud-bulldozer/ripsaw/blob/master/README.md#optional-workload-images

+1 to what @dry923 has suggested here. pinning images to sha doesn't scale well, as everytime a new patch makes it through we'll have to update sha. As @dry923 suggested you can reference image by sha using workload image option, for the operator image.. I've a PR in progress with operatorhub to use shas - https://github.com/operator-framework/community-operators/pull/1690.

jeanchlopez:
Will take a look and see it that can work.

dry923:
As this PR would require a lot of regular maintenance to update the sha every time an image is updated and there is already a feature that allows the user to achieve this functionality I will be closing this request.

PR #341, opened 2020-05-18
Allow customizing smallfile publisher image

rsevilla87:
Allow customizing smallfile publisher image


dry923:
> This would be useful for benchmark developers because they would not have to rebuild benchmark operator just to get it to use a different image name. Is that the intent here? if so, I'd like to do it for fs-drift and fio as well.

@bengland2 this functionality was added to the other workloads. It was just missed in the publisher section of smallfile. @rsevilla87 ++ for catching it

PR #343, opened 2020-05-19
Updated roles to use snafu as a package

amitsagtani97:


rsevilla87:
@amitsagtani97, can you rebase this one, please?

roles/load-ycsb/tasks/main.yml
```diff
@@ -0,0 +1,73 @@
+---
```
aakarshg, line 1: delete this file?

roles/ycsb-bench/tasks/main.yml
```diff
@@ -0,0 +1,73 @@
+---
```
aakarshg, line 1: this can also be deleted

PR #348, opened 2020-05-26
Cblum cleanup

mulbc:
Not tested yet, prettified the distributed_fio server and client yaml files to make them easier to read

@dry923 - this touches mostly your code


dry923:
@mulbc can you please rebase and resolve the conflicts? Thanks!

roles/fio_distributed/templates/servers.yaml
```diff
@@ -59,15 +59,23 @@ spec:
...
+{% if metadata is defined and (metadata.collection|default(false)) and (metadata.targeted|default(true)) %}
       initContainers:
       - name: backpack-{{ trunc_uuid }}
         image: {{ metadata_image | default('quay.io/cloud-bulldozer/backpack:latest') }}
         command: ["/bin/sh", "-c"]
-{% if metadata is defined and metadata.force|default(false) is sameas true %}
-        args: ["python3 stockpile-wrapper.py -s {{ elasticsearch.server }} -p {{ elasticsearch.port }} -u {{ uuid }} -n $my_node_name -N $my_pod_name --redisip {{ bo.resources[0].status.podIP }} --redisport 6379 --force"]
-{% else %}
-        args: ["python3 stockpile-wrapper.py -s {{ elasticsearch.server }} -p {{ elasticsearch.port }} -u {{ uuid }} -n $my_node_name -N $my_pod_name --redisip {{ bo.resources[0].status.podIP }} --redisport 6379"]
+        args:
```
jtaleric, line 67: :) should all the stockpile runs (across benchmarks) be updated to be "pretty"

mulbc:
I don't see why the test failed, but with the recent Dockerfile update, we have switched to running in python3 now... which have broken a few things in the main.yml.
I have fixed these in my new PR regarding fio in VMs

dry923:
> I don't see why the test failed, but with the recent Dockerfile update, we have switched to running in python3 now... which have broken a few things in the main.yml.
> I have fixed these in my new PR regarding fio in VMs

@mulbc Looking at the output it is error-ing when grepping in the logs. This grep has recently become redundant since we check ES for a definitive success and I've opened (https://github.com/cloud-bulldozer/ripsaw/pull/356) to remove them. That said, I don't know if that will fix the issue or just move the error elsewhere. I'll re-test this again once 356 is merged.

mulbc:
Yes this is still relevant and I'm waiting on the tests to pass so that @dry923 can merge it

mulbc:
@dry923 - can you make the output of the tests available to me so I can see that this gets merged?

dry923:
@mulbc apologies for the delay. We had some major hardware issues and the CI env was down and needed a rebuild. I'm going to re-run this now

roles/fio_distributed/templates/client.yaml
```diff
@@ -36,25 +36,27 @@ spec:
 {% endif %}
         command: ["/bin/sh", "-c"]
         args:
-          - "cat /tmp/host/hosts;
-{% if workload_args.prefill is defined and workload_args.prefill is sameas true %}
-             echo ***************Prefill*****************;
-             cat /tmp/fio/fiojob-prefill;
-             mkdir -p /tmp/fiod-{{ uuid }}/fiojob-prefill;
-             fio --client=/tmp/host/hosts /tmp/fio/fiojob-prefill --output-format=json;
+          - cat /tmp/host/hosts;
```
dry923, line 39: @mulbc after testing this a few times this change does not work when adding each item to its own -'d line as it stands. It would only run correctly when listed out like before, in a "" list. If this is something you want to keep working on that's fine however it does not pass ci like this.

roles/fio_distributed/templates/client.yaml
```diff
@@ -70,15 +72,23 @@ spec:
...
+{% if metadata is defined and (metadata.collection|default(false)) and (metadata.targeted|default(true)) %}
       initContainers:
       - name: backpack-{{ trunc_uuid }}
         image: {{ metadata_image | default('quay.io/cloud-bulldozer/backpack:latest') }}
         command: ["/bin/sh", "-c"]
-{% if metadata is defined and metadata.force|default(false) is sameas true %}
-        args: ["python3 stockpile-wrapper.py -s {{ elasticsearch.server }} -p {{ elasticsearch.port }} -u {{ uuid }} -n $my_node_name -N $my_pod_name --redisip {{ bo.resources[0].status.podIP }} --redisport 6379 --force"]
-{% else %}
-        args: ["python3 stockpile-wrapper.py -s {{ elasticsearch.server }} -p {{ elasticsearch.port }} -u {{ uuid }} -n $my_node_name -N $my_pod_name --redisip {{ bo.resources[0].status.podIP }} --redisport 6379"]
+        args:
```
dry923, line 80: @mulbc I like this layout much better than the single line approach that was in there prior. That said it runs into the same errors as above and does not actually execute as is. However, changing it to this did work:

```
        args:
          - "python3
             stockpile-wrapper.py
             -s {{ elasticsearch.server }}
             -p {{ elasticsearch.port }}
             -u {{ uuid }}
             -n $my_node_name
             -N $my_pod_name
             --redisip {{ bo.resources[0].status.podIP }}
             --redisport 6379
{% if metadata is defined and metadata.force|default(false) %}
             --force
{% endif %}
             "
```

roles/fio_distributed/templates/client.yaml
mulbc: Got it - I'll look into this when I have some time and see how I can fix this....

roles/fio_distributed/templates/servers.yaml
rsevilla87: fixed by https://github.com/cloud-bulldozer/ripsaw/pull/383

PR #349, opened 2020-05-26
Fixing if statement for renamed directories

dry923:
Some directories were renamed in a recent PR and the if statements in common.sh were not updated. This resolves the issue.

### DIFF
--- a/tests/common.sh
+++ b/tests/common.sh
@@ -14,16 +14,16 @@ function populate_test_list {
   do
     # Check for changes in roles
     if [[ $(echo ${item} | grep 'roles/fs-drift') ]]; then echo "test_fs_drift.sh" >> tests/iterate_tests; fi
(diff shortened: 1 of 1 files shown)


tests/common.sh
```diff
@@ -15,15 +15,15 @@ function populate_test_list {
     # Check for changes in roles
     if [[ $(echo ${item} | grep 'roles/fs-drift') ]]; then echo "test_fs_drift.sh" >> tests/iterate_tests; fi
     if [[ $(echo ${item} | grep 'roles/uperf-bench') ]]; then echo "test_uperf.sh" >> tests/iterate_tests; fi
```
rsevilla87, line 17: uperf instead uperf-bench

dry923: good catch. Fixed!

PR #352, opened 2020-05-28
ServiceMesh benchmark

rvansa:
Global changes:
* permissions for `benchmark-operator` to create new projects (namespaces) and work with Hyperfoil CRs.
* openssl in image
* image for tests can be customized through `BENCHMARK_OPERATOR_IMAGE` env var

This project does not contain CSV; changes in `deploy` need to be ported there, too.


roles/servicemesh/tasks/main.yml
```diff
@@ -0,0 +1,377 @@
...
+      definition:
+        apiVersion: hyperfoil.io/v1alpha2
+        kind: Hyperfoil
+        metadata:
+          name: '{{ meta.name }}-hyperfoil-{{ trunc_uuid }}'
+          namespace: '{{ operator_namespace }}'
+          ownerReferences: "{{ owner }}"
+        spec:
+          # TODO!
+          version: 0.9-SNAPSHOT
```
jtaleric, line 67: Should this be a var that we default to `0.9-SNAPSHOT` but allow it to be overridden. That way we don't have to rebuild the operator when the version increments?

roles/servicemesh/tasks/main.yml
```diff
@@ -0,0 +1,377 @@
...
+    k8s_info:
+      api_version: v1
+      kind: Job
+      namespace: '{{ operator_namespace }}'
+      name: "{{ meta.name }}-{{ trunc_uuid }}"
+    register: job_state
+
+  - name: Complete benchmark
+    when: job_state.resources[0].status.succeeded is defined and (job_state.resources[0].status.succeeded | int) > 0
+#    when: job_state | json_query('resources[0].status.conditions[]') | selectattr('type','match','Complete') | selectattr('status', 'match', 'True') | list | length
```
jtaleric, line 310: clean up

README.md
```diff
@@ -17,6 +17,7 @@ a performance baseline of Kubernetes cluster on your provider.
 | [Smallfile](docs/smallfile.md) | Storage IO Performance | Working            |  Used, default : 3second  | Not Supported         |
 | [fs-drift](docs/fs-drift.md)   | Storage IO Longevity   | Working            |  Not used                  | Not Supported         |
 | [hammerdb](docs/hammerdb.md)   | Database Performance   | Working            |  Not used                  | Not Supported         |
+| [Service Mesh](docs/servicemesh.md) | Microservices     | Working            |  ?                         | Not Supported         |
```
jtaleric, line 20: Looking through the code it seems that Reconciliation would be used, since you are maintaining state... so the default 3 second.

rvansa:
Thanks for the review. I've released Hyperfoil 0.9 so that I can just set version to `latest`, but quay.io is currently down so the image can't be pushed. I'll try to do a build later today.

deploy/60_self_provisioner.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
dry923, line 1: @aakarshg @jtaleric What are your thoughts on this being in the default deploy directory versus having it be an extra thing you can apply in resources like the backpack role, etc? I'm not necessarily against the self-provisioner role however it would be extending the default permission set so making it more obviously optional might be a good idea.

roles/servicemesh/tasks/main.yml
```diff
@@ -0,0 +1,375 @@
...
+    k8s_info:
+      api_version: v1
+      kind: Pod
+      namespace: '{{ operator_namespace }}'
+      label_selectors:
+      - app = {{ meta.name }}-hyperfoil-{{ trunc_uuid }}
+      - role = controller
+    register: hf_pod
+
+  - name: "Create job"
```
dry923, line 243: Can we move this into a template file?

deploy/60_self_provisioner.yaml
aakarshg: I think its a good idea to move this to resources, so that it's not auto applied. @rvansa can you please move it to resources dir?

rvansa:
Updated. Also, HF 0.9 is in quay.io so this should be testable.

What about the CI? How can I tell it to pick up the requirements?

dry923:
> Updated. Also, HF 0.9 is in quay.io so this should be testable.
> 
> What about the CI? How can I tell it to pick up the requirements?

@rvansa What kind of resources are required for service mesh to be stood up for CI? If it is something lightweight we could just add it to the test script. If it needs to be more static you'll need to give me some direction on the best way to install it on something like minikube or crc (if possible) as those are what our CI uses atm.

rvansa:
@dry923 The operator itself it quite lightweight; without the `servicemeshcontrolplane` which is installed only for the duration of the test it is just go-based operator (right now in my cluster it's 32 MB in RSS, no resource limits). Then there's a daemonset with pod just installing CNI on each node. Hyperfoil operator takes 55 MB of RSS in my cluster.

However based on our talk last Wednesday I understand that benchmark-operator should *not* install another operators.

Regarding installation, both of these are available on the OperatorHub for Openshift. I am not sure if what I mean by CI is what you mean here - I was told that you're running certain tests periodically on a production grade environments, really testing for regressions in Openshift performance.

dry923:
> @dry923 The operator itself it quite lightweight; without the `servicemeshcontrolplane` which is installed only for the duration of the test it is just go-based operator (right now in my cluster it's 32 MB in RSS, no resource limits). Then there's a daemonset with pod just installing CNI on each node. Hyperfoil operator takes 55 MB of RSS in my cluster.
> 
> However based on our talk last Wednesday I understand that benchmark-operator should _not_ install another operators.
> 
> Regarding installation, both of these are available on the OperatorHub for Openshift. I am not sure if what I mean by CI is what you mean here - I was told that you're running certain tests periodically on a production grade environments, really testing for regressions in Openshift performance.

CI in this case was meant to be our internal test system that we run all the ripsaw workloads against to verify they work before merging a PR. So in this case I'm looking to find the best route to getting it stood up on a simplified environment to verify this PR and subsequent ones work with service mesh. Does that help?

rvansa:
Okay, so just a functional check. So in this case my major concern would be rather than resource utilization the fact that both operators are Openshift-specific and therefore you can't use minikube (SM is tightly coupled to OCP, HF is pretty simple but it uses Route resource and it was not tested on vanilla Kubernetes). I haven't had the chance to run CRC - hopefully it works the same way as full size OCP.

rvansa:
Is there anything else I should to with regards to this PR?

dry923:
> Is there anything else I should to with regards to this PR?

@rvansa You are good for the moment. I'm working on getting the CI env sorted for this. Once it passes that we'll be good to go.

rvansa:
@dry923 Hi, what's the state of this PR? Can you integrate it (when I fix the git conflict)?

rvansa:
@jtaleric Hi, I've rebased the PR to fix the conflict.

The last CI results don't include service mesh test; I don't know if you want to have that in.

jtaleric:
@dry923 @rsevilla87 Can I get some eyes here?

roles/servicemesh/tasks/main.yml
```diff
@@ -0,0 +1,343 @@
...
+    k8s:
+      definition: "{{ lookup('template', 'routing.yaml.j2') | from_yaml }}"
+
+  - name: Fetch pods
+    k8s_info:
+      api_version: v1
+      kind: Pod
+      namespace: '{{ workload_namespace }}'
+      label_selectors:
+      - app = app
```
rsevilla87, line 191: Small nit: k8s_info module accepts a field_selectors: argument. You can add
```
    field_selectors:
      - status.containerStatuses.ready=true
```

To simply the set_fact task below:
```
 ready_containers: "{{ pods.resources|length }}"
```

rvansa: @rsevilla87 Thanks, thought IIUC that wouldn't work 100% - such field selector would probably select pods that have at least one ready container, while I need both containers to be ready. Or do I misunderstand?

rsevilla87: Absolutely, I wasn't aware there are pods having several containers here.

rsevilla87:
@rvansa , can you rebase please?

rvansa:
@rsevilla87 Rebased.

roles/servicemesh/tasks/main.yml
```diff
@@ -0,0 +1,343 @@
...
+        apiVersion: project.openshift.io/v1
+        kind: ProjectRequest
+        metadata:
+          name: "{{ controlplane_namespace }}"
+
+  - name: Create ServiceMesh control plane
+    k8s:
+      definition: "{{ lookup('template', 'smcp.yaml.j2') | from_yaml }}"
+
+  - name: Create workload namespace
```
jtaleric, line 36: Instead of having the operator create the namespace, can we just not have the prereq to have this namespace already created?

roles/servicemesh/tasks/main.yml
```diff
@@ -0,0 +1,343 @@
...
+    owner:
+      - apiVersion: ripsaw.cloudbulldozer.io/v1alpha1
+        kind: Benchmark
+        name: "{{ meta.name }}"
+        namespace: "{{ operator_namespace }}"
+        uid: "{{ resource_state.resources[0].metadata.uid }}"
+
+- when: resource_state.resources[0].status.state is not defined
+  block:
+  - name: Create control plane namespace
```
jtaleric, line 24: Can we just have this be something the user provides? `ie in the CR provide us with the namespace for the control plane? Ensure it has this as part of the RBAC?`

roles/servicemesh/tasks/main.yml
```diff
@@ -0,0 +1,343 @@
+---
+- name: Get current state
```
jtaleric, line 2: I wonder if we can test to see if `Kind: ServiceMesh , ServiceMeshMemberRoll ` is even something we can create before we jump through the tasks... To catch if the user has not deployed ServiceMesh?

roles/servicemesh/tasks/main.yml
jtaleric: Reading through your commit again, it might be necessary for service mesh... Maybe we can have you demo this to us, this week @rvansa ?

roles/servicemesh/tasks/main.yml
rvansa: Failing fast is definitely a good thing. Do you know if the k8s_info allows you to query if certain CRD is installed?

roles/servicemesh/tasks/main.yml
rvansa: My take on this is that what can be automated should be automated. ProjectRequest RBAC is not something too security sensitive, so why not give the right to the operator instead of having it as prereq?

roles/servicemesh/tasks/main.yml
rvansa: Again, that would be another step backwards from automation.

Regarding a demo, this week I don't have OCP installation available (our scalelab allocation has expired and we're in the process of reinstalling OCP in our own lab). Have you tried to install SM and HF operators from the catalog and just start the benchmark?

roles/servicemesh/tasks/main.yml
rsevilla87: Something like this should work:
```yaml
- name: Get CRDs
  k8s_info:
    api_version: apiextensions.k8s.io/v1
    kind: CustomResourceDefinition
    name: apiservers.config.openshift.io
  register: apiservers
```

rvansa:
@jtaleric Hi, are you waiting only for the fail-fast check for SM installation (before integrating this PR), or is there anything else I should change?

jtaleric:
> @jtaleric Hi, are you waiting only for the fail-fast check for SM installation (before integrating this PR), or is there anything else I should change?

Not particularly, i think a demo would be very helpful for the team to see... We have AWS deployments we could share with you, fit hat would help?

roles/servicemesh/tasks/main.yml
dry923: I'm ok w/ this given the user has to make the choice to allow the perms and apply the crd. So there is an implied responsibility on the user

tests/common.sh
```diff
@@ -117,7 +117,10 @@ function pod_count () {
 
 function apply_operator {
   operator_requirements
-  kubectl apply -f resources/operator.yaml
+  BENCHMARK_OPERATOR_IMAGE=${BENCHMARK_OPERATOR_IMAGE:-"quay.io/benchmark-operator/benchmark-operator:master"}
```
dry923, line 120: This seems redundant with the update operator image function below where it patches together the image based on environment variables.

From top of common.sh
```
image_location=${RIPSAW_CI_IMAGE_LOCATION:-quay.io}
image_account=${RIPSAW_CI_IMAGE_ACCOUNT:-rht_perf_ci}
```

tests/test_servicemesh.sh
```diff
@@ -0,0 +1,55 @@
...
+    error
+  fi
+
+  echo "Cleaning up ServiceMesh"
+  long_uuid=$(get_uuid 0)
+  uuid=${long_uuid:0:8}
+  kubectl delete --ignore-not-found=true namespace example-benchmark-controlplane-$uuid
+  kubectl delete --ignore-not-found=true namespace example-benchmark-workload-$uuid
+  kubectl delete -f tests/test_crs/valid_servicemesh.yaml
+  delete_operator
```
dry923, line 18: can we sub in the wait_clean function instead to also clean up the namespace etc.

roles/servicemesh/tasks/main.yml
```diff
@@ -0,0 +1,343 @@
...
+  - name: Get ingress gateway pods
+    k8s_info:
+      api_version: v1
+      kind: Pod
+      namespace: "{{ controlplane_namespace }}"
+      label_selectors:
+      - istio = ingressgateway
+    register: ingressgateways
+
+  - name: Reboot ingress gateway(s)
```
rsevilla87, line 168: I need some background here. Why do you reboot pods?

rsevilla87:
> A few minor nits really but otherwise I'm good with this. Can you rebase to fix the conflicting file? I do agree with @jtaleric re a demo. If you can find time it would be awesome to show to the team.
> 
> @rsevilla87 mind giving this another pass?

I'm good with this one as well. However, tests are going to be useless for the time being since test_list is not updated, which is a good thing because we currently don't have any CI environment with ServiceMesh installed in it.

roles/servicemesh/tasks/main.yml
rvansa: We've just created a new secret that's optionally bound to ingress gw's filesystem. Even if Openshift makes it appear on the filesystem, AFAIK Envoy (the GW) won't pick it up so we need to reboot it. It might be possible to send it some kind of signal, but just deleting the pods is the most reliable way in my experience (these get re-created by the deployment).

PR #355, opened 2020-05-29
Fio in vms

mulbc:
Adds kubevirt option to fio-distributed workload.
This spawns kubevirt VMs where the fio server is run in.

Please first merge https://github.com/cloud-bulldozer/ripsaw/pull/353 before merging this
  --> I will rebase after this is done

@bengland2 @jeniferh

### DIFF
--- a/resources/crds/ripsaw_v1alpha1_fio_distributed_vm_cr.yaml
+++ b/resources/crds/ripsaw_v1alpha1_fio_distributed_vm_cr.yaml
@@ -1,0 +1,114 @@
+apiVersion: ripsaw.cloudbulldozer.io/v1alpha1
+kind: Benchmark
+metadata:
+  name: fio-vm-benchmark
+  namespace: my-ripsaw
+spec:
+#  elasticsearch:
+#    server: <es_server>
+#    port: 9200
+#  clustername: myk8scluster
+#  test_user: ripsaw
+  workload:
+    name: "fio_distributed"
+    args:
+      # if true, do large sequential write to preallocate volume before using
+      prefill: true
+      # number of times each test
+      samples: 3
+      # number of fio pods generating workload
+      servers: 3
+      # put all fio pods on this server
+      pin_server: ''
+      # Run fio servers in VMs
+      kind: vm
+      # Specify kubevirt registry image for server vm
+      # This is helpful in disconnected environments and we assume
+      # this VM image has all requirements already present
+      vm_image: quay.io/mulbc/fed-fio
+      # CPU cores that will be available inside the VM
+      # Default: 1
+      # vm_cores: 2
+      # Memory that will be available inside the VM
+      # Default: 5G
+      # vm_cores: 10G
+      # test types, see fio documentation
+      jobs:
+        - write
+        - read
+        - randwrite
+        - randread
+      # I/O request sizes (also called block size)
+      bs:
+        - 4k
+        - 64k
+      # how many fio processes per pod
+      numjobs:
+        - 1
+      # with libaio ioengine, number of in-flight requests per process
+      iodepth: 4
+      # how long to run write tests, this is TOO SHORT DURATION
+      read_runtime: 15
+      # how long to run read tests, this is TOO SHORT DURATION
+      write_runtime: 15
+      # don't start measuring until this many seconds pass, for reads
+      read_ramp_time: 5
+      # don't start measuring until this many seconds pass, for writes
+      write_ramp_time: 5
+      # size of file to access
+      filesize: 2GiB
+      # interval between i/o stat samples in milliseconds
+      log_sample_rate: 3000
+      storageclass: ocs-storagecluster-ceph-rbd
+      storagesize: 5Gi
+      #rook_ceph_drop_caches: True
+      #rook_ceph_drop_cache_pod_ip: 192.168.111.20
+#######################################
+#  EXPERT AREA - MODIFY WITH CAUTION  #
+#######################################
+#  global_overrides:
+#  NOTE: Dropping caches as per this example can only be done if the
+#        fio server is running in a privileged pod
+#    - exec_prerun=bash -c 'sync && echo 3 > /proc/sys/vm/drop_caches'
+  job_params:
+    - jobname_match: write
+      params:
+        - fsync_on_close=1
+        - create_on_open=1
+        - runtime={{workload_args.write_runtime }}
+        - ramp_time={{workload_args.write_ramp_time }}
+    - jobname_match: read
+      params:
+        - time_based=1
+        - runtime={{workload_args.read_runtime }}
+        - ramp_time={{workload_args.read_ramp_time }}
+    - jobname_match: rw
+      params:
+        - rwmixread=50
+        - time_based=1
+        - runtime={{workload_args.read_runtime }}
+        - ramp_time={{workload_args.read_ramp_time }}
+    - jobname_match: readwrite
+      params:
+        - rwmixread=50
+        - time_based=1
+        - runtime={{workload_args.read_runtime }}
+        - ramp_time={{workload_args.read_ramp_time }}
+    - jobname_match: randread
+      params:
+        - time_based=1
+        - runtime={{workload_args.read_runtime }}
+        - ramp_time={{workload_args.read_ramp_time }}
+    - jobname_match: randwrite
+      params:
+        - time_based=1
+        - runtime={{workload_args.write_runtime }}
+        - ramp_time={{workload_args.write_ramp_time }}
+    - jobname_match: randrw
+      params:
+        - time_based=1
+        - runtime={{workload_args.write_runtime }}
+        - ramp_time={{workload_args.write_ramp_time }}
+#    - jobname_match: <search_string>
+#      params:
+#        - key=value
--- a/roles/fio_distributed/templates/server_vm.yml.j2
+++ b/roles/fio_distributed/templates/server_vm.yml.j2
@@ -1,0 +1,70 @@
+---
+apiVersion: kubevirt.io/v1alpha3
+kind: VirtualMachineInstance
+metadata:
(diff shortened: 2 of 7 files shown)


mulbc:
FYI - I have tested that this code works for the pod and vm fio workloads using the supplied CRs in the resource folder

jtaleric:
With the continued addition of kubevirt workloads, we need to have an option to verify these sort of patches... @dry923 thoughts?

roles/fio_distributed/templates/server_vm.yml.j2
```diff
@@ -0,0 +1,63 @@
...
+apiVersion: kubevirt.io/v1alpha3
+kind: VirtualMachineInstance
+metadata:
+  name: 'fio-server-{{item | string}}-{{ trunc_uuid }}'
+  namespace: '{{ operator_namespace }}'
+  labels:
+    app: fio-benchmark-{{ trunc_uuid }}
+spec:
+  domain:
+    devices:
```
jeniferh, line 11: Unless a 1vcpu guest is always sufficient for the fio server, it would be good to add vcpu statements and make it configurable:
spec:
  domain:
    cpu:
      sockets: {{ vcpus }} 

Making cpus configurable is not as important for pods since they will naturally burst and not be constrained by a single vcpu process in the host by default.

resources/crds/ripsaw_v1alpha1_fio_distributed_vm_cr.yaml
```diff
@@ -0,0 +1,106 @@
...
+      # number of times each test
+      samples: 3
+      # number of fio pods generating workload
+      servers: 3
+      # put all fio pods on this server
+      pin_server: ''
+      # Run fio servers in VMs
+      kind: vm
+      # Specify kubevirt registry image for server vm
+      # image_vm:
```
netzzer, line 26: Specifying diff image and image location did not work using image_vm:. The default image download was tried had to edit pod to get desired image to be used.

roles/fio_distributed/templates/server_vm.yml.j2
```diff
@@ -0,0 +1,63 @@
...
+          chpasswd: { expire: False }
+          bootcmd:
+{% if workload_args.storageclass is defined
+    or hostpath is defined %}
+            - "mkdir -p {{ fio_path }} || true"
+            - mkfs.ext4 /dev/disk/by-id/ata-QEMU_HARDDISK_data
+            - "mount /dev/disk/by-id/ata-QEMU_HARDDISK_data {{ fio_path }}"
+{% endif %}
+          runcmd:
+            - dnf install -y fio wget curl python3 python3-pip
```
netzzer, line 41: Installing these packages in disconnected OCP env fails. Would be better if VM image had all necessary packages installed.

roles/fio_distributed/templates/server_vm.yml.j2
```diff
@@ -0,0 +1,63 @@
...
+            - cd /tmp
+{% if metadata is defined
+    and (metadata.collection|default(false))
+    and (metadata.targeted|default(true)) %}
+            - export ES_SERVER=1.1.1.1
+            - export ES_PORT=1111
+            - wget https://raw.githubusercontent.com/cloud-bulldozer/bohica/master/stockpile-wrapper/stockpile-wrapper.py
```
netzzer, line 48: This most likely will fail for disconnected OCP env.

roles/fio_distributed/templates/server_vm.yml.j2
```diff
@@ -0,0 +1,63 @@
...
+            - pip3 install elasticsearch-dsl openshift kubernetes redis
```
netzzer, line 49: This most likely will fail for disconnected OCP env.

resources/crds/ripsaw_v1alpha1_fio_distributed_vm_cr.yaml
mulbc: yes you're right, I called it `vm_image` in the code... will fix the CR

roles/fio_distributed/templates/server_vm.yml.j2
mulbc: Yes you're right... I've added a check if `vm_image` is defined... if so, it will not try to do these steps (and we assume the requirements are present)

roles/fio_distributed/templates/server_vm.yml.j2
mulbc: Yes you're right... I've added a check if `vm_image` is defined... if so, it will not try to do these steps (and we assume the requirements are present)

roles/fio_distributed/templates/server_vm.yml.j2
mulbc: Yes you're right... I've added a check if vm_image is defined... if so, it will not try to do these steps (and we assume the requirements are present)

roles/fio_distributed/templates/server_vm.yml.j2
mulbc: I've added an option called `vm_cores` that will configure the amount of cores available to the VM... I assume you meant `cores:` and not `sockets:`?

jeniferh: It may not matter much for this type of test, but a general recommendation is to use sockets -- that is what "regular" kvm vcpus default to, a 'sockets' vcpu topology enables a guest sched_domain that has a sync wake sched hint that can help some types of workloads that do message passing, context switching, etc.

roles/fio_distributed/templates/server_vm.yml.j2
```diff
@@ -0,0 +1,69 @@
...
+      cloudInitNoCloud:
+        userData: |-
+          #cloud-config
+          password: fedora
+          chpasswd: { expire: False }
+          bootcmd:
+{% if workload_args.storageclass is defined
+    or hostpath is defined %}
+            - "mkdir -p {{ fio_path }} || true"
+            - mkfs.ext4 /dev/disk/by-id/ata-QEMU_HARDDISK_data
```
rsevilla87, line 39: Do we want to enforce ext4?

mulbc: Good comment... Actually we should make it possible to make a choice here for either a formatted drive with a choice of Filesystem OR using the drive as a bare block device...
Just didn't have time to implement this choice in this very first commit ;)

mulbc: @rsevilla87 can we continue with the PR as is, or do you need me to make the FS configurable?
IMO this is a secondary item that can be implemented by someone who actually needs this, but if you want this, I'll invest the time to get the PR merged

roles/fio_distributed/templates/server_vm.yml.j2
```diff
@@ -0,0 +1,69 @@
...
+            # python3 stockpile-wrapper.py -s {{ elasticsearch.server }} -p {{ elasticsearch.port }} -u {{ uuid }} -n $my_node_name -N $my_pod_name --redisip {{ bo.resources[0].status.podIP }} --redisport 6379 --force
+{% endif %}
+            - fio --server
+{% if workload_args.storageclass is defined %}
+    - name: data-volume
+      persistentVolumeClaim:
+        claimName: claim-{{ item }}-{{ trunc_uuid }}
+{% elif hostpath is defined %}
+    - name: data-volume
+      hostPath:
```
rsevilla87, line 66: According to kubevirt docs, hostpath volume type doesn't exist here. In this context they are called hostDisk and it's options are slightly different from k8s's hostpath volumes.

roles/fio_distributed/templates/server_vm.yml.j2
rsevilla87: Ok, makes sense, it could be implemented in the future.

rsevilla87:
@mulbc can you also update the documentation to include how to execute this type of benchmarks?

roles/fio_distributed/templates/server_vm.yml.j2
mulbc: Good catch! Will move this to a `hostDisk` instead!

mulbc:
Made some further modifications and also all the mentioned changes. Added the new workload parameters to the doc and there is an example CR in the crds folder.

rsevilla87:
@dry923 @jtaleric imo we can move forward with this one as it's not changing anything from the standard fio workload, CI stuff can be implemented later.

roles/fio_distributed/templates/server_vm.yml.j2
```diff
@@ -0,0 +1,70 @@
...
+{% endif %}
+            - cd /tmp
+{% if workload_args.metadata is defined
+    and (workload_args.metadata.collection|default(false))
+    and (workload_args.metadata.targeted|default(true)) %}
+            - export ES_SERVER={{ elasticsearch.server }}
+            - export ES_PORT={{ elasticsearch.port }}
+{% if workload_args.vm_image is undefined %}
+            - wget https://raw.githubusercontent.com/cloud-bulldozer/bohica/master/stockpile-wrapper/stockpile-wrapper.py
+            - pip3 install elasticsearch-dsl openshift kubernetes redis
```
jtaleric, line 54: Do we not want to pin to specific versions? Not worth blocking for, but maybe we should just install https://github.com/cloud-bulldozer/snafu/blob/master/requirements.txt

dry923: +1 to @jtaleric 's comment. Not going to block on it though

dry923:
@jeniferh if you can give this a rubber stamp we should be good to merge.

jeniferh:
LGTM with one small nit: would be good to list the optional hostpath workload arg in the docs (or easy fix later..)
Thanks for adding the vcpu options !

PR #357, opened 2020-06-01
Allow passing of prometheus elasticsearch information 

dry923:
This PR adds variables to be used in snafu for passing prometheus ES variables. This also adds the parallel variable to elasticsearch.

### DIFF
--- a/docs/prometheus.md
+++ b/docs/prometheus.md
@@ -1,0 +1,68 @@
+# Collecting Prometheus Data
+
+How To:
+* [What is it](#what-is-it)
+* [Configuring](#configuring)
+
+# What is it
+
+The ability to capture prometheus data during a benchmark run was added to the [Benchmark-wrapper](https://github.com/cloud-bulldozer/benchmark-wrapper). 
+To enable this functionality and pass prometheus data to an Elasticsearch server the workload CR must be appropriately configured.
+
+
+# Configuring
+
+To enable this functionality a few variables must be set in the workload CR file.
+
+```
+prometheus:
+  server: the elasticsearch server to upload to
+  port: the elasticsearch server port
+  prom_url: the prometheus URL
+  prom_token: a valid access token for prometheus
+  parallel: enable parallel uploads to elasticsearch
+```
+
+The prometheus token can be obtained by running the following.
+
+```
+$ oc -n openshift-monitoring sa get-token prometheus-k8s
+```
+
+Enabling prometheus data capture may greatly increase the load on the Elasticsearch server that is being used. Please be aware and have the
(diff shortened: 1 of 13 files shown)


acalhounRH:
Good to go, testing PR with Snafu PR and worked.

dry923:
removing stale label. needs review @jtaleric @rsevilla87

bengland2:
aakarshg should not be a pending reviewer!  Is he pending on any other PRs at this point?

dry923:
> lacking docs to explain the new vars and how to capture them.

Updated docs. Please have a look @jtaleric

docs/prometheus.md
```diff
@@ -0,0 +1,68 @@
...
+
+To enable this functionality a few variables must be set in the workload CR file.
+
+```
+prometheus:
+  server: the elasticsearch server to upload to
+  port: the elasticsearch server port
+  prom_url: the prometheus URL
+  prom_token: a valid access token for prometheus
+  parallel: enable parallel uploads to elasticsearch
```
rsevilla87, line 23: Can you mention that the "parallel" argument is also available in the metadata collection?

roles/fio_distributed/templates/client.yaml
```diff
@@ -33,6 +33,22 @@ spec:
             value: "{{ elasticsearch.index_name | default("ripsaw-fio") }}"
           - name: es_verify_cert
             value: "{{ elasticsearch.cert_verify | default("true") }}"
+          - name: parallel
+            value: "{{ elasticsearch.parallel | default("false") }}"
+{% endif %}
+{% if prometheus is defined %}
```
rsevilla87, line 39: I think the second conditional is not required here. All values can be set under the first conditional.
  
```suggestion
{% if prometheus is defined %}
          - name: prom_es
            value: "{{ prometheus.server }}"
          - name: prom_port
            value: "{{ prometheus.port }}"
          - name: prom_parallel
            value: "{{ prometheus.parallel | default("false") }}"
          - name: prom_token
            value: "{{ prometheus.prom_token }}"
          - name: prom_url
            value: "{{ promethues.prom_url }}"
{% endif %} 
```

In addition, prom_token and prom_url should be replaced by prometheus.prom_token and prometheus.prom_url

docs/prometheus.md
dry923: I thought about that but really didnt know where it would be best to put it. We should probably just duplicate this doc for the ES section. I'll add it in a few.

roles/fio_distributed/templates/client.yaml
dry923: good catch. thanks!

PR #359, opened 2020-06-11
Add vegeta benchmark

rsevilla87:
Adding vegeta benchmark. It accepts CRDs like the following:

```yaml 
kind: Benchmark
metadata:
  name: vegeta-benchmark
  namespace: my-ripsaw
spec:
  elasticsearch:              <-- Optional
    server: <es_instance>
    port: 9200
    index_name: ripsaw-vegeta
  workload:
    name: vegeta
    args:
      hostnetwork: true                     <-- Optional. Defaults to false
      image: quay.io/rsevilla/vegeta:latest           <-- Optional. Defaults to quay.io/cloud-bulldozer/vegeta:latest
      clients: 2  <-- Optional. Defaults to 1
      targets:
        - name: 100w-ka
          urls:
            - GET https://mydomain.com/test.png
            - GET http://myunsecuredomain.com
          samples: 2   <-- Optional. Defaults to 1
          workers: 100              <- Optonal. Defaults to 1
          duration: 10
          keepalive: true           <-- Optional. Defaults to false
        - name: 20w
          urls:
            - GET https://domain.com/api/endpoint
          samples: 1
          workers: 20
          duration: 10
          keepalive: false
```

With the above Benchmark object. Ripsaw will kick off two vegeta jobs that will run in parallel ( redis is used to synchronize the processes). These pods use anti-affinity rules to be preferably deployed on different nodes.

The targets list allows to trigger multiple serial tests using different urls and options. Specified URLs accept the syntax described at the Vegeta documentation (https://github.com/tsenart/vegeta#simple-targets).

It doesn't include metadata collection, I think It's not necessary here since the pods deployed are just clients and don't perform any workload. @dry923 @jtaleric  thoughts?

### DIFF
--- a/resources/crds/ripsaw_v1alpha1_vegeta_cr.yaml
+++ b/resources/crds/ripsaw_v1alpha1_vegeta_cr.yaml
@@ -1,0 +1,30 @@
+apiVersion: ripsaw.cloudbulldozer.io/v1alpha1
+kind: Benchmark
+metadata:
+  name: vegeta-benchmark
+  namespace: my-ripsaw
+spec:
+# elasticsearch:
+#   server: <es_instance>
+#   port: 9200
+#   index_name: ripsaw-vegeta
+  workload:
+    name: vegeta
+    args:
+      hostnetwork: false
+      image: quay.io/cloud-bulldozer/vegeta:latest
+      clients: 2
+      targets:
+        - name: 100w-ka
+          urls:
+            - GET https://mydomain.com/test.png
+            - GET http://myunsecuredomain.com
+          samples: 2
+          workers: 100
+          duration: 10
+          keepalive: true
+        - name: 20w
+          urls:
+            - GET https://domain.com/api/endpoint
+          workers: 20
+          duration: 10
--- a/resources/crds/ripsaw_v1alpha1_vegeta_cr.yaml
+++ b/resources/crds/ripsaw_v1alpha1_vegeta_cr.yaml
@@ -14,7 +14,6 @@ spec:
     name: vegeta
     args:
       hostnetwork: false
-      image: quay.io/cloud-bulldozer/vegeta:latest
       clients: 2
       targets:
         - name: 100w-ka
--- a/resources/crds/ripsaw_v1alpha1_vegeta_cr.yaml
+++ b/resources/crds/ripsaw_v1alpha1_vegeta_cr.yaml
@@ -8,6 +8,8 @@ spec:
 #   server: <es_instance>
 #   port: 9200
 #   index_name: ripsaw-vegeta
+# test_user: username_to_attach_to_metadata
+  clustername: myk8scluster
   workload:
     name: vegeta
     args:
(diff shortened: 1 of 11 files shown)


dry923:
@rsevilla87 Can we add the metadata collection bits to the template or do you want me to do that post merge?

rsevilla87:
> @rsevilla87 Can we add the metadata collection bits to the template or do you want me to do that post merge?

Done!

resources/crds/ripsaw_v1alpha1_vegeta_cr.yaml
```diff
@@ -0,0 +1,32 @@
...
+#   server: <es_instance>
+#   port: 9200
+#   index_name: ripsaw-vegeta
+# test_user: username_to_attach_to_metadata
+  clustername: myk8scluster
+  workload:
+    name: vegeta
+    args:
+      hostnetwork: false
+      image: quay.io/cloud-bulldozer/vegeta:latest
```
dry923, line 17: Do we need this image line here as its defaulted to ^ in the template

tests/test_crs/valid_vegeta.yaml
```diff
@@ -0,0 +1,32 @@
...
+  elasticsearch:
+    server: marquez.perf.lab.eng.rdu2.redhat.com
+    port: 9200
+    index_name: ripsaw-vegeta
+  metadata:
+    collection: true
+  workload:
+    name: vegeta
+    args:
+      image: quay.io/cloud-bulldozer/vegeta:latest
```
dry923, line 17: Same comment as above re image definition is redundant.

tests/test_crs/valid_vegeta_hostnetwork.yaml
```diff
@@ -0,0 +1,28 @@
...
+    server: marquez.perf.lab.eng.rdu2.redhat.com
+    port: 9200
+    index_name: ripsaw-vegeta
+  metadata:
+    collection: true
+    serviceaccount: benchmark-operator
+  workload:
+    name: vegeta
+    args:
+      image: quay.io/cloud-bulldozer/vegeta:latest
```
dry923, line 18: And again :-)

dry923:
outside of my minor nit it looks good! Good work :-D

tests/test_crs/valid_vegeta_hostnetwork.yaml
rsevilla87: makes sense. Fixed!

tests/common.sh
```diff
@@ -275,12 +277,12 @@ function wait_for_backpack() {
 function check_es() {
```
jtaleric, line 277: Why did this change?

rsevilla87: You're right, this change is not related with the goal of the PR. In the other hand, it fixes this function, since we call it from CI and we evaluate its return code, this was obviously not happening with exit, so I changed it to return.

tests/test_crs/valid_vegeta_hostnetwork.yaml
```diff
@@ -0,0 +1,27 @@
+---
+apiVersion: ripsaw.cloudbulldozer.io/v1alpha1
+kind: Benchmark
+metadata:
+  name: vegeta-benchmark-hostnetwork
+  namespace: my-ripsaw
+spec:
+  elasticsearch:
+    server: marquez.perf.lab.eng.rdu2.redhat.com
```
dry923, line 9: Same as above - This needs to be updated to be ES_SERVER and ES_PORT for the elastic server/port with the new CI build. It passes those vars at runtime.

tests/common.sh
dry923: @rsevilla87 Also, not sure if this caused any issues. Your last PR test should have failed since it couldn't reach the ES server. Any chance the changes to check_es borked that?

PR #361, opened 2020-06-16
service port should match with open server ports

mohit-sheth:
20012 will not be included and will match

https://github.com/cloud-bulldozer/ripsaw/blob/57142bbc053970f2513c19480a8a741f2e13f5f2/roles/uperf/templates/server.yml.j2#L36
with this we can avoid uperf errors
```
** TCP: Cannot connect to 172.30.225.73:20011 No route to host

WARNING: Errors detected during run
```


rsevilla87:
Good catch @mohit-sheth !

PR #362, opened 2020-06-18
enable open and exposed ports to be configured dynamically

mohit-sheth:


mohit-sheth:
@smalleni thanks for pointing it out

PR #364, opened 2020-06-23
Updated the antiaffinity rule to keep server/client pairs separate

dry923:
This keeps uperf server/client pairs on separate hosts if possible.This fixes issue 363.

### DIFF
--- a/roles/uperf/templates/workload.yml.j2
+++ b/roles/uperf/templates/workload.yml.j2
@@ -13,6 +13,7 @@ spec:
     metadata:
       labels:
         app: uperf-bench-client-{{ trunc_uuid }}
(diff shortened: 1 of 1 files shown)


bengland2:
So I've been using basically this PR (slightly different name: field) and its doing what I want.   The anti-affinity at the pair level rather than the node level is the key - I verified that this actually worked perfectly in a 100-pair uperf test, while servers and clients were spread almost evenly across hosts.   I wish servers and clients could be spread more evenly, wouldn't need so many of them then, but that's for another day, see[ this article that mentions controlling the skew in pod count across nodes](https://kubernetes.io/blog/2020/05/introducing-podtopologyspread/).

PR #368, opened 2020-07-06
support elasticsearch indexing with ssl

rsevilla87:
This a rebased version of #316

> 
> this commit is a follow up to the snafu commit
> cloud-bulldozer/snafu#169, to enable indexing to elasticsearch that
> is setup with ssl. also cleaned up elasticsearch related documents,
> and put them in a single new document to allow for easier readability.
> 
> TODO: update backpack to also support indexing to es with ssl.


mulbc:
In the client.yaml code:
```yaml
          - name: es_verify_cert
            value: "{{ elasticsearch.cert_verify | default("true") }}"
```

I think it would make sense to add a `|string` filter so that if someone does a 

```yaml
  elasticsearch:
   cert_verify: false
```

(without quotes) in the CR, it gets re-formatted to a string. Otherwise this will cause the benchmark to hang.

rsevilla87:
Snafu expectes

> In the client.yaml code:
> 
> ```yaml
>           - name: es_verify_cert
>             value: "{{ elasticsearch.cert_verify | default("true") }}"
> ```
> 
> I think it would make sense to add a `|string` filter so that if someone does a
> 
> ```yaml
>   elasticsearch:
>    cert_verify: false
> ```
> 
> (without quotes) in the CR, it gets re-formatted to a string. Otherwise this will cause the benchmark to hang.

Thanks for the suggestion. However, env vars are strings type. Snafu expects "true" or "false" strings, otherwise it would be ignored. https://github.com/cloud-bulldozer/snafu/blob/c5b14fc34a9df3dc147ffdf7b14df3f6b9490c3c/run_snafu.py#L68-L76

In addition K8S api implements this field as a string:

```
$ oc explain pod.spec.containers.env.value
KIND:     Pod
VERSION:  v1

FIELD:    value <string>
```

dry923:
@rsevilla87 cert_verify would need to be added to the ripsaw crd ripsaw/resources/crds/ripsaw_v1alpha1_ripsaw_crd.yaml to allow the new variable.

rsevilla87:
> @rsevilla87 cert_verify would need to be added to the ripsaw crd ripsaw/resources/crds/ripsaw_v1alpha1_ripsaw_crd.yaml to allow the new variable.

Done, I just copied the PR I didn't realize we were missing this.

PR #370, opened 2020-07-16
Allow passing of additional arguments to test.sh

dry923:
Allow passing of max concurrent (parallel), choice of tests, ES server and ES port to test.sh. Omitting the arguments sets them to defaults

Example to set max parallel to 1
```
# bash test.sh -p 1
```

Example to set max parallel to 2 and only run the hammerdb and fs-drift tests
```
# bash test.sh -p 2 -t 'tests/test_hammerdb.sh tests/test_fs_drift.sh'
```

Example to set ES server to foo and ES port to 90
```
# bash test.sh -s foo -x 90
```

This allows the user to still simulate all the work done in the CI environment while targeting specific tests. This will also enable us to potentially take tests as command line flags to /rerun when triggering a rerun of a PR.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -1,23 +1,39 @@
 #!/usr/bin/env bash
 set -x
 
+# The maximum number of concurrent tests to run at one time (0 for unlimited). This can be overriden with -p
+max_concurrent=3
+
+# Presetting test_choice to be blank. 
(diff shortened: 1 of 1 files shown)


jtaleric:
Will merge after CI wraps up..

PR #378, opened 2020-07-31
Fio drop kernel cache

bengland2:
This PR makes fio request a cache drop from a set of K8S hosts in the kernel-cache-dropper daemonset prior to each sample run.   It depends on benchmark-wrapper PR 209, which has been merged.


group_vars/all.yml
```diff
@@ -1,3 +1,4 @@
 ---
 operator_namespace: '{{ meta.namespace }}'
 clustername: 'myk8scluster'
+kernel_cache_drop_svc_port: 9222
```
dry923, line 4: Can this be a default in the fio role instead of being hard coded here? Also, is this going to be user configurable or is it always going to be port 9222

roles/fio_distributed/tasks/main.yml
```diff
@@ -100,7 +106,7 @@
       name: "{{ meta.name }}"
       namespace: "{{ operator_namespace }}"
       status:
-        state: StartingClient
+        state: WaitingForCacheDroppers
```
dry923, line 109: There should be 2 when's to update the status. As it stands we will always drop into the waiting for cache droppers block even if workload_args.drop_cache_kernel is not defined.

group_vars/all.yml
bengland2: the intent was to to make kernel cache dropper a feature that could be used by multiple benchmarks, but that's not the way it's implemented today.   So you mean it could be in roles/fio_distributed/defaults/main.yml or something like that, at least for now?

roles/fio_distributed/tasks/main.yml
bengland2: You are correct that it always enters that state, however, if no cache dropping was requested then it immediately leaves that state (see last task in the block).   This keeps the code path simple.   What's the harm in doing it this way?  More complicated to make it branch to cache drop state only if cache dropping was requested.   Wish this entire thing wasn't a state machine.

group_vars/all.yml
dry923: Would each benchmark be launching its own kernel cache dropper pods?

roles/fio_distributed/tasks/main.yml
dry923: Unless i'm misreading it, it would run all the code from line 121 through 170. which includes creating a configmap, which isn't necessarily a big deal but its creating it for no use.

group_vars/all.yml
dry923: Wondering if it makes sense to have it managed outside the individual benchmarks. Might make sense to treat it kind of like the backpack daemonset option that sits above the launching of the benchmark. So we could check if a var is set and then launch the specified pods. That said I'm guessing we need to know where the fio/other benchmark pods are going to land.

bengland2: @dry923 I like the overall idea.   Specifically I don't like having to put code in the fio benchmark specifically for cache dropping (e.g. finding the pod IPs, formatting them into a list, putting them in a configmap, getting them out of the configmap, sigh).   I did not understand how to pull this out of fio benchmark and make it benchmark-independent, would it be a separate role that fio benchmark would invoke?

roles/fio_distributed/tasks/main.yml
bengland2: @dry923  good point, I'll try and rethink the extra state.

bengland2:
ok, I'll look at it, don't understand why it broke backpack.   But I'm not happy with the implementation , although it works - there is too much to do for each benchmark to use cache dropping, not as simple as I had wanted.

bengland2:
really close now, working through a K8S problem with letting a python kube client ask what nodes have a label

group_vars/all.yml
bengland2: So It now is managed outside the individual benchmarks by a separate role.   Much simpler.

roles/fio_distributed/tasks/main.yml
bengland2: This problem is solved by re-implementing cache dropper as a role outside of any 1 benchmark, and using wait_for_daemonset.py so that no state is needed.

wait_for_daemonset.py
```diff
@@ -0,0 +1,84 @@
...
+    delta_time = now_time - start_time
+    if delta_time > timeout:
+        break
+
+    ret = v1.list_pod_for_all_namespaces(watch=False)
+    matching_pods = 0
+    for i in ret.items:
+        if i.metadata.namespace == ns:
+            if i.metadata.generate_name == generate_name:
+                if i.status.phase == 'Running':
```
rsevilla87, line 74: This can be written as:
```
if i.metadata.namespace == ns and i.metadata.generate_name == generate_name and i.status.phase == 'Running':
```

roles/kernel_cache_drop/tasks/main.yml
```diff
@@ -0,0 +1,32 @@
+---
+- debug:
+    msg: "in cache dropper role"
+
+- name: start kernel cache dropper
+  k8s:
+    definition: "{{ lookup('template', '/opt/kernel_cache_dropper/kernel-cache-drop-daemonset.yaml') | from_yaml }}"
```
rsevilla87, line 7: Moving kernel-cache-drop-daemonset.yaml to the templates dir of the kernel_cache_drop role will avoid to copy it in the Dockerfile and simplify this task

wait_for_daemonset.py
bengland2: sure, that's cleaner.

roles/kernel_cache_drop/tasks/main.yml
bengland2: OK, that makes sense.

wait_for_daemonset.py
bengland2: it is a pain to do this because of pycodestyle-3 indentation rules though!  It doesn't like lines longer than 79 characters, and it doesn't like it if you try to break it into multiple lines, but I found a way.

tests/test_crs/valid_fiod_hostpath.yaml
```diff
@@ -5,8 +5,8 @@ metadata:
   namespace: my-ripsaw
 spec:
   elasticsearch:
-    server: ES_SERVER
-    port: ES_PORT
+    server: "snafu:passwd@ec2-54-201-210-248.us-west-2.compute.amazonaws.com"
+    port: 9200
```
rsevilla87, line 9: These two keys shouldn't be modified. CI replaces these values at https://github.com/cloud-bulldozer/benchmark-operator/blob/master/test.sh#L68-L69

tests/test_fiod.sh
```diff
@@ -44,5 +44,9 @@ function functional_test_fio {
 }
 
 figlet $(basename $0)
+nodelist="$(kubectl get nodes | awk '/Ready/{print $1}')"
+for n in $nodelist ; do
+    kubectl label node $n kernel-cache-dropper=yes
```
rsevilla87, line 49: I've seen CI issues because nodes were already labeled.
```
05:43:19 + kubectl label node ip-10-0-141-67.us-west-2.compute.internal kernel-cache-dropper=yes
05:43:19 error: 'kernel-cache-dropper' already has a value (yes), and --overwrite is false
```

We can fix them by applying the flag --overwrite to kubectl

bengland2:
thank you @rsevilla87 !  I was struggling to figure out what I was doing in my test that was different than the CI.    I'll ask @dry923 if there are objections to this method of resolving the problem, if not then we'll merge it.  Great way to enter the weekend.

bengland2:
I have an idea on how to fix this when I get back to it on Tuesday.  I think it's complaining about my wait_for_daemonset.py getting all the nodes and reading the labels, so I'll try using kubectl to get the list of nodes with the kernel-cache-dropper label, put this into a file, and read the list into wait_for_daemonset.py.   Then wait_for_daemonset.py doesn't have to call v1.list_node() and only has to ask for list of pods running the app name and wait until there are as many of these as there are nodes in the list, hopefully it is allowed to do that!   I'm not happy with security functionality of K8s, the security violation messages don't seem to identify what permission you would need to be able to perform a simple action like v1.list_node().

rsevilla87:
I've detected several issues here, see my findings below:

- In order to list nodes a clusterrole and clusterrolebinding is required since this a cluster resource.
- resources/kernel-cache-drop-daemonset.yaml namespace name must be modified before building the operator container (Not in the ci_dir) and the new namespace name should match the fio namespace since this test is the only one using this feature. my-'s/ripsaw/my-ripsaw-$UUID-test-fiod'
- The line `config.load_kube_config()` from wait_for_daemonset.py should be replaced by `config.load_incluster_config()` so this code can work from a pod.
- wait_for_daemonset.py is listing all nodes, including master nodes. By default, daemonsets are only deployed in worker nodes (the default cluster nodeSelector matches only worker nodes), so the condition of number of cache-dropper pods equals to number of nodes will never be met since no cache-dropper pods will be deployed in the master nodes. (This shouldn't affect to CI tests since the CI cluster is composed by three master/worker nodes)

A fix for the two first issues is available at my fork https://github.com/rsevilla87/ripsaw/tree/378, the third one should be easy to fix :) by applying a labelSelector to the list_node()  call.

bengland2:
Thanks @rsevilla87 .   

My workaround is to use kubectl to get the list of nodes, so wait_for_daemonset.py doesn't have to.  This should solve the first two issues.      

Didn't know about load_incluster_config(), thx.   

One slight correction: wait_for_daemonset.py is waiting for nodes that are labeled with the specified label.   As long as that label is not applied to master nodes when they have a taint, then this should not be a problem.

I'm testing the code now, will drop it in the CI soon and we'll see.

bengland2:
tried your .load_incluster_config(), [which is documented here](https://kubernetes.io/docs/tasks/administer-cluster/access-cluster-api/#accessing-the-api-from-within-a-pod), but got this error, trying with regular load, which got farther.
```
expecting 1 nodes to have kernel cache dropper
  File "roles/kernel_cache_drop/wait_for_daemonset.py", line 48, in <module>
    config.load_incluster_config()
  File "/home/bengland/.local/lib/python3.7/site-packages/kubernetes/config/incluster_config.py", line 94, in load_incluster_config
    cert_filename=SERVICE_CERT_FILENAME).load_and_set()
  File "/home/bengland/.local/lib/python3.7/site-packages/kubernetes/config/incluster_config.py", line 45, in load_and_set
    self._load_config()
  File "/home/bengland/.local/lib/python3.7/site-packages/kubernetes/config/incluster_config.py", line 51, in _load_config
    raise ConfigException("Service host/port is not set.")
kubernetes.config.config_exception.ConfigException: Service host/port is not set.
```

Also, I'm not using kubectl, I'm using k8s_facts module but same idea.

dry923:
> tried your .load_incluster_config(), [which is documented here](https://kubernetes.io/docs/tasks/administer-cluster/access-cluster-api/#accessing-the-api-from-within-a-pod), but got this error, trying with regular load, which got farther.
> 
> ```
> expecting 1 nodes to have kernel cache dropper
> Traceback (most recent call last):
>   File "roles/kernel_cache_drop/wait_for_daemonset.py", line 48, in <module>
>     config.load_incluster_config()
>   File "/home/bengland/.local/lib/python3.7/site-packages/kubernetes/config/incluster_config.py", line 94, in load_incluster_config
>     cert_filename=SERVICE_CERT_FILENAME).load_and_set()
>   File "/home/bengland/.local/lib/python3.7/site-packages/kubernetes/config/incluster_config.py", line 45, in load_and_set
>     self._load_config()
>   File "/home/bengland/.local/lib/python3.7/site-packages/kubernetes/config/incluster_config.py", line 51, in _load_config
>     raise ConfigException("Service host/port is not set.")
> kubernetes.config.config_exception.ConfigException: Service host/port is not set.
> ```
> 
> Also, I'm not using kubectl, I'm using k8s_facts module but same idea.

@bengland2 i'm not seeing in_cluster_config in the current code so you may be testing it locally. I'm using it in my scale PR and the code looks like:
```
            config.load_incluster_config()
            k8s_config = client.Configuration()
            k8s_client = client.api_client.ApiClient(configuration=k8s_config)
            dyn_client = DynamicClient(k8s_client)
```

bengland2:
well my approach completely failed, k8s_info module refused to give me any nodes at all.   So I'll go back to your suggestions folks.

bengland2:
@rsevilla87 So following Russ's suggestion, I reproduced the error in the CI using minikube, running wait_for_daemonset.py from inside the benchmark-operator pod.    This is progress for me because I can now try to fix without running the entire CI to see if it worked.   I added a YAML to make a ClusterRole that applied to the benchmark-operator serviceAccount, which the cache dropper now uses.   The problem seems to go away.   Testing...

bengland2:
The issue now is that I have to remove deploy/40_cluster_role_kubevirt.yaml because it collides with deploy/41_cluster_role.yaml , which is the new yaml that allows wait_for_daemonset.py to work.    If I can't figure out right away, I'll push commits so you can see what I'm seeing.

bengland2:
@dry923  why would test_fiod fail 2 times before it passed in PR 378?  It's done this twice in a row.   Is it not propagating failure up to the CI?  I don't feel confident merging with a result like that.

dry923:
> @dry923 why would test_fiod fail 2 times before it passed in PR 378? It's done this twice in a row. Is it not propagating failure up to the CI? I don't feel confident merging with a result like that.

@bengland2 it hides the output if it passes any of the retries. This was done to avoid clogging up the ci output. I'm rerunning it now w/o any of the other stuff going to see if its just a timeout or what.

dry923:
Appears to be timing out with the kernel drop call:

```
14:04:29 2020-09-08T18:03:04Z - INFO     - MainProcess - run_snafu: logging level is INFO
14:04:29 2020-09-08T18:03:04Z - INFO     - MainProcess - run_snafu: Using elasticsearch server with host:search-cloud-perf-lqrf3jjtaqo7727m7ynd2xyt4y.us-west-2.es.amazonaws.com
14:04:29 2020-09-08T18:03:04Z - INFO     - MainProcess - run_snafu: Using elasticsearch server with port:80
14:04:29 2020-09-08T18:03:04Z - INFO     - MainProcess - run_snafu: Using index prefix for ES:ripsaw-fio
14:04:29 2020-09-08T18:03:09Z - INFO     - MainProcess - run_snafu: Connected to the elasticsearch cluster with info as follows:{'name': '8a53754502eeb862d764feedfc0b4db9', 'cluster_name': '925374498059:cloud-perf', 'cluster_uuid': '6uuXpWTAT9q_9R8GApbrNA', 'version': {'number': '7.1.1', 'build_flavor': 'oss', 'build_type': 'tar', 'build_hash': '7a013de', 'build_date': '2019-09-05T07:25:23.525600Z', 'build_snapshot': False, 'lucene_version': '8.0.0', 'minimum_wire_compatibility_version': '6.8.0', 'minimum_index_compatibility_version': '6.0.0-beta1'}, 'tagline': 'You Know, for Search'}
14:04:29 2020-09-08T18:03:09Z - INFO     - MainProcess - wrapper_factory: identified fio as the benchmark wrapper
14:04:29 2020-09-08T18:03:09Z - INFO     - MainProcess - request_cache_drop: ceph OSD cache drop pod: None
14:04:29 2020-09-08T18:03:09Z - INFO     - MainProcess - request_cache_drop: kernel cache drop pods: 10.130.0.231 10.129.0.85 10.128.6.165 10.128.1.249 
14:04:29 2020-09-08T18:03:09Z - INFO     - MainProcess - request_cache_drop: requesting kernel to drop cache via 10.130.0.231:9222
14:04:29 2020-09-08T18:03:09Z - INFO     - MainProcess - request_cache_drop: requesting kernel to drop cache via 10.129.0.85:9222
14:04:29 Traceback (most recent call last):
14:04:29   File "/usr/local/bin/run_snafu", line 11, in <module>
14:04:29     load_entry_point('snafu', 'console_scripts', 'run_snafu')()
14:04:29   File "/opt/snafu/src/run_snafu.py", line 105, in main
14:04:29     parallel_setting)
14:04:29   File "/opt/snafu/src/utils/py_es_bulk.py", line 166, in streaming_bulk
14:04:29     for ok, resp_payload in streaming_bulk_generator:
14:04:29   File "/usr/local/lib/python3.6/site-packages/elasticsearch/helpers/actions.py", line 212, in streaming_bulk
14:04:29     actions, chunk_size, max_chunk_bytes, client.transport.serializer
14:04:29   File "/usr/local/lib/python3.6/site-packages/elasticsearch/helpers/actions.py", line 63, in _chunk_actions
14:04:29     for action, data in actions:
14:04:29   File "/opt/snafu/src/utils/py_es_bulk.py", line 114, in actions_tracking_closure
14:04:29     for cl_action in cl_actions:
14:04:29   File "/opt/snafu/src/run_snafu.py", line 140, in process_generator
14:04:29     drop_cache()
14:04:29   File "/opt/snafu/src/utils/request_cache_drop.py", line 75, in drop_cache
14:04:29     rsp = conn.getresponse()
14:04:29   File "/usr/lib64/python3.6/http/client.py", line 1346, in getresponse
14:04:29     response.begin()
14:04:29   File "/usr/lib64/python3.6/http/client.py", line 307, in begin
14:04:29     version, status, reason = self._read_status()
14:04:29   File "/usr/lib64/python3.6/http/client.py", line 268, in _read_status
14:04:29     line = str(self.fp.readline(_MAXLINE + 1), "iso-8859-1")
14:04:29   File "/usr/lib64/python3.6/socket.py", line 586, in readinto
14:04:29     return self._sock.recv_into(b)
14:04:29 socket.timeout: timed out
```

bengland2:
going back to snafu, adding a delay after kernel cache drop finishes to allow apps to warmup cache a little before starting workload, and making this configurable.

bengland2:
see [benchmark-wrapper PR 209](https://github.com/cloud-bulldozer/benchmark-wrapper/pull/209)

rsevilla87:
@bengland2 works well now, however, when we run all tests I saw cache drop timeouts as you mentioned before.

rsevilla87 (approved):
LGTM!

dry923 (approved):
LGTM!!

PR #379, opened 2020-08-04
Update backpack_role

rsevilla87:
Update permissions from the backpack role to get every object in the cluster. We're currently missing some metadata such as:

- network-attachment-definitions.k8s.cni.cncf
- networks.operator.openshift.io
- dnses.config.openshift.io

The update looks a bit aggressive, but will make our life easier in the future :)

Signed-off-by: Raul Sevilla rsevilla@redhat.com


dry923:
@rsevilla87 Good for me but it needs a doc update as well as the metadata docs still reference the old resource file.

PR #380, opened 2020-08-04
Adding additional CI test for backpack init containers

dry923:
We currently do not verify all the indexes gathered by the backpack init containers. This adds an additional test to the backpack CI script to verify them.


dry923:
@jtaleric bump for a review and merge

PR #390, opened 2020-08-18
Collect only ocp and k8s related metadata

rsevilla87:
Prevent backpack from trying to collect metadata not related with OCP or k8s.
Also moving the call to backpack_requirements to test_backpack to fix CI issues.

### DIFF
--- a/resources/backpack_role.yaml
+++ b/resources/backpack_role.yaml
@@ -34,19 +34,10 @@ metadata:
   name: backpack-view
   namespace: my-ripsaw
 ---
-apiVersion: v1
-kind: Secret
-metadata:
-  name: backpack-view
-  namespace: my-ripsaw
-  annotations:
-    kubernetes.io/service-account.name: backpack-view
-type: kubernetes.io/service-account-token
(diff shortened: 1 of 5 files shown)


rsevilla87:
FYI: @dry923, find out that the backpack-view clusterrolebinding was being created, and overwritten for each test. That caused  sometimes backpack to fail when it tries to collect k8s metadata

dry923:
> FYI: @dry923, find out that the backpack-view clusterrolebinding was being created, and overwritten for each test. That caused sometimes backpack to fail when it tries to collect k8s metadata

ack. Thanks. I already know how to fix it. I'll PR it shortly.

dry923:
@rsevilla87 if we add in 
```
sed -i "s/backpack_role/backpack_role-$UUID/g" resources/backpack_role.yaml
```

To ~ line 81 in test.sh. It will create a unique clusterrole for each run. This would allow us to potentially run multiple concurrent tests. The we can just clean it up in the tests/test_backpack.sh finish function.

PR #391, opened 2020-08-25
postgresql and mysql features added

mkarg75:
This PR is adding two databases to hammerdb, postgresql and ripsaw. The CR got extended accordingly and the templates for the db creation and the benchmark runs got separated out.


rsevilla87 (approved):
LGTM!

PR #395, opened 2020-08-28
add bus=virtio to data-volume in server_vm template

netzzer:


jeniferh:
LGTM.
We highly recommend virtio bus over sata for performance reasons, didn't realize until now it was going to default to sata! Annette saw some good improvements with this change. BTW, I don't see a reason to make it configurable, we don't focus on sata much from the virt side.

netzzer:
Creating new PR with rebase to 1 commit

PR #398, opened 2020-09-02
Have the benchmark operator prefer to run on a workload node

dry923:
Currently the benchmark operator will land on any available worker node. This is fine in many cases however to avoid having it being impacted by ongoing tests we should try to pin it to the workload node if possible.


rsevilla87 (approved):
LGTM!

PR #399, opened 2020-09-02
Preliminary integration of kube-burner

rsevilla87:
Preliminary implementation. Works well, however it still needs more testing, documentation, improvements and fix some issues.

As you can see, there are some parameters that are hardcoded since this implementation is meant to support the workloads we currently run. Maybe we can add some parameters in order to make them configurable.

@jtaleric @dry923 feel free to provide some idea or feedback before merging the final bits.


jtaleric (commented):
Nice! Great work @rsevilla87 

Curious if we could discuss the benchmark-operator definition. I wonder if we should statically define the metrics, but allow things for each replica defined be overwritten? 

For example, we have cluster-density - this should have a static set of metrics we define and always collect. however, we should allow users to overwrite `replicas` and count. 

So the configuration would be the qps, es, and for `cluster-density`

```
    imagestream :
      count: 3
    buildconfig:
      count: 3 
... 
```

thoughts?

rsevilla87:
I get your idea @jtaleric, however the main problem with cluster-density is that some resources created by this workload have dependencies with other resources.

i.e
```yaml
- objectTemplate: imagestream.yml 
  replicas: 3                                      
  inputVars:                                      
    prefix: cluster-density-src            

- objectTemplate: imagestream.yml
  replicas: 3                                  
  inputVars:                                  
    prefix: cluster-density-dst        

- objectTemplate: buildconfig.yml
  replicas: 3                                  
  inputVars:                                  
    from: cluster-density-src          
    to: cluster-density-dst              
```

The buildconfig above depends on the two previous imagestreams. The same happens with deployments, services and routes.
Meaning that these objects must be created together. 

We could overcome this with the definition below:

```yaml
- objectTemplate: imagestream.yml 
  replicas: {{ builconfig_replicas }}                                   
  inputVars:                                      
    prefix: cluster-density-src            

- objectTemplate: imagestream.yml
  replicas: {{ builconfig_replicas }}                                  
  inputVars:                                  
    prefix: cluster-density-dst        

- objectTemplate: buildconfig.yml
  replicas: {{ builconfig_replicas }}                                  
  inputVars:                                  
    from: cluster-density-src          
    to: cluster-density-dst              
```

And for the deployments/services/routes we could do something like:
```yaml
- objectTemplate: deployment.yml 
  replicas: {{ deployment_replicas }}                                   
  inputVars:                                      
    prefix: cluster-density-src            

- objectTemplate: service.yml
  replicas: {{ deployment_replicas }}                                    

- objectTemplate: route.yml
  replicas: {{ deployment_replicas }}                                     
```

In addition, keep in mind that the parameter `job_iterations` defines the number of iterations to perform of each workload so defining something below will result in creating 10 namespaces with 10 deployments,10 services and 10 routes:

```yaml
kind: Benchmark
metadata:
  name: kube-burner-cluster-density
  namespace: my-ripsaw
spec:
  workload:
    args:
      qps: 25
      burst: 25
      job_iterations: 10
      workload: kubelet-density
      deployment_replicas: 10
      builconfig_replicas: 10
...
```

rsevilla87:
@jtaleric @dry923 PTAL
I don't have added any test yet. Not sure if it's worth for now.

jtaleric (commented):
Thanks for fixing the readme! I will pull this in and give it a go!

jtaleric:
Hit a interesting issue 
```
time="2020-09-10 19:00:15" level=info msg="📁 Creating indexer: elastic"
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x158879d]
goroutine 1 [running]:
github.com/rsevilla87/kube-burner/pkg/indexers.(*Elastic).new(0xc00000ecd8, 0xc0001c0b50, 0x7, 0xc000221ca0, 0x1, 0x1, 0xc0001c0b30, 0xb, 0x0, 0x0, ...)
        /root/kube-burner/pkg/indexers/elastic.go:57 +0x25d
github.com/rsevilla87/kube-burner/pkg/indexers.NewIndexer(0xc0001c0b50, 0x7, 0xc000221ca0, 0x1, 0x1, 0xc0001c0b30, 0xb, 0x0, 0x0, 0x0, ...)
        /root/kube-burner/pkg/indexers/factory.go:36 +0x189
github.com/rsevilla87/kube-burner/cmd.steps(0x7ffc59464af7, 0x24, 0xc000304e80)
        /root/kube-burner/cmd/kube-burner.go:192 +0x70a
github.com/rsevilla87/kube-burner/cmd.initCmd.func1(0xc0002c8840, 0xc00036ea00, 0x0, 0xa)
        /root/kube-burner/cmd/kube-burner.go:71 +0x127
github.com/spf13/cobra.(*Command).execute(0xc0002c8840, 0xc00036e8c0, 0xa, 0xa, 0xc0002c8840, 0xc00036e8c0)
        /root/kube-burner/vendor/github.com/spf13/cobra/command.go:846 +0x2aa
github.com/spf13/cobra.(*Command).ExecuteC(0x2849660, 0x0, 0x169f9a0, 0xc000094058)
        /root/kube-burner/vendor/github.com/spf13/cobra/command.go:950 +0x349
github.com/spf13/cobra.(*Command).Execute(...)
        /root/kube-burner/vendor/github.com/spf13/cobra/command.go:887
github.com/rsevilla87/kube-burner/cmd.Execute(...)
        /root/kube-burner/cmd/kube-burner.go:160
main.main()
        /root/kube-burner/main.go:22 +0x2e
```

Which seems to be related to the port i used for the ES server. I used `:80` however, the protocol was `https`... I have always used `80` and not `443`. Anyway, switching to `443` *fixed* the issue. I wonder if we need some error handling here? Since we are taking user input? Similar, I didn't provide `https` with the prom url.

rsevilla87:
@jtaleric added node_selector feature and also modified cluster-density pods to mount a couple of secrets each.

jtaleric:
Needs a rebase, but overall ready to merge.

jtaleric (approved):
LGTM

jtaleric (approved):
LGTM - there are some vars I would like to see changed eventually... such as having to vars that should be set to the same thing `prometheus.server` and `elasticsearch.server` however those were introduced prior to this PR.

roles/kube-burner/templates/kubelet-density-heavy.yml.j2
```diff
@@ -0,0 +1,40 @@
...
+  indexerConfig:
+    enabled: true
+    esServers: [{{ prometheus.server }}]
+    insecureSkipVerify: true
+    defaultIndex: {{ workload_args.default_index|default("ripsaw-kube-burner") }}
+    type: elastic
+{% endif %}
+
+jobs:
+  - name: kubelet-density-heavy
```
chaitanyaenr, line 14: Can we add a way to label the nodes using custom nodeselector and max them out ( 250 pods per node ) taking into account the running pods on them like how we typically run kubelet density tests or is it already supported by tweaking the config?

jtaleric: I think the user should just provide the label. I don't think Ripsaw needs to be applying labels to the nodes... It would need to understand "what" nodes the user wants... Meaning if their workers were not uniform, how would Ripsaw choose the right nodes to label? The user should simply provide the label to use, versus Ripsaw needing to encapsulate the logic of what nodes to label. Thoughts?

docs/kube-burner.md
```diff
@@ -0,0 +1,88 @@
...
+
+kube-burner is a tool that allows a user to perform scalability tests across Kubernetes and OpenShift clusters by creating thousands of objects. Kube-burner is developed in it's own repository at https://github.com/cloud-bulldozer/kube-burner
+This ripsaw integration is meant to run certain workloads useful to measure certain performance KPIs of a cluster.
+
+## Running kube-burner
+
+Given that you followed instructions to deploy operator. Kube-burner needs an additional serviceaccount and clusterrole to run. Available at [kube-burner-role.yml](../resources/kube-burner-role.yml)
+You can modify kube-burner's [cr.yaml](../resources/crds/ripsaw_v1alpha1_kube-burner_cr.yaml) to fit your requirements.
+
+## Supported workloads
```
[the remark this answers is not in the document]
rsevilla87: Added more detail about the created objects by each workload. Configuration is well explained at the ## configuration section of the same file

roles/kube-burner/templates/kubelet-density-heavy.yml.j2
chaitanyaenr: Makes sense, we need to document it for the kubelet-density tests to make sure user adds a label to max out the pod capacity on selected nodes.

rsevilla87: The idea is to do this kind of actions from an e2e script. I haven't started coding it yet though

chaitanyaenr: Right but we also need to make sure it's documented in Ripsaw as well since we are explicitly supporting kubelet density workload IMHO.

resources/crds/ripsaw_v1alpha1_kube-burner_cr.yaml
```diff
@@ -0,0 +1,53 @@
...
+  prometheus:
+    # Elastic search instance with full URL format. https://elastic.apps.org:9200
+    server: <ES_URL>
+    # Prometheus bearer token
+    prom_token: <PROMETHEUS_BEARER_TOKEN>
+    # Prometheus URL with full URL format. https://prometheus-k8s.openshift-monitoring.svc.cluster.local:9091
+    prom_url: <PROMETHEUS_URL>
+  workload:
+    name: kube-burner
+    args:
```
chaitanyaenr, line 20: We might want to add an option to delete/re-use the namespace/objects under test if they already exist. Thoughts?

docs/kube-burner.md
```diff
@@ -0,0 +1,116 @@
...
+
+- **workload**: Type of kube-burner workload. As mentioned before, allowed values are cluster-density, kubelet-density and kubelet-density-heavy
+- **default_index**: Elasticsearch index name. Defaults to __ripsaw-kube-burner__
+- **job_iterations**: How many iterations to execute of the specified kube-burner workload
+- **qps**: Limit object creation queries per second. Defaults to __5__
+- **burst**: Maximum burst for throttle. Defaults to __10__
+- **image**: Allows to use an alternative kube-burner container image. Defaults to `quay.io/cloud-bulldozer/kube-burner:latest`
+- **wait_when_finished**: Makes kube-burner to wait for all objects created to be ready/completed before index metrics and finishing the job. Defaults to __true__
+- **pod_wait**: Wait for all pods to be running before moving forward to the next job iteration. Defaults to __false__
+- **verify_objects**: Verify object count after running each job. Defaults to __true__
```
chaitanyaenr, line 52: What's the default timeout after which we report the job as failed?

rsevilla87: We don't have timeout at the moment. Shall we add it?
Maybe something using activeDeadlineSeconds could do the trick. What do you think?

```
$ oc explain job.spec.activeDeadlineSeconds
KIND:     Job
VERSION:  batch/v1

FIELD:    activeDeadlineSeconds <integer>

DESCRIPTION:
     Specifies the duration in seconds relative to the startTime that the job
     may be active before the system tries to terminate it; value must be
     positive integer
```

rsevilla87: I finally added the parameter job_timeout. Defaults to 3600

chaitanyaenr: We might want the timeout to be applicable just for the "verify_objects" action run at the end of the job as it might be tricky to calculate the the total time a job might take from start to finish. Thoughts?

chaitanyaenr: Thinking a bit, we might also want to take the status of the objects into account i.e fail if the state of the object is "failed" or "error" since setting a low timeout might mislead us if the objects are still being created. Thoughts?

chaitanyaenr (approved):
LGTM, great work @rsevilla87.

PR #402, opened 2020-09-09
try to remove all ripsaw state from K8S cluster before starting test

bengland2:
including security and CRD.
tested with smallfile CI on AWS EC2 cluster

### DIFF
--- a/tests/common.sh
+++ b/tests/common.sh
@@ -52,6 +52,7 @@ function wait_clean {
   then
     kubectl delete benchmarks -n my-ripsaw --all --ignore-not-found
   fi
+  kubectl delete -f deploy/ --ignore-not-found
   kubectl delete namespace my-ripsaw --ignore-not-found
 }
 


bengland2:
I was deleting the wrong CRD.   This needs to delete the CRD that references the UUID, and it can only be done at the end of a CI test, not at the beginning of the next one.  And in fact it is being done there in cleanup_resources.      Trying again by just deleting deploy/, not CRD.   Still an improvement IMO.

bengland2:
pausing this because I'm concerned that deploy/4* are global, not namespace-local, so deleting in each test case could impact other concurrent CI runs in the same cluster.

dry923:
> pausing this because I'm concerned that deploy/4* are global, not namespace-local, so deleting in each test case could impact other concurrent CI runs in the same cluster.

@bengland2 I took a look at the test.sh and I think you are right. We would need to add in some logic to tag benchmark-operator-kube with a -$UUID as well

bengland2:
closing this for now.

PR #403, opened 2020-09-09
fixed client.yaml to handle bsrange variable

acalhounRH:
fixed client.yaml to handle bsrange variable


acalhounRH:
@bengland2 @dry923 @jtaleric

Issue #413, opened 2020-09-17
Unable to run iperf benchmark


kpouget:
I'm unable to run the `iperf` benchmark, I don't know if it's due to some issues in my CR (I fixed the `workload` part [a few minutes ago](https://github.com/cloud-bulldozer/benchmark-operator/pull/411)) or if it's the code:

```
apiVersion: ripsaw.cloudbulldozer.io/v1alpha1
kind: Benchmark
metadata:
  name: iperf3-benchmark-sdn
  namespace: my-ripsaw
spec:
  workload:
    name: iperf3
    args:
      pairs: 1
      hostnetwork: false
      pin: true
      pin_server: "worker01"
      pin_client: "worker02"
      port: 5201
      transmit_type: time
      transmit_value: 60
      omit_start: 0
      length_buffer: 128K
      window_size: 64k
      ip_tos: 0
      mss: 900
      streams: 1
      extra_options_client: ' '
      extra_options_server: ' '
      #retries: 200
```
`worker01` and `worker02` are node name in my cluster

```
TASK [Start Server(s)] ********************************
fatal: [localhost]: FAILED! => {
    "msg": "The task includes an option with an undefined variable. The error was: 'iperf3' is undefined\n\nThe error appears to be in '/opt/ansible/roles/iperf3/tasks/main.yml': line 30, column 5, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n  - name: Start Server(s)\n    ^ here\n"
}
```

rsevilla87:
Just detected an issue in the current implementation:
https://github.com/cloud-bulldozer/benchmark-operator/blob/master/roles/iperf3/templates/server.yml.j2#L15
This conditional should be:
{% if workload_args.hostnetwork %}

Im gonna push these changes to fix it. For the time being I think you can overcome this issue removing the line `hostnetwork: false` from your CR.

jtaleric:
@kpouget iperf isn't heavily used by the team... we have some limited CI testing for it. Nice catch!

kpouget:
@rsevilla87, indeed, that seems to be the issue, and without `hostnetwork: false` it goes better :) ... but `hostnetwork: true` is part of what I wanted to test :#

is there an easy way to run the operator from the source? I may try to contribute some MPI benchmarks, so eventually I'll have to run from the source

rsevilla87:
[...]

PR #414, opened 2020-09-17
Fix iperf3 hostnetwork and add test to it

rsevilla87:
iperf blows up if we use hostnetwork as mentioned at https://github.com/cloud-bulldozer/benchmark-operator/issues/413. This PR fixes it and also adds a test case for iperf3 w/ hostnetwork: true


dry923 (approved):
LGTM!

jtaleric (approved):
LGTM

PR #418, opened 2020-09-18
add pairs and multus_client to es

mohit-sheth:


roles/uperf/templates/workload.yml.j2
```diff
@@ -88,13 +88,15 @@ spec:
             export h={{item.spec.clusterIP}};
 {% else %}
 {% if workload_args.multus.client is defined %}
+          - export multus_client={{workload_args.multus.client}}
```
dry923, line 91: I think this is missing "" 's around it. Or at least the opening one and remove the opener from the below line.

roles/uperf/templates/workload.yml.j2
```diff
@@ -84,17 +84,19 @@ spec:
         command: ["/bin/sh", "-c"]
         args:
 {% if workload_args.serviceip is sameas true %}
-          - "export serviceip=true;
```
jtaleric, line 87: Why did this change for ServiceIP?

roles/uperf/templates/workload.yml.j2
```diff
@@ -84,17 +84,19 @@ spec:
...
+          - "export multus_client={{workload_args.multus.client}};
+          - export h={{ (item['metadata']['annotations']['k8s.v1.cni.cncf.io/networks-status'] | from_json)[1]['ips'][0] }};
 {% else %}
           - "export h={{item.status.podIP}};
 {% endif %}
 {% endif %}
              export hostnet={{workload_args.hostnetwork}};
+             export serviceip={{workload_args.serviceip}};
              export ips=$(hostname -I);
+             export pairs={{workload_args.pair}};
```
jtaleric, line 99: s/pairs/num_pairs

roles/uperf/templates/workload.yml.j2
mohit-sheth: ^ this way it will only export serviceip=true. have added it outside the if condition below so that it catches the serviceip=false too

roles/uperf/templates/workload.yml.j2
mohit-sheth: done

dry923 (approved):
LGTM!

roles/uperf/templates/workload.yml.j2
jtaleric: ack I would prefer we break this out of the current PR and put it into a different PR...

mohit-sheth: done

jtaleric (approved):

PR #426, opened 2020-10-01
Allow passing workload.args.prom_es_user and prom_es_pass

dry923:
If login is required for the ES server when pushing prometheus data. It can be supplied in the workload.args section of the yaml as prom_es_user and prom_es_pass.


rsevilla87 (approved):
LGTM!

rsevilla87:
@jtaleric PTAL

jtaleric (approved):
LGTM!

Issue #417, opened 2020-09-17
Feature requests: fio json output retrieval, fio librados and librbd


fritchie:
Feature Requests

1. The fio-client pod goes to a "Completed" status when the fio job is finished. This makes it difficult to
copy the fio json files from /tmp when elasticsearch is not available. An option to keep the fio-client pod running after the job is complete would be nice to have for this situation. Another option would be to have all the json output sent to the pod log as is done for the prefill job.

2. It would be nice to be able to test ceph performance using the librados and librbd fio io engines. Is this something you would be willing to integrate? The fio-server pods would need a valid ceph keyring and ceph.conf file.

jtaleric:
@fritchie Agreed - having a better option for results versus always relying on Elasticsearch. 

@acalhounRH I wonder if we could send result data to our redis backend, and have Ripsaw present the user with results after the run? Since we know the workload has completed?

bengland2:
+1.   However, if you send result to redis, how do results get cleaned up?   If JSON is in the log, it's easy to write a python script that digs it out of the log, and then you can do whatever you want with it.  However, it would be so big that it would make the log harder to read.  Example: the prefill JSON output.

fritchie:
@bengland2 for cleaning up redis I believe it is possible to set an expire time on a key (I have never personally used this though):

https://redis.io/commands/expire

PR #427, opened 2020-10-02
Option to send fio results to client pod log

fritchie:
This commit adds the ability to send fio results to the client pod
logs. This was bought up in

https://github.com/cloud-bulldozer/benchmark-operator/issues/417

If workload_args include

fio_json_to_log: true

all the fio-result.json files found in /tmp/fiod-$uuid will
be sent to the log.

### DIFF
--- a/docs/fio_distributed.md
+++ b/docs/fio_distributed.md
@@ -207,6 +207,7 @@ The workload loops are nested as such from the CR options:
   > Technical Note: If you are running kube/openshift on VMs make sure the diskimage or volume is preallocated.
 - **prefill**: (Optional) boolean to enable/disable prefill SDS
   - prefill requirement stems from Ceph RBD thin-provisioning - just creating the RBD volume doesn't mean that there is space allocated to read and write out there. For example, reads to an uninitialized volume don't even talk to the Ceph OSDs, they just return immediately with zeroes in the client.
+- **fio_json_to_log**: (Optional) boolean to enable/disable sending job results in json format to client pod log.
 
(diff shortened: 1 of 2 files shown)


jtaleric:
@bengland2 @acalhounRH Can I get one of you two +1 this....

jtaleric (approved):
LGTM

rsevilla87 (changes requested):
Hey @fritchie, mind you add this new parameter to the docs?

acalhounRH:
Can we add a logging statement, something like FIO JSON OUTPUT for fiod-{{ uuid }}. that way we can search for the text and make finding the exact sample(s) faster.

acalhounRH:
Okay, so the current way that this is printing out the json files they will be strung together and IMHO will be too messy to make sense of. I would recommend moving this piece of code into the the for-loop on line 74, that way the output is next to benchmark-wrappers execution and there is a clear delineation between samples.

I would not merge as is due to this.

fritchie:
> Can we add a logging statement, something like FIO JSON OUTPUT for fiod-{{ uuid }}. that way we can search for the text and make finding the exact sample(s) faster.

I just used START and END with the filename:

echo START...$jsonfile;cat $jsonfile;echo END...$jsonfile

but I can change to something more pronounced

fritchie:
Requested changes have been made.

rsevilla87 (approved):
LGTM!

jtaleric (approved):
LGTM

acalhounRH (approved):
LGTM

PR #430, opened 2020-10-06
Update test.sh to allow a http address

dry923:
test.sh was using a sed command with / as the delimiter this would cause an error if you passed a http address


rsevilla87:
This change is made as part of https://github.com/cloud-bulldozer/benchmark-operator/pull/438

rsevilla87:
I think we can close this one

PR #431, opened 2020-10-08
Fix uperf in VMs

rsevilla87:
Fix uperf test in virtualized environments. Use uperf from the official Fedora repositories instead the pbench ones.
Fixes: https://github.com/cloud-bulldozer/benchmark-operator/pull/431

### DIFF
--- a/roles/uperf/templates/server_vm.yml.j2
+++ b/roles/uperf/templates/server_vm.yml.j2
@@ -55,12 +55,10 @@ spec:
         chpasswd: { expire: False }
         runcmd:
 {% if workload_args.server_vm.network.multiqueue.enabled %}
-          - yum install -y ethtool
+          - dnf install -y ethtool
           - ethtool -L eth0 combined {{ workload_args.server_vm.network.multiqueue.queues }}
 {% endif %}
-          - yum install -y yum-plugin-copr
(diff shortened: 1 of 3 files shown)


roles/uperf/templates/server_vm.yml.j2
```diff
@@ -55,12 +55,10 @@ spec:
...
         runcmd:
 {% if workload_args.server_vm.network.multiqueue.enabled %}
-          - yum install -y ethtool
+          - dnf install -y ethtool
           - ethtool -L eth0 combined {{ workload_args.server_vm.network.multiqueue.queues }}
 {% endif %}
-          - yum install -y yum-plugin-copr
-          - yum copr enable ndokos/pbench -y
-          - yum install -y pbench-uperf redis git
+          - dnf install -y uperf redis
```
jtaleric, line 61: Do we need to enable the appstream?

rsevilla87: No, it's in the default repositories. (tested it locally ;) )

jtaleric: nice!

jtaleric (approved):
LGTM!

PR #435, opened 2020-10-13
adding makefile and updating contribute docs to recommend makefile

whitleykeith:
Changes:
* Adds Makefile at the root of the project that assists in building the operator image and deploying it to a local k8s cluster
* Updates contribution docs to show usage of Makefile

### DIFF
--- a/CONTRIBUTE.md
+++ b/CONTRIBUTE.md
@@ -130,38 +130,52 @@ case of failure or when disabled. This ensures no interference with subsequent w
 ### The operator container image
 Any changes to the [roles](roles/) tree or to the [playbook](playbook.yml) file will necessitate a new build of the operator container image.
 The container is built using the [Operator SDK](https://github.com/operator-framework/operator-sdk) and pushed to a public repository.
-The public repository could be [quay](https://quay.io) in which case you'll need to:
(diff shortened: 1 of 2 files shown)


CONTRIBUTE.md
```diff
@@ -130,38 +130,56 @@ case of failure or when disabled. This ensures no interference with subsequent w
...
-```bash
-$ operator-sdk build quay.io/<username>/benchmark-operator:testing --image-builder podman 
-$ podman push quay.io/<username>/benchmark-operator:testing
-```
+> Note: Make sure your quay repository has been created before running the below commands, otherwise they'll fail :D
+
+There is a Makefile at the root of the project that will assist in bootstraping your local Kubernetes cluster with the benchmark-operator. 
 
-Note: you can also use docker, and no, we'll not judge ;)
+Running `make` at the root of the project will build the image as `quay.io/$USER/benchmark-operator:$GIT_COMMIT`. However, it does not push by default. In order to push, you can set `PUSH_IMAGE=true` before running `make`.
```
dry923, line 139: Can we default pushing to true?

whitleykeith: done

whitleykeith:
+1 what do we want to do about this?

PR #436, opened 2020-10-14
Rename the prometheus ES server|port|parallel to es_server|port|parallel

dry923:
Per our previous discussions this PR renames the prometheus.server|port|parallel to add es_server|port|parallel. This aims to reduce some confusion from the end user.

NOTE: this does not change the environment variable passed to benchmark-wrapper


jtaleric (approved):
LGTM! Thanks!

rsevilla87 (approved):
LGTM

PR #442, opened 2020-10-20
Fix global overrides

rsevilla87:
Global_overrides was not present in the Benchmark CRD, so the parameters from this option vanish when creating the benchmark.

On the other hand, this is a quick fix for this. As a long term fix we should nest job_params, global_overrides among others under workload.


jtaleric (approved):
LGTM!

dry923 (approved):
LGTM!

PR #453, opened 2020-10-22
adding fio nodeselector and tolerations with docs

whitleykeith:
Changes:

* Fio workloads will be able to add generic node selectors as well as generic tolerations
* The `pin_server` arg will be removed, but the feature is still available by using the `nodeselector` with the `kubernetes.io/hostname` label

### DIFF
--- a/docs/fio_distributed.md
+++ b/docs/fio_distributed.md
@@ -32,7 +32,6 @@ spec:
       prefill: true
       samples: 3
       servers: 3
-      pin_server: ''
       # Chose to run servers in 'pod' or 'vm'
       # 'vm' needs kubevirt to be available
       # Default: pod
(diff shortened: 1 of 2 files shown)


rsevilla87:
@whitleykeith I wonder if we should get rid of pin_server as we can achieve the same results with the your nodeSelector implementation.
@bengland2 @acalhounRH what do you think?

jtaleric (commented):
LGTM, but I agree with @rsevilla87 we should remove pin_server so it is obvious what method to use, and since `nodeselector` can achieve the same thing, no need to keep both. 

The *ONLY* reason we would keep it is if other teams use this feature... which I am not sure if the OCS team uses pin_server.

rsevilla87:
Hey @whitleykeith, mind you remove references to pin_server from the docs?

whitleykeith:
> Hey @whitleykeith, mind you remove references to pin_server from the docs?

done

rsevilla87 (approved):
LGTM!

PR #454, opened 2020-10-22
Typo in uperf tasks for VM run

sarahbx:
During uperf run using OpenShift Virtualization, typo found which prevents changing the client state to "Clients Running" in the operator.


rsevilla87 (approved):
LGTM!

PR #456, opened 2020-10-22
Adding hexfusion's recommendations

### LINKED ISSUE #455: Add hexfusion's go to queries to our metrics.yaml
Sam recommends we should add

```
sum(rate(container_fs_writes_total[5m])) by (device, container, node) != 0
```

and

```
sum(rate(etcd_object_counts{}[5m])) by (resource) != 0
```

And add these to our ectd panels.

jtaleric:
Adding the etcd guru's goto prom queries to our collection

Fixes #455

### DIFF
--- a/roles/kube-burner/files/metrics-aggregated.yaml
+++ b/roles/kube-burner/files/metrics-aggregated.yaml
@@ -163,3 +163,9 @@ metrics:
 
   - query: sum(kube_node_status_condition{status="true"}) by (condition)
     metricName: nodeStatus
+
+  - query: (sum(rate(container_fs_writes_total{device!~".+dm.+"}[5m])) by (device, container, node) and on (node) kube_node_role{role="master"}) > 0
+    metricName: containerDiskUsage
+
+  - query: sum(rate(etcd_object_counts{}[5m])) by (resource) > 0
+    metricName: etcdObjectCount
--- a/roles/kube-burner/files/metrics.yaml
+++ b/roles/kube-burner/files/metrics.yaml
@@ -88,3 +88,10 @@ metrics:
 
   - query: sum(kube_node_status_condition{status="true"}) by (condition)
     metricName: nodeStatus
+
+  - query: (sum(rate(container_fs_writes_total{device!~".+dm.+"}[5m])) by (device, container, node) and on (node) kube_node_role{role="master"}) > 0
+    metricName: containerDiskUsage
+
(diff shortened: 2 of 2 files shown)


roles/kube-burner/files/metrics-aggregated.yaml
```diff
@@ -163,3 +163,10 @@ metrics:
 
   - query: sum(kube_node_status_condition{status="true"}) by (condition)
     metricName: nodeStatus
+
+  - query: sum(rate(container_fs_writes_total[5m])) by (device, container, node)
```
rsevilla87, line 167: This query returns a bunch of timeseries. Roughtly number of containers * number of devices * number of nodes
This is a capture from a 75 node cluster.

>image

As the goal of this query is to measure the disk usage made by etcd we can use this query instead:
>`sum(rate(container_fs_writes_total{container="etcd", device!~".+dm.+"}[5m])) by (device, container, node) > 0`
image

rsevilla87 (changes requested):

roles/kube-burner/files/metrics-aggregated.yaml
```diff
@@ -163,3 +163,10 @@ metrics:
...
+    metricName: containerDiskUsage
+
+  - query: sum(rate(etcd_object_counts{}[5m])) by (resource)
```
rsevilla87, line 170: Adding > 0 at the end will reduce the number of timeseries:
Before:
image

After:
image

rsevilla87 (changes requested):

roles/kube-burner/files/metrics-aggregated.yaml
jtaleric: This isn't what we are looking for though...

We could limit the query to just the master, but just etcd isn't helpful. This is to look for other containers who are competing for the disk.

roles/kube-burner/files/metrics-aggregated.yaml
jtaleric: ack, i'll add that.

roles/kube-burner/files/metrics-aggregated.yaml
rsevilla87:
```
(sum(rate(container_fs_writes_total{device!~".+dm.+"}[5m])) by (device, container, node) and on (node) kube_node_role{role="master"}) >  0

rsevilla87 (approved):
LGTM!

PR #458, opened 2020-10-27
Add ConfigMaps to cluster-density kube-burner tests

smalleni:
It is common in production deployments for pods in deployments to use
ConfigMaps. We should try to replicate that.


smalleni:
@rsevilla87 PTAL

jtaleric (approved):
LGTM

rsevilla87 (approved):
LGTM!

PR #459, opened 2020-10-28
Update node selector in stressng job spec

martineg:
Intendation for the nodeselector field in the stressng job template contained a hard tab, have updated to soft spaces like the rest of the template so the node selector is in the right level within the pod spec. Also converting the field to JSON to allow passing in a map of node selector key/value pairs similar to the vegeta workload.


rsevilla87 (approved):
Thanks for the contribution @martineg  LGTM!

jtaleric (approved):
LGTM!

PR #463, opened 2020-10-30
Use default stockpile tags [k8s,openshift,common]

rsevilla87:
Default tags to prevent executing useless stockpile tasks for containerized environments.


templates/metadata.yml.j2
```diff
@@ -3,7 +3,7 @@
       serviceAccountName: {{ metadata.serviceaccount }}
 {% endif %}
       initContainers:
-      - name: backpack-{{ trunc_uuid }}
```
jtaleric, line 6: so we don't need this to be unique anymore?

PR #468, opened 2020-11-10
Enable simple NetworkPolicy

smalleni:

### DIFF
--- a/playbook.yml
+++ b/playbook.yml
@@ -82,6 +82,9 @@
 
     - block:
 
+      - include_role:
+          name: "common"
+
       - include_role: 
           name: "{{ workload.name }}"
         vars:
--- a/roles/common/tasks/main.yml
+++ b/roles/common/tasks/main.yml
@@ -1,0 +1,17 @@
+# common tasks across multiple roles go here
+- block:
+  - name: Get Network Policy
+    k8s_facts:
+      kind: NetworkPolicy
+      api_version: networking.k8s.io/v1
+      namespace: '{{ operator_namespace }}'
+      name: "{{ meta.name }}-networkpolicy-{{ trunc_uuid }}"
+    register: network_policy
+
+  - name: Create Network policy if enabled
+    k8s:
+      definition: "{{ lookup('template', 'networkpolicy.yml.j2') | from_yaml }}"
+    when: network_policy.resources | length < 1
+
+  when: workload.args.networkpolicy is defined and workload.args.networkpolicy
+
--- a/roles/common/templates/networkpolicy.yml.j2
+++ b/roles/common/templates/networkpolicy.yml.j2
@@ -1,0 +1,14 @@
+kind: NetworkPolicy
+apiVersion: networking.k8s.io/v1
+metadata:
+  name: "{{ meta.name }}-networkpolicy-{{ trunc_uuid }}"
+  namespace: '{{ operator_namespace }}'
+spec:
+  podSelector: 
+    matchLabels:
+      type: "{{ meta.name }}-bench-server-{{ trunc_uuid }}"
+  ingress:
+  - from:
+    - podSelector: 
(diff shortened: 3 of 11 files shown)


rsevilla87 (changes requested):
Nice one @smalleni!, can we add a test case for this?

roles/common/templates/networkpolicy.yml.j2
```diff
@@ -0,0 +1,14 @@
+kind: NetworkPolicy
```
jtaleric, line 1: should we name this uperf-networkpolicy, or better yet, should we template the workload name?

smalleni: That is the kind which is `NetworkPolicy`..the name as you can see below is 

"{{ meta.name }}-networkpolicy-{{ trunc_uuid }}"

Am I missing something?

roles/common/tasks/main.yml
```diff
@@ -0,0 +1,7 @@
+# common tasks across multiple roles go here
+- name: Create Network policy if enabled
+  k8s:
+   definition: "{{ lookup('template', 'networkpolicy.yml.j2') | from_yaml }}"
+  vars:
+    trunc_uuid: "{{ cr_state.resources[0].status.suuid }}"
+  when: workload.args.networkpolicy is defined and workload.args.networkpolicy
```
dry923, line 7: As it is written the template would be called every time the reconciliation loop triggered. While it's not really going to break anything we should probably put in a check to see if it exists and then execute if it does not.

something like:
```
block:
  k8s_facts:
      kind: NetworkPolicy
      api_version: networking.k8s.io/v1
      name: "{{ meta.name }}-networkpolicy-{{ trunc_uuid }}"
    register: net_policy
    
  - name: Create Network policy if enabled
    k8s:
     definition: "{{ lookup('template', 'networkpolicy.yml.j2') | from_yaml }}"
    vars:
      trunc_uuid: "{{ cr_state.resources[0].status.suuid }}"
  when: net_policy | json_query(some_valid_query

 when: workload.args.networkpolicy is defined and workload.args.networkpolicy
```

Thoughts?

roles/common/templates/networkpolicy.yml.j2
```diff
@@ -0,0 +1,14 @@
...
+apiVersion: networking.k8s.io/v1
+metadata:
+  name: "{{ meta.name }}-networkpolicy-{{ trunc_uuid }}"
+  namespace: '{{ operator_namespace }}'
+spec:
+  podSelector: 
+    matchLabels:
+      type: uperf-bench-server-{{ trunc_uuid }}
```
jtaleric, line 9: The reason I mentioned my comment is because of the "uperf"- here.

playbook.yml
```diff
@@ -56,6 +56,9 @@
     
     when: workload is defined and (cr_state.resources[0].status is not defined or cr_state.resources[0].status.uuid is not defined)
 
+  - include_role:
+      name: "common"
```
dry923, line 60: Should this be in the lower Run workload block?

roles/common/tasks/main.yml
smalleni: Yeh it didn't seem like there was a state that i could take advantage of like Running/Created. I''l check again.

roles/common/templates/networkpolicy.yml.j2
smalleni: done

roles/common/tasks/main.yml
smalleni: Done

rsevilla87 (approved):
LGTM!

dry923 (approved):
LGTM!

jtaleric (approved):
LGTM

PR #472, opened 2020-11-13
Fix uperf hostNetwork and add test case for it

rsevilla87:
The title says it all.
The benchmark-operator and backpack (this one is used in case of privileged metadata collection) SAs are already bound to the privileged SCC, so no further actions are required.

### DIFF
--- a/docs/uperf.md
+++ b/docs/uperf.md
@@ -66,14 +66,6 @@ spec:
 
 `networkpolicy` will create a simple networkpolicy for ingress
 
-*Note:* If you want to run with hostnetwork on `OpenShift`, you will need to execute the following:
-
-```bash
-
-$ oc adm policy add-scc-to-user privileged -z benchmark-operator
-
-```
-
 `pin` will allow the benchmark runner place nodes on specific nodes, using the `hostname` label.
 
 `pin_server` what node to pin the server pod to.
--- a/roles/uperf/templates/server.yml.j2
+++ b/roles/uperf/templates/server.yml.j2
@@ -23,7 +23,6 @@ spec:
 {% if workload_args.hostnetwork is sameas true %}
       hostNetwork: true
       serviceAccountName: benchmark-operator
-      serviceAccount: benchmark-operator
 {% endif %}
       containers:
       - name: benchmark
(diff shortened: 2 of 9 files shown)


jtaleric (approved):
LGTM

docs/uperf.md
```diff
@@ -66,14 +66,6 @@ spec:
 
 `networkpolicy` will create a simple networkpolicy for ingress
 
-*Note:* If you want to run with hostnetwork on `OpenShift`, you will need to execute the following:
```
smalleni, line 69: Nice, yeh removing any manual steps is a big +1.

roles/uperf/templates/server.yml.j2
```diff
@@ -22,6 +22,8 @@ spec:
 {% endif %}
 {% if workload_args.hostnetwork is sameas true %}
       hostNetwork: true
+      serviceAccountName: benchmark-operator
+      serviceAccount: benchmark-operator
```
smalleni, line 26: Quick question, do we need both `serviceAccountName` and `serviceAccount` ?

tests/test_crs/valid_uperf_resources.yaml
```diff
@@ -22,7 +22,7 @@ spec:
         requests:
           cpu: 100m
           memory: 100Mi
-      hostnetwork: false
+      hostnetwork: true
```
smalleni, line 25: So idea is to reuse this test for hostnetwork instead of net new test that adds time?

rsevilla87: yes, I think it's not worth to add a new test for this so I took advantage of this one

roles/uperf/templates/server.yml.j2
rsevilla87: nope, in fact serviceAccount is deprecated...
```
[root@ip-172-31-78-162 ~]# oc explain pod.spec.serviceAccount
KIND:     Pod
VERSION:  v1

FIELD:    serviceAccount <string>

DESCRIPTION:
     DeprecatedServiceAccount is a depreciated alias for ServiceAccountName.
     Deprecated: Use serviceAccountName instead.
```

Fixing it...

tests/test_crs/valid_uperf_resources.yaml
smalleni: Makes sense to me.

smalleni (approved):

PR #478, opened 2020-11-17
Add max-services and max-namespaces workloads

rsevilla87:
Also adding instant Prometheus queries for kube_node_role and update kubelet metrics as topk doesn't get along well with rate.

Update: pin_server now accepts generic nodeSelector


docs/kube-burner.md
```diff
@@ -36,6 +36,15 @@ Each iteration of this workload creates the following objects:
   - 1 deployment holding a client application for the previous database
   - 1 service pointing to the postgresl database
```
smalleni, line 38: space between as and many :)

docs/kube-burner.md
```diff
@@ -36,6 +36,15 @@ Each iteration of this workload creates the following objects:
...
 
+- **max-namespaces**: This workload is a cluster limits focused test which creates maximum possible namespaces across the cluster. This is a namespaced workload, meaning that kube-burner **will createas many namespaces with these objects as the configured job_iterations**. Each namespace will be populated with the following objects.:
+  - 1 deployment holding a postgresql database
+  - 1 deployment holding a client application for the previous database
+  - 1 service pointing to the postgresl database
+
+- **max-services**: This workload is a cluster limits focused test which creates maximum possible services per each namespace. Each namespace will be populated with the following objects.:
```
smalleni, line 44: Maybe add that job_iterations defines the number of namespaces?

rsevilla87: no, this workload only creates one namespace, I've made some modifications in docs to clarify this.

smalleni: Ack, thanks!

docs/kube-burner.md
```diff
@@ -36,6 +36,15 @@ Each iteration of this workload creates the following objects:
   - 1 deployment holding a client application for the previous database
   - 1 service pointing to the postgresl database
 
+- **max-namespaces**: This workload is a cluster limits focused test which creates maximum possible namespaces across the cluster. This is a namespaced workload, meaning that kube-burner **will create as many namespaces with these objects as the configured job_iterations**.
+  - 1 deployment holding a postgresql database
```
chaitanyaenr, line 40: We might want to make the max-namespaces test a little bit more heavy. We usually have 10k namespaces with ~60k pods using 10k deployments, ~190k secrets, image streams, environment variables and services - https://github.com/openshift/svt/blob/master/openshift_scalability/config/golang/cluster-limits-namespaces.yaml. Thoughts?

rsevilla87: makes sense, I didn't realize that deployment had 6 replicas

docs/kube-burner.md
```diff
@@ -36,6 +36,15 @@ Each iteration of this workload creates the following objects:
...
+  - 5 deployments consisting of a client application for the previous database
+  - 1 service pointing to the postgresl database
+
+- **max-services**: This workload is a cluster limits focused test which creates maximum possible services per namespace. It **will create a single namespace, each iteration of this workload will populate that namespace with these objects:**
```
chaitanyaenr, line 44: The number of pods/deployments that can run with 5000 services in the namespace is limited by the byte size of the service name as there's a limit on the size of the arguments that be passed during exec. Should we expose the service name as a config option to be able to tweak it during the test? Thoughts?

docs/kube-burner.md
rsevilla87: fixed!

docs/kube-burner.md
rsevilla87: As we discussed offline, I removed UUID suffix from resource names. so we shouldn't overcome the byte size limit.

chaitanyaenr (approved):
LGTM, nice work @rsevilla87.

Issue #473, opened 2020-11-13
[Uperf] Allow different request and response sizes


smalleni:
Currently we support 
```
- sizes
```

as a list of message sizes in the CR for uperf.

This translates into the same size for request and response in the uperf xml. We need a way to specify a response size that is differetn from the request size to do things like request-response tests with 64B requests and 1K responses for example (typical HTTP request response sizes, with smaller requests than responses).

bengland2:
definitely want this for Ceph, since reads and writes are a perfect example of asymmetric message sizes.

dustinblack:
Done with PR #481

PR #481, opened 2020-11-20
add functionality for uperf rr asymmetric write/read packet sizes

dustinblack:
Addresses issue #473 

Without requiring any changes to the CR YAML spec beyond a documentation update, a user can now pass values to `sizes` in a sub-list format. For request-response (rr) tests, the first sub-list item will be treated as the write size value, and the second sub-list item will be treated as the read value. This does not supersede the ability to pass a single value for symmetric write/read sizes, and it is possible to mix data types in the parent `sizes` list.

```yaml
      samples: 3
      pair: 1
      test_types:
        - rr
      protos:
        - tcp
      sizes:
        - 1024
        - [8192, 4096]
      nthrs:
        - 1
      runtime: 30
```


rsevilla87 (approved):
LGTM!

dry923 (approved):
LGTM!

PR #491, opened 2020-12-02
uperf with sctp

mkarg75:
Added updated documentation as well as the required CR files to enable sctp on an openshift cluster.


smalleni (commented):
Just my 0.02. Since this PR is about setting up the env to enable SCTP, I think this is beyond the scope of the benchmark-operator. For example, the cyclictest, the oslat, servicemesh benchmarks: none of them setup the env. 

@rsevilla87 thoughts?

jtaleric (changes requested):
IMHO this sound live in e2e-benchmarking versus in the operator.

rsevilla87:
Closing in favor of a e2e-benchmark implementation

PR #438, opened 2020-10-15
Remove ElasticSearch port variable and update references in documenta…

rsevilla87:
…tion.

With this change, the ElasticSearch URL should be passed through the variable elasticsearch.url and prometheus.es_url.
Further changes will be required in [benchmark-wrapper](https://github.com/cloud-bulldozer/benchmark-wrapper/pull/219), [e2e-benchmarking](https://github.com/cloud-bulldozer/e2e-benchmarking/pull/61)

Signed-off-by: Raul Sevilla <rsevilla@redhat.com>

PR #492, opened 2020-12-02
Added documentation about prometheus collection integration and how t…

acalhounRH:
Added documentation about Prometheus collection integration and how to extend functionality to other workloads

@dry923 
@bengland2 
@ekuric 
@jtaleric 
@rsevilla87

### DIFF
--- a/docs/elastic.md
+++ b/docs/elastic.md
@@ -2,14 +2,52 @@
 
 How To:
 * [What is it](#what-is-it)
-* [Configuring](#configuring)
+* [Enabling Collection](#enabling-collection)
+    * [Benchmark Data](#benchmark-data)
+    * [Prometheus Data](#prometheus-data)
+* [Extending Prometheus Collection](#extending-prometheus-collection)
+    * [Elasticsearch-Prometheus Data model](#elasticsearch-prometheus-data-model)
+    * [Explaining Workload Include Files](#explaining-workload-include-files)
+    * [Adding Prometheus collection trigger](#adding-prometheus-collection-trigger)
+
 
 # What is it
 
-Most workloads are configured to send the benchmark data to Elasticsearch if given the proper information.
+Elasticsearch is a distributed NoSQL database that allows users to quick search and analyze data. After execution of
+each benchmark benchmark-operator(via benchmark-wrapper) collects test data and indexes it into Elasticsearch. Besides 
+initial setup and deployment of the elasticsearch cluster, workloads are configured to send the benchmark data to Elasticsearch
+automatically, when users provide the proper information. 
 
-# Configuring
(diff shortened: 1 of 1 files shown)


bengland2:
@acalhounRH , thank you, this is much better.   One thing I don't understand though - why in the prometheus: section do you have to specify es_server: and es_port: when it is already specified under elasticsearch: above?   Seems like you wouldn't use a different ES server for prom data and test results would you?  At least the default should be the same ES server used in the elasticsearch: section?   Anything that reduces number of input parameters needed for the average user significantly increases the chance that they will succeed at using it.  -dumb ole ben

acalhounRH:
Thank you Ben for the input, lets submit a issues for defaulting the prom ES instance if the es_server is not present.

docs/elastic.md
```diff
@@ -50,3 +88,311 @@ spec:
...
+      threads: 1
+      file_size: 0
+      files: 100000
+```
+   
+#### Retrieving Openshift Prometheus info
+
+In order to query Prometheus we must provide the URL and access token. 
+
+The URL can be captured by executing `oc -n openshift-monitoring get routes | grep prometheus`.
```
rsevilla87, line 148: It's also possible (and recommended), to use the internal prometheus service addres, which is always the same for every cluster and prevents traffic to go out and in the cluster. (https://prometheus-k8s.openshift-monitoring.svc.cluster.local:9091)

docs/elastic.md
```diff
@@ -50,3 +88,311 @@ spec:
...
+
+Command to setup a service account and retrieve token: 
+
+```
+oc create serviceaccount snafu -n my-ripsaw
+oc create clusterrolebinding grafana-cluster-monitoring-view --clusterrole=cluster-monitoring-view --serviceaccount=my-ripsaw:snafu
+oc sa get-token snafu -n my-ripsaw
+```
+
+*NOTE:* Service account tokens can expire, for long running test it is recommend to either set the necessary permissions
```
rsevilla87, line 161: By default service account tokens doesn't expire. Rather user tokens expire. i.e (oc whoami -t)

acalhounRH:
Updated and corrected doc's based on @rsevilla87 comments. 

Ready to merge.

rsevilla87 (approved):
This is great @acalhounRH . LGTM!

rsevilla87:
> @acalhounRH , thank you, this is much better. One thing I don't understand though - why in the prometheus: section do you have to specify es_server: and es_port: when it is already specified under elasticsearch: above? Seems like you wouldn't use a different ES server for prom data and test results would you? At least the default should be the same ES server used in the elasticsearch: section? Anything that reduces number of input parameters needed for the average user significantly increases the chance that they will succeed at using it. -dumb ole ben

FYI: There's an ongoing effort to get rid of ES port variables at https://github.com/cloud-bulldozer/benchmark-operator/pull/438

PR #493, opened 2020-12-07
Fix pinned server in uperf workload vm template. Courtesy of @jeniferh.

sarahbx:
@jeniferh found uperf VM workload had same host pinned as server.
Thank you for finding this! 😃


jeniferh:
LGTM, thanks for submitting.

Issue #496, opened 2020-12-10
Implement operator loop state tasks as a "library" item


dustinblack:
The tasks for operator loop state detection have become boilerplate across the workload roles. I think we should consider moving this into the "common" role and importing the tasks as effectively "library" items into the workload `tasks/main.yml` playbooks.

Ultimately, I think a lot of workload role functionality ends up being reusable, and it would be nice to see other functionality get moved into the "common" role. These loop state tasks should be fairly low effort and straightforward to convert to this "library" approach and can serve as a template for how this can be leveraged for other functionality.

```yaml
- name: Get current state
  k8s_info:
    api_version: ripsaw.cloudbulldozer.io/v1alpha1
    kind: Benchmark
    name: '{{ meta.name }}'
    namespace: '{{ operator_namespace }}'
  register: resource_state

- operator_sdk.util.k8s_status:
    api_version: ripsaw.cloudbulldozer.io/v1alpha1
    kind: Benchmark
    name: "{{ meta.name }}"
    namespace: "{{ operator_namespace }}"
    status:
      state: Building
      complete: false
  when: resource_state.resources[0].status.state is not defined

- name: Get current state - If it has changed
  k8s_info:
    api_version: ripsaw.cloudbulldozer.io/v1alpha1
    kind: Benchmark
    name: '{{ meta.name }}'
    namespace: '{{ operator_namespace }}'
  register: resource_state
```

PR #497, opened 2020-12-11
move enumeration logic to common role for reuse by other roles

dustinblack:
This PR is tangentially related to Issue #496 in that it implements the proposed "library" approach for a set of tasks, moving those tasks to the common role and making them more generically applicable.

The functionality moved to the common role is that which takes a list and makes a second list that is enumerated via a pattern up to the source list length. This is used by the pgbench role to create a pattern of running through databases, such as a "doubling" approach where we will test against 1 database, then 2, then 4, 8, 16, and so on until we reach the the highest number below the total number of databases.

With the development of enhanced functionality of the uperf workload, a similar patterning approach may be needed, so moving these tasks to the common role will make them reusable for that purpose. CC @mrbojangles3 and @HughNhan


roles/common/tasks/enumeration_by_pattern.yml
```diff
@@ -0,0 +1,86 @@
...
+  with_indexed_items: "{{ enumeration_list }}"
+  when: enumeration_pattern == "by2" and item.0|int is odd
+
+- name: Set enumeration to exponential doubling pattern
+  set_fact:
+    enumeration: "{{ enumeration|default([]) + [(item.0|int + base)] }}"
+  with_indexed_items: "{{ enumeration_list }}"
+  when: enumeration_pattern == "doubling" and (((item.0|int + base)|log(2)) is even or ((item.0|int + base)|log(2)) is odd)
+
+- name: Set enumeration to all pattern
```
smalleni, line 83: @dustinblack  If I understand the logic correctly, the `enumeration` list gives you a list on indices in the `enumeration_list` based on the `enumeration_pattern` you want and where you start counting from (base).  How will this work in the case of `enumeration_pattern` of `all`. The enumeration list will have a number greater than the highest index in the list.

if `enumeration_list` is [1,2,3], we end up with `enumeration` = 3 in the case of `base` =0 and `enumeration` = 4,, and both the index 3 and 4 are not present in the list. 

Is my understanding right?

smalleni (changes requested):

PR #498, opened 2020-12-11
Support for uperf multi-pair VM runs

sarahbx:
Tested running up to five uperf VM client-server pairs.
 
* Add a "client ready" state for VMs to allow for any installation and setup prior to test
* Use SETNX/INCR for both "client ready" and "server ready" to ensure we have an even count before triggering start
* When "blocking client from running uperf", use 'start' key instead of altering server 'ready'
* Increase VM memory requirements to reduce memory pressure during startup
* Add a 1/10s sleep in the "client ready" loop to prevent lock-up
* Changes to match pod uperf test
  * VM CPU topology to remain on a single socket
  * Uperf test arguments

### DIFF
--- a/roles/uperf/tasks/main.yml
+++ b/roles/uperf/tasks/main.yml
@@ -256,17 +256,17 @@
       command: "redis-cli get clients-{{ trunc_uuid }}"
       register: clients_ready_count
 
-    when: resource_kind == "vm"
+    - name: Update resource state
+      operator_sdk.util.k8s_status:
+        api_version: ripsaw.cloudbulldozer.io/v1alpha1
+        kind: Benchmark
+        name: "{{ meta.name }}"
+        namespace: "{{ operator_namespace }}"
+        status:
+          state: Clients Running
+      when: "workload_args.pair|default('1')|int == clients_ready_count.stdout|int"
 
-  - name: Update resource state
-    operator_sdk.util.k8s_status:
-      api_version: ripsaw.cloudbulldozer.io/v1alpha1
-      kind: Benchmark
-      name: "{{ meta.name }}"
-      namespace: "{{ operator_namespace }}"
-      status:
-        state: Clients Running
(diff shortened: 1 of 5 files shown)


roles/uperf/tasks/main.yml
```diff
@@ -253,27 +252,22 @@
...
+  - name: Update resource state
+    operator_sdk.util.k8s_status:
+      api_version: ripsaw.cloudbulldozer.io/v1alpha1
+      kind: Benchmark
+      name: "{{ meta.name }}"
+      namespace: "{{ operator_namespace }}"
+      status:
+        state: Clients Running
+    when: "workload_args.pair|default('1')|int == clients_ready_count.stdout|int"
+
```
rsevilla87, line 270: I think this task should be in the previous block as clients_ready_count is only defined when using VMs

rsevilla87 (changes requested):
Thanks for the PR @sarahbx , have added some comments

roles/uperf/tasks/main.yml
sarahbx: Good catch. I'll move it up. It would be good for us to do the same with pods, make sure everything is up before triggering start, but I only wanted to change VMs with this PR :)

rsevilla87: Yeah, makes sense. With pods we don't have to install RPMs so the gap between clients is not that high. But we need to start working in a mechanism to make sure the different samples/test-cases are also synchronized.

sarahbx: For sure. I think we could use what I did here with some slight changes for the rest to ensure that syncronization. I also have a ticket to create a VM image for bo so we don't have to do the installs at runtime, save a bunch of time. :)

rsevilla87 (approved):
LGTM!

jtaleric (approved):
LGTM, spoke with @sarahbx on Slack regarding this PR.

PR #503, opened 2020-12-15
Create Helm Chart for benchmark-operator

whitleykeith:
This PR creates a helm chart for the benchmark-operator, which can be used to provide an easier installation experience compared to the current process.

With this chart, installing the operator only requires the following:

```bash
oc adm policy -n my-ripsaw add-scc-to-user privileged -z benchmark-operator
helm install ripsaw $CHART_PATH -n my-ripsaw --create-namespace
```

deleting the operator is also simple, requiring the following:

```bash
helm delete ripsaw -n my-ripsaw --purge
```

It's also much easier to parameterize things through the use of Helm Values. For example, changing the operator image and tag is done in the same line as the install:

```bash
cd charts/benchmark-operator
helm install ripsaw . -n my-ripsaw --create-namespace --set operator.image.repository=$OPERATOR_REPO,operator.image.tag=$OPERATOR_TAG
```


jtaleric (changes requested):
I am not too familiar with Helm, but this seems to be missing a readme and test for the Helm install... i wonder if we could add a simple test to install via the helm chart?

whitleykeith:
@jtaleric I added a README for the helm chart, but I'm not sure how to handle adding a test with the E2E refactoring pending since it would create a good bit of conflicts

charts/benchmark-operator/templates/operator.yaml
```diff
@@ -0,0 +1,76 @@
...
+        name: {{ include "benchmark-operator.fullname" . }}
+    spec:
+      serviceAccountName: {{ include "benchmark-operator.fullname" . }}
+      containers:
+        - name: ansible
+          command:
+          - /usr/local/bin/ao-logs
+          - /tmp/ansible-operator/runner
+          - stdout
+          image: "{{ .Values.operator.image.repository }}:{{ .Values.operator.image.tag | default "master" }}"
```
rsevilla87, line 23: Why don't you default this variable in the values.yaml file?

charts/benchmark-operator/templates/serivce-account.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
```
rsevilla87, line 1: s/serivce-account.yaml/service-account.yaml

rsevilla87 (changes requested):
Just a couple of small nits

README.md
```diff
@@ -3,6 +3,18 @@
...
+## Installation
+
+Installing the benchmark-operator is easiest by using the helm chart and can be done with the following commands.
+
+> Note: If running on openshift you'll need to run this command before installing the chart. `oc adm policy -n my-ripsaw add-scc-to-user privileged -z benchmark-operator`
+
+```bash
+git clone https://github.com/cloud-bulldozer/benchmark-operator
+cd benchmark-operator/charts/benchmark-operator
+helm install benchmark-operator . -n my-ripsaw --create-namespace
```
chaitanyaenr, line 15: We might want to mention that helm needs to be installed on the box and may be also include the helm command to delete the operator that you have in the commit message.

chaitanyaenr (commented):
Minor nit, otherwise LGTM.

rsevilla87 (approved):
LGTM!, nice work @whitleykeith

chaitanyaenr (approved):
LGTM

PR #507, opened 2021-01-06
Enable ssl verification for Elasticsearch in Backpack

dry923:
Allow ssl verification for elasticsearch in backpack. To enable add ssl: true to the metadata section of the cr.

```
metadata:
  ssl: true
```


jtaleric (approved):
LGTM

PR #509, opened 2021-01-12
Add Iperf3 pod anti-affinity rules

martineg:
Add pod anti-affinity rules to place iperf3 client and server benchmark pods on different nodes


jtaleric:
Do we want this by default? Or allow a switch to disable the affinity?

martineg:
@jtaleric Added this on by default similar to what's already in the uperf workload (#238) - but a switch to disable might make sense.

rsevilla87 (approved):
LGTM

jtaleric:
LGTM!

PR #511, opened 2021-01-15
Add node name to environment variables

smalleni:
This helps in cases where uperf is run at scale (across multiple nodes)
and we want to figure out throughput per worker node. Tis information can
be consumed in the benchmark-wrapper.

### DIFF
--- a/roles/uperf/templates/workload.yml.j2
+++ b/roles/uperf/templates/workload.yml.j2
@@ -67,12 +67,12 @@ spec:
           - name: prom_url
             value: "{{ prometheus.prom_url | default() }}"
 {% endif %}
-{% if uperf.pin is sameas true %}
           - name: client_node
-            client_node: {{ uperf.pin_client }}
-          - name: client_node
-            server_node: {{ uperf.pin_server }}
-{% endif %}
+            valueFrom:
+              fieldRef:
+                fieldPath: spec.nodeName
+          - name: server_node
+            value: "{{ uperf.pin_server|default("unknown") }}"
 {% if workload_args.client_resources is defined %}
         resources: {{ workload_args.client_resources | to_json }}
 {% endif %}


rsevilla87:
> This helps in cases where uperf is run at scale (across multiple nodes)
> and we want to figure out throughput per worker node. Tis information can
> be consumed in the benchmark-wrapper.
> 
> Signed-off-by: Sai Sindhur Malleni [smalleni@redhat.com](mailto:smalleni@redhat.com)

Hey Sai, I like this PR, I have a suggestion to make it greater. 
we might take advantage of the client_node variable that is already there and are only set with using pin currently:
https://github.com/cloud-bulldozer/benchmark-operator/blob/03463b2c668584119eaea7bed2201fe455bbe4b0/roles/uperf/templates/workload.yml.j2#L74-L79
As you can see, this implementation is not correct at all (`client_node:` and `server_node:` must be replaced by `value:`), I think we can improve it with with something like:

```
          - name: client_node
            valueFrom:
              fieldRef:
                fieldPath: spec.nodeName
          - name: server_node
            value: "{{ uperf.pin_server|default("unknown") }}"
```

With this implementation, no further changes would be needed in benchmark-wrapper. What do you think?

smalleni:
@rsevilla87 Right, yeh I was clueless about 
``` {% if uperf.pin is sameas true %} 
           - name: client_node 
             client_node: {{ uperf.pin_client }} 
           - name: client_node 
             server_node: {{ uperf.pin_server }} 
 {% endif %} 
```

Definitely didn't look like environment variables :) I like your suggestion, I'll do that.

rsevilla87:
Some logs from CI:
image

rsevilla87 (approved):
LGTM!

jtaleric (approved):
LGTM

PR #517, opened 2021-01-26
Update kubelet-density name to node-density.

rsevilla87:
Will prevent issues like the below in managed services deployed clusters.
```
time="2021-01-26 05:06:06" level=error msg="Unexpected error creating namespace: admission webhook \"namespace-validation.managed.openshift.io\" denied the request: Prevented from accessing Red Hat managed namespaces. Customer workloads should be placed in customer namespaces, and should not match this regular expression: (^kube.*|^openshift.*|^default$|^redhat.*)"
```


dry923 (approved):
LGTM!

chaitanyaenr (approved):
LGTM

PR #525, opened 2021-02-08
Add UPERF Scale Mode Telco-5G

HughNhan:
### Summary of Changes:
This PR adds scale mode enhancement to UPERF workload for Telco-5G such as VZ's Webscale where there are up to 250 nodes at the moment.  The scale mode are realized by the following new UPERF CR options

- colocate: true/false
- density_range: [low,high]
- node_range:[ low,high]
- step_size: addN, log2
- exclude_labels: []

See uperf.md for more details.  The following preso is the project status sandbox but you could find a few useful explanations,
https://docs.google.com/presentation/d/11zCqw8K98jams8wYO24ATwpKzNYTK-xuC-gbvAddbLY/edit#slide=id.g78a1e8b739_0_913 

### Limitations:
Only support "pod' type.  "VM" type functionality remains unchanged. Future enhancement is being considered

### Scope of Changes:

1. Add ability to gather node inventory at run-time to build-up worker node list.
2. Add ability to instantiate the number of server pods according to `node_range` and `density_range`,
3. Add ability to instantiate client pod with correct affinity based on `colocate` value.
4. Add additional state to uperf workload's main state machine so that it can iterate over `node_range` and `density_range`
5. Add additional states to UPERF client workload.yml.j2 to that it can run again and again under the control of the main state machine.
6. Change serviceip, server, and client spec to list to speedup instantiation.
7. While at it, restructure the UPERF workload main.yaml to help with adding new states, isolate pod type from VM type specific code.
8. Enhance 'pin' mode to use density_range[], so pin mode now can iterate
9. remove 'pair' - now we will use 'density_range'

### Files:

- main.yml  - this was one monolithic UPERF file. Now it is just the "state machine"/main loop
- setup.yml  - contains initial setup of dynamic node list, indices representing node_range[] and density_range[]
- init.yml      - init code used mostly for creating server pods
- start_server.yml - create server pods  
- wait_server_ready.yml - wait for server pods become ready
- start_client.yml  - create client pods  
- wait_client_ready.yml  - wait for client pod becomes ready

The nest 3 files are responsible for running the loop to iterate node_range[] and density_range[]

- run_a_set.yml 
- wait_set_done.yml
- next_set.yml  

send_client_run_signal.yml  - for VM specific

The next 2 are common code to cleanup after the run>

- wait_client_done.yml  
- cleanup.yml 

### Unit-test

- PASS - test colocate, density_range, node_range, step_size
- PASS-  test pin mode
- PASS - test serviceip
- PASS - test VM mode for no regression - Need to use valid images and CR

### DIFF
--- a/deploy/25_role.yaml
+++ b/deploy/25_role.yaml
@@ -1,13 +1,14 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  creationTimestamp: null
   name: benchmark-operator
 rules:
 - apiGroups:
-  - "*"
+  - ""
   resources:
-  - "*"
+  - nodes
   verbs:
-  - '*'
+  - get
+  - list
+  - patch
 
--- a/deploy/25_role.yaml
+++ b/deploy/25_role.yaml
@@ -1,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  creationTimestamp: null
+  name: benchmark-operator
+rules:
+- apiGroups:
+  - "*"
+  resources:
+  - "*"
+  verbs:
+  - '*'
+
--- a/docs/uperf.md
+++ b/docs/uperf.md
@@ -44,7 +44,6 @@ spec:
       multus:
         enabled: false
       samples: 1
-      pair: 1
       test_types:
         - stream
       protos:
@@ -54,6 +53,10 @@ spec:
       nthrs:
         - 1
       runtime: 30
+      colocate: false
+      density_range: [low, high]
+      node_range: [low, high]
+      step_size: addN, log2
 ```
 
 `client_resources` and `server_resources` will create uperf client's and server's containers with the given k8s compute resources respectively [k8s resources](https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/)
@@ -82,7 +85,7 @@ spec:
 
 ```yaml
       samples: 3
-      pair: 1
+      density: [1,1]
       test_types:
         - stream
       protos:
@@ -108,7 +111,7 @@ size.
 For example:
 ```yaml
       samples: 3
-      pair: 1
+      density: [1,1]
       test_types:
         - rr
       protos:
@@ -189,6 +192,55 @@ To enable Multus in Ripsaw, here is the relevant config.
       ...
 
 ```
+### Scale
+Scale in this context refers to the ability to enumerate UPERF 
+client-server pairs during test in a control fashion using the following knobs.
+
+`colocate: true` will place each client and server pod pair on the same node.
+
+`density_range` to specify the range of client-server pairs that the test will iterate.
+
+`node_range` to specify the range of nodes that the test will iterate.
+
+`step_size` to specify the incrementing method.
+
+Here is one scale example:
+
+```
+      ...
+      pin: false
+      colocate: false
+      density_range: [1,10]
+      node_range: [1,128]
+      step_size: log2
+      ...
+```
+Note, the `scale` mode is mutual exlusive to `pin` mode with the `pin` mode has higher precedence.
+In other words, if `pin:true` the test will deploy pods on `pin_server` and `pin_client` nodes
+and ignore `colocate`, `node_range`, and the number of pairs to deploy is specified by the
+ `density_range.high` value.
+
+In the above sample, the `scale` mode will be activated since `pin: false`. In the first phase, the 
+pod instantion phase, the system gathers node inventory and may reduce the `node_range.high` value 
+to match the number of worker node available in the cluster.
+
+According to `node_range: [1,128]`, and `density_range:[1,10]`, the system will instantiate 10 pairs on 
+each of 128 nodes. Each pair has a node_idx and a pod_idx that are used later to control
+which one and when they should run the UPERF workload, After all pairs are up and ready,
+next comes the test execution phase.
+
+The scale mode iterates the test as a double nested loop as follows:
+```
+   for node with node_idx less-or-equal node_range(low, high. step_size):
+      for pod with pod_idx less-or-equal density_range(low, high, step_size):
+          run uperf 
+```
+Hence, with the above params, the first iteration runs the pair with node_idx/pod_idx of {1,1}. After the first
+run has completed, the second interation runs 2 pairs of {1,1} and {1,2} and so on.
+
+The valid `step_size` methods are: addN and log2. `N` can be any integer and `log2` will double the value at each iteration i.e. 1,2,4,8,16 ...
+By choosing the appropriate values for `density_range` and `node_range`, the user can generate most if not all
+combinations of UPERF data points to exercise datapath performance from many angles.
 
 Once done creating/editing the resource file, you can run it by:
 
--- a/docs/uperf.md
+++ b/docs/uperf.md
@@ -221,7 +221,7 @@ Here is one scale example:
       step_size: log2
       ...
 ```
-Note, the `scale` mode is mutual exlusive to `pin` mode with the `pin` mode has higher precedence.
+Note, the `scale` mode is mutually exlusive to `pin` mode with the `pin` mode having higher precedence.
 In other words, if `pin:true` the test will deploy pods on `pin_server` and `pin_client` nodes
 and ignore `colocate`, `node_range`, and the number of pairs to deploy is specified by the
  `density_range.high` value.
--- a/docs/uperf.md
+++ b/docs/uperf.md
@@ -41,6 +41,7 @@ spec:
       kind: pod
       pin_server: "node-0"
       pin_client: "node-1"
+      pair: 1
       multus:
         enabled: false
       samples: 1
@@ -77,6 +78,11 @@ spec:
 
 `pin_client` what node to pin the client pod to.
 
+`pair` how many instances of uperf client-server pairs. `pair` is applicable for `pin: true` only.
+If `pair` is not specified, the operator will use the value in `density_range` to detemine the number of pairs.
+See **Scale** section for more info. `density_range` can do more than `pair` can, but `pair` support is retained 
+for backward compatibility.
+
 `multus[1]` Configure our pods to use multus.
 
 `samples` how many times to run the tests. For example
@@ -85,7 +91,7 @@ spec:
 
 ```yaml
       samples: 3
-      density: [1,1]
+      density_range: [1,1]
       test_types:
         - stream
       protos:
@@ -111,7 +117,7 @@ size.
 For example:
 ```yaml
       samples: 3
-      density: [1,1]
+      density_range: [1,1]
       test_types:
         - rr
       protos:
--- a/resources/operator.yaml
+++ b/resources/operator.yaml
@@ -34,14 +34,14 @@ spec:
           - /usr/local/bin/ao-logs
           - /tmp/ansible-operator/runner
           - stdout
-          image: quay.io/hnhan/benchmark-operator:uperf-scale
+          image: quay.io/benchmark-operator/benchmark-operator:master
           imagePullPolicy: "Always"
           volumeMounts:
           - mountPath: /tmp/ansible-operator/runner
             name: runner
             readOnly: true
         - name: benchmark-operator
-          image: quay.io/hnhan/benchmark-operator:uperf-scale
+          image: quay.io/benchmark-operator/benchmark-operator:master
           imagePullPolicy: Always
           env:
             - name: WATCH_NAMESPACE
--- a/resources/operator.yaml
+++ b/resources/operator.yaml
@@ -34,14 +34,14 @@ spec:
           - /usr/local/bin/ao-logs
           - /tmp/ansible-operator/runner
           - stdout
-          image: quay.io/benchmark-operator/benchmark-operator:master
+          image: quay.io/hnhan/benchmark-operator:uperf-scale
           imagePullPolicy: "Always"
           volumeMounts:
           - mountPath: /tmp/ansible-operator/runner
             name: runner
             readOnly: true
         - name: benchmark-operator
-          image: quay.io/benchmark-operator/benchmark-operator:master
+          image: quay.io/hnhan/benchmark-operator:uperf-scale
           imagePullPolicy: Always
           env:
             - name: WATCH_NAMESPACE
@@ -70,7 +70,7 @@ spec:
             - containerPort: 6379
           resources:
             limits:
-              cpu: "0.1"
+              cpu: "2.0"
           volumeMounts:
             - mountPath: /redis-master-data
               name: data
--- a/roles/uperf/tasks/cleanup.yml
+++ b/roles/uperf/tasks/cleanup.yml
@@ -31,8 +31,10 @@
     when: cleanup
 
   - name: Cleanup redis
-     command: "redis-cli del num_completion-{{trunc_uuid}}"
-     command: "redis-cli del start-{{trunc_uuid}}"
+    command: "{{ item }}"
+    with_items:
+      - redis-cli del num_completion-{{trunc_uuid}}
+      - redis-cli del start-{{trunc_uuid}}
 
   when: resource_kind == "pod"
 
--- a/roles/uperf/tasks/cleanup.yml
+++ b/roles/uperf/tasks/cleanup.yml
@@ -1,0 +1,45 @@
+---
+
+- block:
+
+  - block:
+    - name: P29 Get Server Pods
+      k8s_facts:
+        kind: Pod
+        api_version: v1
+        namespace: '{{ operator_namespace }}'
+        label_selectors:
+          - type = uperf-bench-server-{{ trunc_uuid }}
+      register: server_pods
+
+    - name: P30 Pod names - to clean
+      set_fact:
+        clean_pods: |
+            [
+            {% for item in server_pods.resources %}
+              "{{ item['metadata']['name'] }}",
+            {% endfor %}
+            ]
+
+    - name: P31 Cleanup run
+      k8s:
+        kind: pod
+        api_version: v1
+        namespace: '{{ operator_namespace }}'
+        state: absent
+        name: "{{ item }}"
+      with_items: "{{ clean_pods }}"
+      when: cleanup
+    when: resource_kind == "pod"
+
+  - operator_sdk.util.k8s_status:
+      api_version: ripsaw.cloudbulldozer.io/v1alpha1
+      kind: Benchmark
+      name: "{{ meta.name }}"
+      namespace: "{{ operator_namespace }}"
+      status:
+        state: Complete
+        complete: true
+
+  when: resource_state.resources[0].status.state == "Cleanup"
+
--- a/roles/uperf/tasks/cleanup.yml
+++ b/roles/uperf/tasks/cleanup.yml
@@ -1,45 +1,46 @@
 ---
 
 - block:
+  ### <POD> kind
+  - name: Get Server Pods
+    k8s_facts:
+      kind: Pod
+      api_version: v1
+      namespace: '{{ operator_namespace }}'
+      label_selectors:
+        - type = uperf-bench-server-{{ trunc_uuid }}
+    register: server_pods
 
-  - block:
-    - name: Get Server Pods
-      k8s_facts:
-        kind: Pod
-        api_version: v1
-        namespace: '{{ operator_namespace }}'
-        label_selectors:
-          - type = uperf-bench-server-{{ trunc_uuid }}
-      register: server_pods
+  - name: Pod names - to clean
+    set_fact:
+      clean_pods: |
+          [
+          {% for item in server_pods.resources %}
+            "{{ item['metadata']['name'] }}",
+          {% endfor %}
+          ]
 
-    - name: Pod names - to clean
-      set_fact:
-        clean_pods: |
-            [
-            {% for item in server_pods.resources %}
-              "{{ item['metadata']['name'] }}",
-            {% endfor %}
-            ]
+  - name: Cleanup run
+    k8s:
+      kind: pod
+      api_version: v1
+      namespace: '{{ operator_namespace }}'
+      state: absent
+      name: "{{ item }}"
+    with_items: "{{ clean_pods }}"
+    when: cleanup
+  when: resource_kind == "pod"
 
-    - name: Cleanup run
-      k8s:
-        kind: pod
-        api_version: v1
-        namespace: '{{ operator_namespace }}'
-        state: absent
-        name: "{{ item }}"
-      with_items: "{{ clean_pods }}"
-      when: cleanup
-    when: resource_kind == "pod"
+#
+# no <VM> kind block - We leave VM running
+#
 
-  - operator_sdk.util.k8s_status:
-      api_version: ripsaw.cloudbulldozer.io/v1alpha1
-      kind: Benchmark
-      name: "{{ meta.name }}"
-      namespace: "{{ operator_namespace }}"
-      status:
-        state: Complete
-        complete: true
-
-  when: resource_state.resources[0].status.state == "Cleanup"
+- operator_sdk.util.k8s_status:
+    api_version: ripsaw.cloudbulldozer.io/v1alpha1
+    kind: Benchmark
+    name: "{{ meta.name }}"
+    namespace: "{{ operator_namespace }}"
+    status:
+      state: Complete
+      complete: true
 
--- a/roles/uperf/tasks/cleanup.yml
+++ b/roles/uperf/tasks/cleanup.yml
@@ -2,6 +2,16 @@
 
 - block:
   ### <POD> kind
+  # Cleanup servers, but leave clients around mostly for further examining of results.
+  - name: Get Server Jobs
+    k8s_facts:
+      kind: Job
+      api_version: v1
+      namespace: '{{ operator_namespace }}'
+      label_selectors:
+        - type = uperf-bench-server-{{ trunc_uuid }}
+    register: server_jobs
+
   - name: Get Server Pods
     k8s_facts:
       kind: Pod
@@ -11,8 +21,14 @@
         - type = uperf-bench-server-{{ trunc_uuid }}
     register: server_pods
 
-  - name: Pod names - to clean
+  - name: Server Job and Pod names - to clean
     set_fact:
+      clean_jobs: |
+          [
+          {% for item in server_jobs.resources %}
+            "{{ item['metadata']['name'] }}",
+          {% endfor %}
+          ]
       clean_pods: |
           [
           {% for item in server_pods.resources %}
@@ -20,23 +36,34 @@
           {% endfor %}
           ]
 
-  - name: Cleanup run
+  - name: Cleanup server Job 
     k8s:
-      kind: pod
+      kind: Job
+      api_version: v1
+      namespace: '{{ operator_namespace }}'
+      state: absent
+      name: "{{ item }}"
+    with_items: "{{ clean_jobs }}"
+
+  - name: Cleanup server Pod 
+    k8s:
+      kind: Pod
       api_version: v1
       namespace: '{{ operator_namespace }}'
       state: absent
       name: "{{ item }}"
     with_items: "{{ clean_pods }}"
-    when: cleanup
 
+  when: resource_kind == "pod" and cleanup == True
+
+- block:
   - name: Cleanup redis
     command: "{{ item }}"
     with_items:
       - redis-cli del num_completion-{{trunc_uuid}}
       - redis-cli del start-{{trunc_uuid}}
+  when: resource_kind == "pod" 
 
-  when: resource_kind == "pod"
 
 
 
--- a/roles/uperf/tasks/cleanup.yml
+++ b/roles/uperf/tasks/cleanup.yml
@@ -3,7 +3,7 @@
 - block:
 
   - block:
-    - name: P29 Get Server Pods
+    - name: Get Server Pods
       k8s_facts:
         kind: Pod
         api_version: v1
@@ -12,7 +12,7 @@
           - type = uperf-bench-server-{{ trunc_uuid }}
       register: server_pods
 
-    - name: P30 Pod names - to clean
+    - name: Pod names - to clean
       set_fact:
         clean_pods: |
             [
@@ -21,7 +21,7 @@
             {% endfor %}
             ]
 
-    - name: P31 Cleanup run
+    - name: Cleanup run
       k8s:
         kind: pod
         api_version: v1
--- a/roles/uperf/tasks/cleanup.yml
+++ b/roles/uperf/tasks/cleanup.yml
@@ -29,7 +29,14 @@
       name: "{{ item }}"
     with_items: "{{ clean_pods }}"
     when: cleanup
+
+  - name: Cleanup redis
+     command: "redis-cli del num_completion-{{trunc_uuid}}"
+     command: "redis-cli del start-{{trunc_uuid}}"
+
   when: resource_kind == "pod"
+
+
 
 #
 # no <VM> kind block - We leave VM running
--- a/roles/uperf/tasks/cleanup.yml
+++ b/roles/uperf/tasks/cleanup.yml
@@ -9,7 +9,7 @@
       api_version: v1
       namespace: '{{ operator_namespace }}'
       label_selectors:
-        - type = uperf-bench-server-{{ trunc_uuid }}
+        - type = {{ meta.name }}-bench-server-{{ trunc_uuid }}
     register: server_jobs
 
   - name: Get Server Pods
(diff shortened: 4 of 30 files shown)


resources/operator.yaml
```diff
@@ -34,14 +34,14 @@ spec:
           - /usr/local/bin/ao-logs
           - /tmp/ansible-operator/runner
           - stdout
-          image: quay.io/benchmark-operator/benchmark-operator:master
+          image: quay.io/hnhan/benchmark-operator:uperf-scale
```
jtaleric, line 37: Need to leave this as our operator image.

roles/uperf/templates/workload.yml.j2
```diff
@@ -1,103 +1,139 @@
...
+                  labelSelector:
+                    matchExpressions:
+                    - key: app
+                      operator: In
+                      values:
+                      - {{ item.metadata.labels.app }}
+                  topologyKey: kubernetes.io/hostname
+          containers:
+          - name: benchmark
+            image: {{ workload_args.image | default('quay.io/lblyth/lblyth-uperf:latest') }}
```
jtaleric, line 49: Return to the cloud-bulldozer image.

docs/uperf.md
```diff
@@ -44,7 +44,6 @@ spec:
       multus:
         enabled: false
       samples: 1
-      pair: 1
```
jtaleric, line 47: so, I wonder if we can keep pairs for the simple use case... Similar to the pin/collocate options.... if pin is set, it ignores the other options.

resources/operator.yaml
HughNhan: Will chnage this and the other images back to official

roles/uperf/templates/workload.yml.j2
HughNhan: Yes.

docs/uperf.md
HughNhan: OK. Will keep pair and make it higher precedence than 'density_range' for pin mode.

jtaleric:
nice, so we addressed the 

```
{\\\"reason\\\":\\\"FieldValueInvalid\\\",\\\"message\\\":\\\"Invalid value: \\\\\\\\\\\"uperf-bench-server-ip-10-0-133-30.us-west-2.compute.internal-0-d444e7b3\\\\\\\\\\\": must be no more than 63 characters\\\",\\\"field\\\":\\\"spec.template.labels\\\"},
```

Uncovered by the CI, but UPerf is still having issues...

jtaleric:
@HughNhan seems like we are still hitting the same issue?

```
is invalid: [metadata.labels: Invalid value: \\\\\\\\\\\"uperf-server-ip-10-0-133-30.us-west-2.compute.internal-0-69610215\\\\\\\\\\\": must be no more than 63 characters, spec.template.labels: Invalid value: \\\\\\\\\\\"uperf-server-ip-10-0-133-30.us-west-2.compute.internal-0-69610215\\\\\\\\\\\": must be no more than 63 characters]\\\",\\\"reason\\\":\\\"Invalid\\\",\\\"details\\\":{\\\"name\\\":\\\"uperf-server-ip-10-0-133-30.us-west-2.compute.internal-0-69610215\\\",\\\"group\\\":\\\"batch\\\",\\\"kind\\\":\\\"Job\\\",\\\"causes\\\":[{\\\"reason\\\":\\\"FieldValueInvalid\\\",\\\"message\\\":\\\"Invalid value: \\\\\\\\\\\"uperf-server-ip-10-0-133-30.us-west-2.compute.internal-0-69610215\\\\\\\\\\\": must be no more than 63 characters\\\",\\\"field\\\":\\\"metadata.labels\\\"}
```

roles/uperf/templates/server.yml.j2
```diff
@@ -1,47 +1,65 @@
...
+    metadata:    
+      name: 'uperf-server-{{worker_node_list[ node_idx_item ] }}-{{ item }}-{{ trunc_uuid }}'
+      namespace: "{{ operator_namespace }}"
+    spec:
+      ttlSecondsAfterFinished: 600
+      backoffLimit: 0
+      template:
+        metadata:
+          labels:
+            app: uperf-bench-server-{{ worker_node_list[node_idx_item] | truncate(16,true,'') }}-{{ item  }}-{{ trunc_uuid }}
```
jtaleric, line 18: s/bench// 

I wonder if we can do something like:  `worker_node_list[node_idx_item].split('.')[0]` 

Which would get us something like `uperf-server-ip-10-0-133-30`

roles/uperf/templates/service.yml.j2
```diff
@@ -1,27 +1,46 @@
...
-    protocol: TCP
+kind: List
+metadata: {}
+items:
+{% macro job_template(item, node_idx_item='') %}
+  - kind: Service
+    apiVersion: v1
+    metadata:
+      #name: uperf-service-{{ item }}-{{ trunc_uuid }}
+      name:  uperf-service-{{worker_node_list[ node_idx_item ] }}-{{ item }}-{{ trunc_uuid }}
```
jtaleric, line 11: Similar to my comment above, I wonder if we can do something like: worker_node_list[node_idx_item].split('.')[0]

roles/uperf/templates/server.yml.j2
HughNhan: I intentionally put node name in there to identify them easier. Truncating the long node name still leaves good context.

docs/uperf.md
```diff
@@ -189,6 +192,55 @@ To enable Multus in Ripsaw, here is the relevant config.
...
+Note, the `scale` mode is mutual exlusive to `pin` mode with the `pin` mode has higher precedence.
+In other words, if `pin:true` the test will deploy pods on `pin_server` and `pin_client` nodes
+and ignore `colocate`, `node_range`, and the number of pairs to deploy is specified by the
+ `density_range.high` value.
+
+In the above sample, the `scale` mode will be activated since `pin: false`. In the first phase, the 
+pod instantion phase, the system gathers node inventory and may reduce the `node_range.high` value 
+to match the number of worker node available in the cluster.
+
+According to `node_range: [1,128]`, and `density_range:[1,10]`, the system will instantiate 10 pairs on
```
rsevilla87, line 227: Update "will instantiate 10 pairs" by "will instantiate up to 10 pairs"

deploy/25_role.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
rsevilla87, line 1: I guess we need this clusterrole to get and label nodes, right?

roles/uperf/tasks/start_server.yml
```diff
@@ -0,0 +1,74 @@
+---
+
+- block:
+  ### <POD> kind
+  - include_tasks: init.yml
```
rsevilla87, line 5: We can get rid of init.yml file and include the tasks here

roles/uperf/tasks/setup.yml
```diff
@@ -0,0 +1,167 @@
...
+    #
+    # This block is for scale mode where client and server pods are spreaded
+    # across all eligible nodes
+    #
+  - name: List Nodes Labeled as Workers
+    k8s_info:
+      api_version: v1
+      kind: Node
+      label_selectors:
+        - "node-role.kubernetes.io/worker="
```
rsevilla87, line 48: Can you also include node-role.kubernetes.io/infra!= and node-role.kubernetes.io/workload!= as well? There're situations when those nodes are labeled as workers and we want to get them out of this list.

deploy/25_role.yaml
HughNhan: To be able to see all nodes, we need clusterrole

roles/uperf/tasks/setup.yml
```diff
@@ -0,0 +1,167 @@
...
+        node_hi_idx: "{{ workload_args.node_range[1]  | default('1')|int - 1 }}"
+    #
+    # Next sanity check and massage the indices if necessary.
+    # We shall complete gracefully and not iterate wildly.
+    #
+  - name: Adjust node_hi_idx if cluster has less nodes
+    set_fact:
+        node_hi_idx: "{{ worker_node_list|length| default('0')|int -1 }}"
+    when: "node_hi_idx|int >= worker_node_list|length| default('0')|int "
+
```
rsevilla87, line 92: Maybe we should consider hanging if  # worker nodes < pod_hi_idx

roles/uperf/tasks/start_server.yml
HughNhan: There are options, and they should be fine too. I thought the init part although being used by start_server task only, but it is a little too specific and has churned more often at least while I was working on  this pr. That was the reason I separated it out.

roles/uperf/tasks/setup.yml
HughNhan: This PR supports the "exclude_labels:[]" for this purpose. Is it what you are looking for?

roles/uperf/tasks/setup.yml
HughNhan: "node" goes with "node_range". "node_hi_idx" does not associate with pod/density_range.

roles/uperf/templates/service.yml.j2
```diff
@@ -1,27 +1,46 @@
...
-  selector:
-    app: uperf-bench-server-{{ item }}-{{ trunc_uuid }}
-  ports:
-  - name: uperf
-    port: 20000
-    targetPort: 20000
-    protocol: TCP
+kind: List
+metadata: {}
+items:
```
rsevilla87, line 5: This is a list of resources with only one service, we should get rid of this

roles/uperf/templates/service.yml.j2
```diff
@@ -1,27 +1,46 @@
...
+{% macro job_template(item, node_idx_item='') %}
+  - kind: Service
+    apiVersion: v1
+    metadata:
+      #name: uperf-service-{{ item }}-{{ trunc_uuid }}
```
rsevilla87, line 10: Can you remove these comments?

roles/uperf/templates/service.yml.j2
```diff
@@ -1,27 +1,46 @@
...
+      name:  uperf-service-{{worker_node_list[ node_idx_item ] | truncate(32,true,'') }}-{{ item }}-{{ trunc_uuid }}
+
+      namespace: '{{ operator_namespace }}'
+      labels:
+        #app: uperf-bench-server-{{ item }}-{{ trunc_uuid }}
```
rsevilla87, line 15: same

roles/uperf/templates/service.yml.j2
```diff
@@ -1,27 +1,46 @@
...
+        app:  uperf-bench-server-{{worker_node_list[ node_idx_item ] | truncate(32,true,'') }}-{{ item }}-{{ trunc_uuid }}
+        type: uperf-bench-server-{{ trunc_uuid }}
+      annotations:
+            node_idx: '{{ node_idx_item }}'
+            pod_idx: '{{ item }}'
+    spec:
+      selector:
+        #app: uperf-bench-server-{{ item }}-{{ trunc_uuid }}
```
rsevilla87, line 23: and again

roles/uperf/templates/service.yml.j2
HughNhan: Good catch. Will do it and all similar

smalleni (commented):
Quick question @HughNhan. Can you show me where in the code `colocate` is being used to determine placement of the client pods in relation to the servers. I'd just like to see how it is being used

HughNhan:
> Quick question @HughNhan. Can you show me where in the code `colocate` is being used to determine placement of the client pods in relation to the servers. I'd just like to see how it is being used

Sai - it is at https://github.com/cloud-bulldozer/benchmark-operator/pull/525/files#diff-99e2408015ad226f7a189e6b020a3405875d80fb50a369577c2b0458c6d2f8f2R193

smalleni:
> > Quick question @HughNhan. Can you show me where in the code `colocate` is being used to determine placement of the client pods in relation to the servers. I'd just like to see how it is being used
> 
> Sai - it is at https://github.com/cloud-bulldozer/benchmark-operator/pull/525/files#diff-99e2408015ad226f7a189e6b020a3405875d80fb50a369577c2b0458c6d2f8f2R193

Thank you! YUGE PR, so taking thing slowly:)

docs/uperf.md
HughNhan: It will instantiate 10 pairs not "up to".  For nodes we may reduce if the number of nodes in the cluster is less than CR. However, for pod, it will do as density_range specifies.

roles/uperf/templates/service.yml.j2
HughNhan: Ditto

roles/uperf/tasks/start_client.yml
```diff
@@ -0,0 +1,73 @@
...
+    definition: "{{ lookup('template', 'configmap.yml.j2') | from_yaml }}"
+
+- block:
+  ### <POD> kind 
+  - name: Start Client(s) w/o serviceIP
+    k8s:
+      definition: "{{ lookup('template', 'workload.yml.j2') | from_yaml }}"
+    vars: 
+      resource_item: "{{ server_pods.resources }}"
+    when: workload_args.serviceip is defined and not workload_args.serviceip|default('false') and server_pods.resources|length > 0
```
smalleni, line 23: We don't need a default value here as you are checking if it's defined before checking for false, correct?

smalleni (changes requested):

deploy/25_role.yaml
rsevilla87: In that case we only need those permissions

```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: node-reader
rules:
- apiGroups:
  - ""
  resources:
  - nodes
  verbs:
  - get
  - list
  - patch
```

roles/uperf/tasks/setup.yml
rsevilla87: I actually was suggesting to change the logic in setup.yml to hang if the number of nodes is lower than the higher number of pods to deploy (I think density_range[1]).

roles/uperf/tasks/start_client.yml
rsevilla87: I think it could be shortened with: `not workload_args.serviceip|default('false', True) and server_pods.resources|length > 0`

roles/uperf/templates/workload.yml.j2
```diff
@@ -1,105 +1,142 @@
...
-        command: ["/bin/sh", "-c"]
-        args:
+            imagePullPolicy: Always
+            command: ["/bin/sh", "-c"]
+            args:
 {% if workload_args.serviceip is sameas true %}
-          - "export serviceip=true;
-            export h={{item.spec.clusterIP}};
+              - "export serviceip=true;
+              export h={{item.spec.clusterIP}};
```
rsevilla87, line 91: Check indent

roles/uperf/templates/workload.yml.j2
```diff
@@ -1,105 +1,142 @@
...
 {% else %}
 {% if workload_args.multus.client is defined %}
-          - "export multus_client={{workload_args.multus.client}};
-            export h={{ (item['metadata']['annotations']['k8s.v1.cni.cncf.io/networks-status'] | from_json)[1]['ips'][0] }};
+              - "export multus_client={{workload_args.multus.client}};
+              export h={{ (item['metadata']['annotations']['k8s.v1.cni.cncf.io/networks-status'] | from_json)[1]['ips'][0] }};
```
rsevilla87, line 95: Check indent

roles/uperf/tasks/init.yml
```diff
@@ -0,0 +1,22 @@
...
+     kind: Benchmark
+     name: "{{ meta.name }}"
+     namespace: "{{ operator_namespace }}"
+     status:
+        pod_hi_idx: "{{pod_hi_idx}}"
+        pod_low_idx: "{{pod_low_idx}}"
+        node_hi_idx: "{{node_hi_idx}}"
+        node_low_idx: "{{node_low_idx}}"
+        node_idx: "{{node_low_idx}}"
+        pod_idx: "{{pod_low_idx}}"
```
rsevilla87, line 21: Why are we setting these variables in the bechmark status?

We should use redis to store these kind of data.

HughNhan: There are a couple of reasons:
1. redis should be used for shared data between the operator and the workload pods. These variables are the operator's private data like state i.e "Building" - The workload pods do not use/see them.
2. redis-server has already shown performance pressure in scale runs. We must scrub all unnecessary redis usages.
3. Putting these variables in benchmark status allows a *natural* way to monitor the progress by invoking "oc describe benchmark" and see.
Reason (1) and (2) are compelling.

roles/uperf/tasks/setup.yml
HughNhan: Robustness is very desirable if not required. This is a universal goal we should continue to try.  Regarding about your statement "number of nodes is lower than the number of pods", this is irrelevant because there is no restriction in that regard. You can run node_range:[2,2], density_range[1,100]. In this case we have 100 pairs per nodes for total of 200 pairs.

roles/uperf/templates/workload.yml.j2
HughNhan: Good catch. I will fix them all.

deploy/25_role.yaml
HughNhan: OK. reduce the permission scope to your suggestion

roles/uperf/tasks/start_client.yml
HughNhan: Yes, it was extraneous. Fix done.

roles/uperf/templates/service.yml.j2
HughNhan: fixed.

roles/uperf/tasks/setup.yml
```diff
@@ -0,0 +1,167 @@
...
+      api_version: v1
+      kind: Node
+      label_selectors:
+        - "node-role.kubernetes.io/worker="
+    register: node_list
+    no_log: True
+  
+  - name: Isolate Worker Role Hostnames
+    set_fact:
+      worker_node_list: "{{ node_list | json_query('resources[].metadata.name') | list }}"
```
rsevilla87, line 54: As we talked internally I think this expression would fit better and prevent some of the issues the CI has.

```suggestion
      worker_node_list: "{{ node_list | json_query('resources[].metadata.metadata.labels.kubernetes\.io/hostname') | list }}"
```

HughNhan: Yes. Let me do it

rsevilla87:
CI is failing since the labels from the uperf server have been updated. CI expects:
app=uperf-bench-server-0-$uuid

And they it has been updated to:

```
app: uperf-bench-server-{{ worker_node_list[node_idx_item] | truncate(32,true,'') }}-{{ item  }}-{{ trunc_uuid }}
```

In order to fix it my suggestion is to use a different rather than "app" to avoid modifying the CI scripts.

roles/uperf/templates/service.yml.j2
```diff
@@ -1,27 +1,43 @@
...
+      namespace: '{{ operator_namespace }}'
+      labels:
+        app:  uperf-bench-server-{{worker_node_list[ node_idx_item ] | truncate(32,true,'') }}-{{ item }}-{{ trunc_uuid }}
+        type: uperf-bench-server-{{ trunc_uuid }}
+      annotations:
+            node_idx: '{{ node_idx_item }}'
+            pod_idx: '{{ item }}'
+    spec:
+      selector:
+        app:  uperf-bench-server-{{worker_node_list[ node_idx_item ] | truncate(32,true,'')}}-{{ item }}-{{ trunc_uuid }}
```
rsevilla87, line 21: The service selector also needs to be updated.

```suggestion
        app: uperf-bench-server-{{ item  }}-{{ trunc_uuid }}
```

rsevilla87:
@HughNhan I reverted the changes in the uperf server labels and updated the CI testcase to use a different label Selector

docs/uperf.md
```diff
@@ -108,7 +111,7 @@ size.
 For example:
 ```yaml
       samples: 3
-      pair: 1
```
jtaleric, line 111: can we have an example of pairs and an example of density?

resources/operator.yaml
```diff
@@ -70,7 +70,7 @@ spec:
             - containerPort: 6379
           resources:
             limits:
-              cpu: "0.1"
+              cpu: "2.0"
```
jtaleric, line 73: This is a pretty big change... Any specific reason ?

jtaleric (changes requested):

docs/uperf.md
HughNhan: ok. Will add. 
TL;TR we support both 'pair' and 'density' notations with 'pair' of higher precedence. If 'pair' is defined, it will use 'pair', and the behavior is backward compatible i.e 'pair' specifies number of client-server pairs. In the absence of  'pair', it will use 'density_range', and 'density_range' supports iteration. For example denstity_range: [4,4] is the same as pair: 4.  density_range: [1,4] will iterate 1 pair, 2 pair, 3 and 4.

resources/operator.yaml
HughNhan: When running scale, I found that each skylake core used by redis-server can sustain about 1500 redis ops per second. We had runs with 200 nodes x 16 density = 3200 client pods. Hence it needs 2 cores. Besides upping the cpu to 2, I also put 0.5 sec delays in the client's while loop that spins and pulls redis for 'start' signal.

rsevilla87:
@HughNhan , during my tests I've found an issue in the cleanup phase.
```
 $ oc get benchmark
NAME              TYPE    STATE     METADATA STATE   CERBERUS        UUID                                   AGE
uperf-benchmark   uperf   Cleanup   not collected    not connected   102ba633-5a23-5158-b629-5d9b1e9f7397   15m

$ oc logs -f benchmark-operator-58945b66b8-9625b -c ansible --tail 100
TASK [uperf : include_tasks] ***************************************************
task path: /opt/ansible/roles/uperf/tasks/main.yml:39
fatal: [localhost]: FAILED! => {
    "reason": "We were unable to read either as JSON nor YAML, these are the errors we got from each:\nJSON: Expecting value: line 1 column 1 (char 0)\n\nSyntax Error while loading YAML.\n  mapping values are not allowed here\n\nThe error appears to be in '/opt/ansible/roles/uperf/tasks/cleanup.yml': line 34, column 13, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n  - name: Cleanup redis\n     command: \"redis-cli del num_completion-{{trunc_uuid}}\"\n            ^ here\nWe could be wrong, but this one looks like it might be an issue with\nmissing quotes. Always quote template expression brackets when they\nstart a value. For instance:\n\n    with_items:\n      - {{ foo }}\n\nShould be written as:\n\n    with_items:\n      - \"{{ foo }}\"\n"
}

```

# Found this issue with:
      colocate: false
      density_range: [1, 3]
      node_range: [1, 3]

roles/uperf/tasks/cleanup.yml
```diff
@@ -0,0 +1,53 @@
...
+      api_version: v1
+      namespace: '{{ operator_namespace }}'
+      state: absent
+      name: "{{ item }}"
+    with_items: "{{ clean_pods }}"
+    when: cleanup
+
+  - name: Cleanup redis
+     command: "redis-cli del num_completion-{{trunc_uuid}}"
+     command: "redis-cli del start-{{trunc_uuid}}"
```
rsevilla87, line 35: The issue with cleanup is here

rsevilla87: Bad indentation + only 1 module execution per task is allowed

roles/uperf/tasks/cleanup.yml
```diff
@@ -0,0 +1,55 @@
...
+  - name: Pod names - to clean
+    set_fact:
+      clean_pods: |
+          [
+          {% for item in server_pods.resources %}
+            "{{ item['metadata']['name'] }}",
+          {% endfor %}
+          ]
+
+  - name: Cleanup run
```
rsevilla87, line 23: We must delete jobs rather than pods. if we don't do that jobs will be restarted by the kube job controller.
i.e
```
$ oc get job
oc get pod
NAME                                      COMPLETIONS   DURATION   AGE
uperf-client-10.131.0.74-102ba633         1/1           9m47s      5h8m
uperf-client-10.131.0.75-102ba633         1/1           9m48s      5h8m
uperf-client-10.131.0.76-102ba633         1/1           9m48s      5h8m
uperf-client-10.132.0.100-102ba633        1/1           9m47s      5h8m
uperf-client-10.132.0.101-102ba633        1/1           9m48s      5h8m
uperf-client-10.132.0.99-102ba633         1/1           9m48s      5h8m
uperf-client-10.133.0.100-102ba633        1/1           9m48s      5h8m
uperf-client-10.133.0.98-102ba633         1/1           9m48s      5h8m
uperf-client-10.133.0.99-102ba633         1/1           9m48s      5h8m
uperf-server-ip-10-0-149-84-0-102ba633    0/1           5h9m       5h9m
uperf-server-ip-10-0-149-84-1-102ba633    0/1           5h9m       5h9m
uperf-server-ip-10-0-149-84-2-102ba633    0/1           5h9m       5h9m
uperf-server-ip-10-0-187-173-0-102ba633   0/1           5h9m       5h9m
uperf-server-ip-10-0-187-173-1-102ba633   0/1           5h9m       5h9m
uperf-server-ip-10-0-187-173-2-102ba633   0/1           5h9m       5h9m
uperf-server-ip-10-0-220-135-0-102ba633   0/1           5h9m       5h9m
uperf-server-ip-10-0-220-135-1-102ba633   0/1           5h9m       5h9m
uperf-server-ip-10-0-220-135-2-102ba633   0/1           5h9m       5h9m
```

These jobs were created some hours ago  ^

The server pods were recreated (age 15m)
```
$ oc get pod
NAME                                            READY   STATUS      RESTARTS   AGE
benchmark-operator-58945b66b8-srzd6             3/3     Running     0          16m
uperf-client-10.131.0.74-102ba633-x69c8         0/1     Completed   0          5h8m
uperf-client-10.131.0.75-102ba633-5vx8m         0/1     Completed   0          5h8m
uperf-client-10.131.0.76-102ba633-q9lhh         0/1     Completed   0          5h8m
uperf-client-10.132.0.100-102ba633-qb29z        0/1     Completed   0          5h8m
uperf-client-10.132.0.101-102ba633-2tzd5        0/1     Completed   0          5h8m
uperf-client-10.132.0.99-102ba633-d2rqx         0/1     Completed   0          5h8m
uperf-client-10.133.0.100-102ba633-z4vft        0/1     Completed   0          5h8m
uperf-client-10.133.0.98-102ba633-p5rzc         0/1     Completed   0          5h8m
uperf-client-10.133.0.99-102ba633-9pvkp         0/1     Completed   0          5h8m
uperf-server-ip-10-0-149-84-0-102ba633-r2fbn    1/1     Running     0          15m
uperf-server-ip-10-0-149-84-1-102ba633-2kxgg    1/1     Running     0          15m
uperf-server-ip-10-0-149-84-2-102ba633-86t9h    1/1     Running     0          15m
uperf-server-ip-10-0-187-173-0-102ba633-5jmbn   1/1     Running     0          15m
uperf-server-ip-10-0-187-173-1-102ba633-gprm8   1/1     Running     0          15m
uperf-server-ip-10-0-187-173-2-102ba633-hg4nd   1/1     Running     0          15m
uperf-server-ip-10-0-220-135-0-102ba633-t5fwm   1/1     Running     0          15m
uperf-server-ip-10-0-220-135-1-102ba633-rj7qw   1/1     Running     0          15m
uperf-server-ip-10-0-220-135-2-102ba633-h4ggn   1/1     Running     0          15m
```

On the other hand, are we keeping the client pods?, is this expected?

HughNhan: This clean up logic and behavior was of the master branch. I think it has progressed to this state unintentionally.  I am changing it so that the workload will cleanup the server pods (and server jobs), but leave the client pods around.  I have some personal tools that analyze the results for correctness. So leaving the client pods around has a purpose.

rsevilla87 (approved):
LGTM!

docs/uperf.md
```diff
@@ -189,6 +198,55 @@ To enable Multus in Ripsaw, here is the relevant config.
...
+```
+      ...
+      pin: false
+      colocate: false
+      density_range: [1,10]
+      node_range: [1,128]
+      step_size: log2
+      ...
+```
+Note, the `scale` mode is mutual exlusive to `pin` mode with the `pin` mode has higher precedence.
```
smalleni, line 224: "mutually"

smalleni: also with the `pin` mode "having" higher precendence.

HughNhan: good checks

jtaleric:
> /rerun all tests/test_uperf.sh

We are trying to debug why it is failing in benchmark-wrapper..

rsevilla87:
The latest commit has broken CI. CI tries to get the uperf-server pod using this expression:
```
[root@ip-172-31-40-110 cwd]# tail -f test_uperf.out 
+ '[' 0 == 1 ']'
+ sleep 5
++ kubectl get pods -n my-ripsaw-a2f58bdd-435c-4a1e-9edb-test-uperf -l type=uperf-bench-server-66d6884b -o name
++ wc -l
+ pod_count=0
+ '[' -z 0 ']'
+ counter=79
+ '[' 79 -eq 180 ']'
+ '[' 0 == 1 ']'
+ sleep 5
```

however the labels are incorrect now:
```
[root@ip-172-31-40-110 cwd]# oc get pod --show-labels 
NAME                                           READY   STATUS      RESTARTS   AGE     LABELS
benchmark-operator-6c4db48f44-zqjlv            3/3     Running     0          7m42s   name=benchmark-operator,pod-template-hash=6c4db48f44
uperf-client-172.30.29.24-66d6884b-x2r5c       0/1     Completed   0          6m20s   app=uperf-bench-client-66d6884b,clientfor=uperf-bench-server-ip-10-0-133-30-0-66d6884b,controller-uid=2c2f207d-1a37-4732-b689-44bf6b764e22,job-name=uperf-client-172.30.29.24-66d6884b,type=example-benchmark-bench-client-66d6884b
uperf-server-ip-10-0-133-30-0-66d6884b-r5d7z   1/1     Running     0          6m52s   app=uperf-bench-server-ip-10-0-133-30-0-66d6884b,controller-uid=c06870fb-68ba-4a5a-881d-b4f795d3a7d6,job-name=uperf-server-ip-10-0-133-30-0-66d6884b,type=example-benchmark-bench-server-66d6884b

```

mukrishn:
@rsevilla87 I had changed CI test CR name here https://github.com/cloud-bulldozer/benchmark-operator/pull/525/commits/b2e0bfe504a5cf4387606508cfa2a6aaf9a965b3?file-filters%5B%5D=.yaml#diff-2f7089c70c2c6394f267c11d6b2f020f99a861eca48692bef3de2b7492e273cfR4, it should have taken this new name or am I missing few more places ? 
Actually the recent commit will bring up the pod with dynamic label based on `meta.name` from CR, which will fix the network policy for uperf.

rsevilla87:
> @rsevilla87 I had changed CI test CR name here [b2e0bfe?file-filters%5B%5D=.yaml#diff-2f7089c70c2c6394f267c11d6b2f020f99a861eca48692bef3de2b7492e273cfR4](https://github.com/cloud-bulldozer/benchmark-operator/commit/b2e0bfe504a5cf4387606508cfa2a6aaf9a965b3?file-filters%5B%5D=.yaml#diff-2f7089c70c2c6394f267c11d6b2f020f99a861eca48692bef3de2b7492e273cfR4), it should have taken this new name or am I missing few more places ?
> Actually the recent commit will bring up the pod with dynamic label based on `meta.name` from CR, which will fix the network policy for uperf.

CI script (tests/test_uperf.sh) needs to be updated to use these new labels as well.

mukrishn:
> CI script (tests/test_uperf.sh) needs to be updated to use these new labels as well.

test_uperf.sh is already having this hardcoded label(type=uperf-bench-server-$uuid), just wonder who is adding this label `type=example-benchmark-bench-server-66d6884b` ? 
I think, it still using some old CR file(named example-benchmark) instead of my updated one.

mukrishn:
@rsevilla87 found few CI test CRs name aren't updated, committed them as well. Good to re-run.

smalleni:
Rerunning as per @mukrishn's ask

jtaleric (approved):
LGTM!

mukrishn (approved):

smalleni (approved):
LGTM

PR #526, opened 2021-02-15
Remove runAsUser in mssql

rsevilla87:
Should prevent using the `oc adm policy add-scc-to-group nonroot system:authenticated` workaround for mssql which seems causing other issues.


dry923 (approved):
LGTM!

PR #539, opened 2021-03-15
ES env var

rsevilla87:
Inject ES as env var. With this commit done in kube-burner https://github.com/cloud-bulldozer/kube-burner/pull/67, we can use environment variables to define the configuration. Adding prom_es  as envvar will allow us to overwrite the ElasticSearch endpoint in jobs using a remote configuration file, hence we can avoid setting ES secrets in the remote configuration file, as they can be configured with this mechanism


jtaleric (approved):
LGTM

jtaleric (changes requested):
one nit- this isn't covered int he doc... https://github.com/cloud-bulldozer/benchmark-operator/blob/master/docs/kube-burner.md

jtaleric (approved):
LGTM

PR #545, opened 2021-03-18
Add log generator workload

dry923:
# Description
Adding a log generator workload that will generate X messages of Y size (in bytes) per second/minute over a duration of Z minutes. If provided it will also check the backed log aggregator to confirm it received the expected number of messages. Supported backends are elasticsearch and AWS cloudwatch.

NOTE: This requires a backend storage to be deployed if we want to create a valid ci test for this.

benchmark-wrapper PR: https://github.com/cloud-bulldozer/benchmark-wrapper/pull/254 must be merged prior to this.

### DIFF
--- a/roles/log_generator/templates/log_generator.yml
+++ b/roles/log_generator/templates/log_generator.yml
@@ -1,0 +1,115 @@
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: 'log-generator-{{ trunc_uuid }}'
+  namespace: '{{ operator_namespace }}'
+spec:
+  parallelism: {{ workload_args.pod_count | default(1) | int }}
+  template:
+    metadata:
+      labels:
+        app: log-generator-{{ trunc_uuid }}
+    spec:
+{% if workload_args.runtime_class is defined %}
+      runtimeClassName: "{{ workload_args.runtime_class }}"
+{% endif %}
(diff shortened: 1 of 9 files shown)


jtaleric (commented):
Do we want to add a test? I realize that might mean ES, but I wonder if we could use cloudwatch?

dry923:
> Do we want to add a test? I realize that might mean ES, but I wonder if we could use cloudwatch?

Using cloudwatch is definitely an option. I'll look to see how to set fluentd to log there.

rsevilla87 (approved):
LGTM!

roles/log_generator/templates/log_generator.yml
```diff
@@ -0,0 +1,115 @@
...
+            value: "{{ uuid }}"
+          - name: test_user
+            value: "{{ test_user | default("ripsaw") }}"
+          - name: clustername
+            value: "{{ clustername }}"
+{% if elasticsearch is defined %}
+          - name: es
+            value: "{{ elasticsearch.url }}"
+          - name: es_index
+            value: "{{ elasticsearch.index_name | default("openshift-cluster-timings") }}"
```
rsevilla87, line 56: Should we use a different index name?

dry923: HA! good catch. Gotta love copy/paste fail :-)

PR #546, opened 2021-03-19
Adding debug workload option for snafu related jobs and enabling it for CI

dry923:
# Description
snafu allows passing the -v flag to enable debug logging. This PR allows the user to set workloads.debug to true if they want to pass the debug flag to snafu. This also enables it for all pertinent CI tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,32 @@ spec:
       image: my.location/foo:latest
 ```
 
+## Optional debug out for benchmark-wrapper workloads
+Workloads that are triggered through [benchmark-wrapper](https://github.com/cloud-bulldozer/benchmark-wrapper)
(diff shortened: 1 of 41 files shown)


rsevilla87 (approved):
awesome!, lgtm!

jtaleric (commented):
For our CI, lets print output from the workload regardless if it succeeds or fails... Thoughts?

dry923:
> For our CI, lets print output from the workload regardless if it succeeds or fails... Thoughts?

So we changed from that to only printing when it fails because the output was becoming unmanageable to view and debug. If we want to put it back we should probably do something where it writes the log output to a file that jenkins collects. That would be outside the scope of this PR though I would think.

PR #548, opened 2021-03-23
System-metrics collection

### LINKED ISSUE #542: [RFE] Capture and store system metrics for workloads
**Is your feature request related to a problem? Please describe.**
Nope, this is purely an enhancement.

**Describe the solution you'd like**
Today the `benchmark-operator` using `benchmark-wrapper` sends workload results to Elasticsearch. However, the only workload that sends Prometheus metrics (system metrics) to Elasticsearch is `kube-burner`. 

kube-burner also supports querying and indexing results without running a workload. 

The idea is to have kube-burner as a side-car container to benchmark-operator, create a simple API to kube-burner to accept start/end/metrics.yaml. 

`start` - when the workload started from benchmark-operator
`end` - when the workload completed
`metrics.yaml` - Each workload we will define a set of metrics we want to capture... There are obvious metrics we will capture with all workloads (CPU/Memory/Network/Disk) -- however there should be more targeted metrics for each workloads (ie, fio w/o OCS and fio w/ OCS). 

**Describe alternatives you've considered**
Prometheus `remote_write` with Elasticsearch. This seems like a viable solution, however we would either need to enable/disable this feature when the workload starts/ends or just stream the results during the duration of all workloads. 

We prefer to have a more targeted collection versus streaming tons of metrics in.

rsevilla87:
## Description
System metrics collection implementation. Can be enabled by adding the following lines to the prometheus section of the Benchmark CR:

```yaml
  system_metrics:
    collection: true (Defaults to false)
    prom_url: <Valid prometheus endpoint, by default https://prometheus-k8s.openshift-monitoring.svc.cluster.local:9091>
    index_name: <ES index name, defaults to system-metrics>
    es_url: <valid ES endpoint>
    prom_token: <valid prometheus token, can be obtained with oc sa get-token -n openshift-monitoring prometheus-k8s>
    metrics_profile: <valid metrics profile name or URL, by default it uses node-metrics.yml>

```
When the workload is marked as Complete, benchmark-operator creates a kube-burner based job that collect the metrics configured in the metric profile using the benchmark creationTimestamp as start date and the current date as end date. This system metrics collection is available for all workloads except kube-burner and backpack.

## Fixes
#542 

### Task list

- [x] Initial implementation
- [x] Update documentation
- [x] Include some tests

### DIFF
--- a/docs/system-metrics.md
+++ b/docs/system-metrics.md
@@ -2,16 +2,18 @@
 
 Benchmark-operator is able to collect prometheus metrics from the cluster at the end of a benchmark. To do so, it creates a k8s job that uses [kube-burner](https://github.com/cloud-bulldozer/kube-burner) to collect the Prometheus metrics given by a configuration file. This system metrics collection mechanism in available for all workloads except `kube-burner` and `backpack`.
 
-This feature is disabled by default, it can be enabled by adding the following lines to the prometheus section of the benchmark's CR.
+This feature is disabled by default, it can be enabled by adding a `system_metrics` section to a benchmark CR.
 
 ```yaml
-  prometheus:
-    system_metrics_collection: true (Defaults to false)
+  system_metrics:
+    collection: true (Defaults to false)
     prom_url: <Valid prometheus endpoint, by default https://prometheus-k8s.openshift-monitoring.svc.cluster.local:9091>
-    system_metrics_index: <ES index name, defaults to system-metrics>
+    index_name: <ES index name, defaults to system-metrics>
     es_url: <valid ES endpoint>
     prom_token: <valid prometheus token, in an OpenShift environment can be obtained with oc sa get-token -n openshift-monitoring prometheus-k8s>
     metrics_profile: <valid metric profile name or URL pointing to it, by default node-metrics.yml>
+    step: <Prometheus step size, by default 30s>
+    image: <Kube-burner image, by default quay.io/cloud-bulldozer/kube-burner:latest>
 ```
 
 As stated in the example above, `metrics_profile` points to node-metrics.yml, (this file is available within the system-metrics role of this repo), however it can be configured pointing to an external URL like in the example below:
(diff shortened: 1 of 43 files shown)


resources/crds/ripsaw_v1alpha1_ripsaw_crd.yaml
```diff
@@ -45,6 +45,13 @@ spec:
                     type: string
                   prom_token:
                     type: string
+                  system_metrics_collection:
+                    type: boolean
+                    default: false
+                  system_metrics_index:
```
dry923, line 51: In the description of the PR it states the metrics index and profile are defaulted to X. I assume that default is in the kube-burner code. Would it make sense to default it hear as well so it's more obvious to the end user?

roles/system-metrics/templates/kube-burner.yml.j2
```diff
@@ -0,0 +1,42 @@
...
+      labels:
+        app: system-metrics-collector-{{ trunc_uuid }}
+    spec:
+{% if workload_args.runtime_class is defined %}
+      runtimeClassName: "{{ workload_args.runtime_class }}"
+{% endif %}
+      restartPolicy: Never
+      containers:
+      - name: kube-burner
+        image: quay.io/cloud-bulldozer/kube-burner:latest
```
dry923, line 21: Thoughts on adding the optional image argument here as well?

rsevilla87: done

docs/system-metrics.md
```diff
@@ -0,0 +1,57 @@
...
+
+```yaml
+  system_metrics:
+    collection: true (Defaults to false)
+    prom_url: <Valid prometheus endpoint, by default https://prometheus-k8s.openshift-monitoring.svc.cluster.local:9091>
+    index_name: <ES index name, defaults to system-metrics>
+    es_url: <valid ES endpoint>
+    prom_token: <valid prometheus token, in an OpenShift environment can be obtained with oc sa get-token -n openshift-monitoring prometheus-k8s>
+    metrics_profile: <valid metric profile name or URL pointing to it, by default node-metrics.yml>
+	step: <Prometheus step size, by default 30s>
```
jtaleric, line 15: Something weird going on here with the spacing?

rsevilla87: Thanks, fixed

jtaleric (approved):
This is a great start to a big gap! 

Eventually I want to replace `node-metrics.yml` with `{{workload}}-metrics.yml` -- that way we default to a set of metrics specifically for a workload.

The base of those metrics should be `node-metrics` but I am sure there might be some special sauce we want to add. This can be tackled in a later commit. Great work @rsevilla87

chaitanyaenr (approved):
Have not tested it but the implementation LGTM, nice work @rsevilla87.

whitleykeith (approved):
lgtm

PR #551, opened 2021-03-24
Truncate UPERF node names to 27 chars max.

HughNhan:
### Description
This PR truncates names to 27 chars. It was discovered that jinja 'truncate' filter does not truncate for less than 5 chars.  So the current code that relies on truncate(32) would not take effect for names with length between 32 to 37 chars. Without truncation, these node names produce UPERF objects with meta.names and labels that exceed 63 chars violating k8s API.
### Fixes
Truncate 5 more chars.
### Test
On a cluster that has node names of 37 chars (berrymar19-9wc5t-worker-eastus1-pnfrk) UPERF failed before the fix, but with the fix it works.

### DIFF
--- a/roles/uperf/templates/server.yml.j2
+++ b/roles/uperf/templates/server.yml.j2
@@ -7,7 +7,7 @@ items:
   - kind: Job
     apiVersion: batch/v1
     metadata:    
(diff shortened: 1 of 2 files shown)


rsevilla87 (approved):
LGTM

PR #553, opened 2021-03-25
Fix ansible logs

jtaleric:
### Description
Output from our CI today is wrapping and hard to read. This is an
attempt to clean the output up.


jtaleric:
Testing... with uperf.

jtaleric:
With the migration to the new sdk, I am going to close this... I think there is more logging the new sdk provides.

PR #555, opened 2021-03-26
Always use benchmark-operator for service account

fritchie:
The service account name is hard coded in some of the workload
roles, ie:

https://github.com/cloud-bulldozer/benchmark-operator/blob/master/roles/fio_distributed/templates/servers.yaml#L48

If someone were to change the fullnameOverride in values.yaml the
workload would fail.


rsevilla87 (approved):
LGTM

PR #556, opened 2021-03-26
Add more metrics to capture during the kube-burner runs

chaitanyaenr:
This commit:
- Modifies the metrics profiles used by the kube-burner to capture the
  prometheus replicas CPU and Memory usage as the current implementation
  only collects the aggregated metrics ( average ) on the infra nodes.
- Also modifies the aggregated queries to include registry and logging
  components resource usage on the infra nodes.

### DIFF
--- a/roles/kube-burner/files/metrics-aggregated.yaml
+++ b/roles/kube-burner/files/metrics-aggregated.yaml
@@ -2,10 +2,13 @@ metrics:
   - query: (sum(irate(container_cpu_usage_seconds_total{name!="",container!="POD",namespace=~"openshift-(etcd|oauth-apiserver|sdn|ovn-kubernetes|.*apiserver|authentication|.*controller-manager|.*scheduler)"}[2m]) * 100) by (container, pod, namespace, node) and on (node) kube_node_role{role="master"}) > 0
     metricName: containerCPU-Masters
 
+  - query: (sum(irate(container_cpu_usage_seconds_total{pod!="",container="prometheus",namespace="openshift-monitoring"}[2m]) * 100) by (container, pod, namespace, node) and on (node) kube_node_role{role="infra"}) > 0
+    metricName: containerCPU-Prometheus
+
   - query: (avg(irate(container_cpu_usage_seconds_total{name!="",container!="POD",namespace=~"openshift-(sdn|ovn-kubernetes|ingress)"}[2m]) * 100 and on (node) kube_node_role{role="worker"}) by (namespace, container)) > 0
     metricName: containerCPU-AggregatedWorkers
 
-  - query: (avg(irate(container_cpu_usage_seconds_total{name!="",container!="POD",namespace=~"openshift-(sdn|ovn-kubernetes|ingress|monitoring)"}[2m]) * 100 and on (node) kube_node_role{role="infra"}) by (namespace, container)) > 0
+  - query: (avg(irate(container_cpu_usage_seconds_total{name!="",container!="POD",namespace=~"openshift-(sdn|ovn-kubernetes|ingress|monitoring|image-registry|logging)"}[2m]) * 100 and on (node) kube_node_role{role="infra"}) by (namespace, container)) > 0
     metricName: containerCPU-AggregatedInfra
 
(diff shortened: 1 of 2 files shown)


roles/kube-burner/files/metrics.yaml
```diff
@@ -1,11 +1,11 @@
 metrics:
-  - query: sum(irate(container_cpu_usage_seconds_total{name!="",namespace=~"openshift-(etcd|oauth-apiserver|.*apiserver|ovn-kubernetes|sdn|ingress|authentication|.*controller-manager|.*scheduler)"}[2m]) * 100) by (pod, namespace, node)
+  - query: sum(irate(container_cpu_usage_seconds_total{name!="",namespace=~"openshift-(etcd|oauth-apiserver|.*apiserver|ovn-kubernetes|sdn|ingress|authentication|.*controller-manager|.*scheduler|monitoring|logging|image-registry)"}[2m]) * 100) by (pod, namespace, node)
```
chaitanyaenr, line 2: This will generate lot of documents on a large scale cluster since there's going to be a replica of node-exporter, fluentd ( when logging is installed ) on each of the nodes.

chaitanyaenr: Adding specific queries to grab the resource usage of prometheus and elasticsearch pods ( if deployed ) might be the way to go. Thoughts?

rsevilla87: Looks good to me, you're right about the high number of documents generated, however, the metrics.yaml is meant to be used in smaller scale scenarios (i.e node-density), so I think your suggestion of adding a prometheus container specific metric is a good choice in metrics-aggregated.yaml.

roles/kube-burner/files/metrics-aggregated.yaml
```diff
@@ -2,10 +2,13 @@ metrics:
   - query: (sum(irate(container_cpu_usage_seconds_total{name!="",container!="POD",namespace=~"openshift-(etcd|oauth-apiserver|sdn|ovn-kubernetes|.*apiserver|authentication|.*controller-manager|.*scheduler)"}[2m]) * 100) by (container, pod, namespace, node) and on (node) kube_node_role{role="master"}) > 0
     metricName: containerCPU-Masters
 
+  - query: (sum(irate(container_cpu_usage_seconds_total{pod!="",container="prometheus",namespace="openshift-monitoring"}[2m]) * 100) by (pod, node) and on (node) kube_node_role{role="infra"}) > 0
```
rsevilla87, line 5: I think is better to aggregate by (container, pod, namespace, node) rather than only (pod, node), it will increase the number of documents, however as there's only two prometheus replicas, that increase won't be too much

roles/kube-burner/files/metrics-aggregated.yaml
```diff
@@ -14,10 +17,13 @@ metrics:
   - query: (sum(container_memory_rss{name!="",container!="POD",namespace=~"openshift-(etcd|oauth-apiserver|.*apiserver|ovn-kubernetes|sdn|ingress|authentication|.*controller-manager|.*scheduler)"}) by (container, pod, namespace, node) and on (node) kube_node_role{role="master"}) > 0
     metricName: containerMemory-Masters
 
+  - query: (sum(container_memory_rss{pod!="",namespace="openshift-monitoring",name!="",container="prometheus"}) by (pod, node) and on (node) kube_node_role{role="infra"}) > 0
```
rsevilla87, line 20: Same as above, missing aggregating by `(container, pod, namespace, node) `

roles/kube-burner/files/metrics.yaml
chaitanyaenr: Makes sense.

roles/kube-burner/files/metrics-aggregated.yaml
chaitanyaenr: Fixed it, thanks.

rsevilla87 (approved):
LGTM

jtaleric (approved):
LGTM

PR #563, opened 2021-04-12
API & golang metrics

rsevilla87:
### Description

- Add more API metrics to metrics-aggregated.yaml
- Add some comments
- Remove drop packet metrics. We should not use these reports to check for dropped packets. The dashboards provided by dittybopper should be used for this purpose
- Remove kubelet and cri-o metrics from metrics-aggregated.yaml. We don' use them here, these metrics fit better on the metrics.yaml profile (default for node-density)
- Add heap and gc metrics


whitleykeith (approved):
LGTM

jtaleric (approved):
LGTM!

PR #564, opened 2021-04-12
Enhance node-density heavy to use livenessprobes

smalleni:
Using livenessprobes in addition to readinessprobes increases the load
on the kubelet which is what we want with this test.

### DIFF
--- a/roles/kube-burner/files/app-deployment.yml
+++ b/roles/kube-burner/files/app-deployment.yml
@@ -46,7 +46,6 @@ spec:
           value: '5432'
         - name: POSTGRESQL_RETRY_INTERVAL
           value: '5'
-        imagePullPolicy: IfNotPresent
         securityContext:
           privileged: false
       restartPolicy: Always
--- a/roles/kube-burner/files/app-deployment.yml
+++ b/roles/kube-burner/files/app-deployment.yml
@@ -13,6 +13,7 @@ spec:
       containers:
       - name: perfapp
         image: quay.io/rsevilla/perfapp:latest
+        imagePullPolicy: Always
         readinessProbe:
           httpGet:
             path: /ready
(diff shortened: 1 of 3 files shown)


roles/kube-burner/templates/node-density-heavy.yml.j2
```diff
@@ -37,6 +37,7 @@ jobs:
         replicas: 1
         inputVars:
           readinessPeriod: 10
+          livenessPeriod: 10
```
rsevilla87, line 40: You need to set this var in the max-namespaces.yml.j2 template as well

roles/kube-burner/files/app-deployment.yml
```diff
@@ -21,6 +21,14 @@ spec:
...
           timeoutSeconds: 60
           initialDelaySeconds: 30
+        livenessProbe:
+          httpGet:
+            path: /health
+            port: 8080
+          periodSeconds: {{ .livenessPeriod }}
+          failureThreshold: 1
+          timeoutSeconds: 15
+          initialDelaySeconds: 30
```
rsevilla87, line 31: @smalleni can you update the imagePullPolicy to `Always`, seems it's still using the old image?

smalleni: > @smalleni can you update the imagePullPolicy to `Always`, seems it's still using the old image?

@rsevilla87 do you have triggers in your perfApp repo that would have built a new container image from the PR?

smalleni: @rsevilla87 do you have triggers in your perfApp repo that would have built a new container image from the PR?

rsevilla87:
@smalleni fixed!, we had a duplicated imagePullPolicy param. And no, I don't have a trigger for perfApp atm (I pushed the image manually yesterday)

smalleni:
> @smalleni fixed!, we had a duplicated imagePullPolicy param. And no, I don't have a trigger for perfApp atm (I pushed the image manually yesterday)

AHA, yes my bad, the default is Always anyways, so yes the the other imagePullPolicy was the problem. Thanks for taking care of it, hopefully we stress the kubelet more now :)

PR #566, opened 2021-04-16
Adding concurrent builds to benchmark operator

paigerube14:
### Description
This creates the start of the objects needed to run the concurrent builds test case. It will create a buildconfig and an image stream in each namespace. Because of the set up of the buildconfig a build and pod will also get created and run during the kube-burner run. 
The addition of starting builds concurrently seemed like too much to add to kube-burner at this time but I would be open to adding it to get the necessary benchmarks we want for this test. 

This is a little different from some of the other workloads in kube-burner because it just creates builds to later be used to start concurrent builds using the workloads/kube-burner/run_concurrentbuilds_test_fromgit.sh script in e2e-benchmarking. 

Not sure if the resources/crds/ripsaw_v1alpha1_concurrent_builds_cr.yaml file is necessary, but might be nice to give an example of working parameters for the kube-burner

### DIFF
--- a/docs/kube-burner.md
+++ b/docs/kube-burner.md
@@ -63,6 +63,13 @@ Each iteration of this workload creates the following objects:
 Each iteration of this workload creates the following object:
   - 1 pod. (sleep)
 
+- **concurrent-builds**: Creates a buildconfig, imagestream and corresponding build for a set application. **This will create as many namespaces with these objects as the configured job_iterations**. 
+See https://github.com/cloud-bulldozer/e2e-benchmarking/tree/master/workloads/kube-burner/builds for example parameters for each application
+Each iteration of this workload creates the following object:
+  - 1 imagestream (dependent on application type set)
+  - 1 buildconfig (also dependent on application type set)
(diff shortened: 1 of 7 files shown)


jtaleric:
@paigerube14 
```
The task includes an option with an undefined variable. The error was: 'dict object' has no attribute 'build_image'\\n\\nThe error appears to be in '/opt/ansible/roles/kube-burner/tasks/main.yml': line 140, column 5, but may\\nbe elsewhere in the file depending on the exact syntax problem.\\n\\nThe offending line appears to be:\\n\\n\\n  - name: Create concurrent builds\\n    ^ here\\n\"\u001b[0m\r\n\u001b[0;31m}\u001b[0m\n\r
```

paigerube14:
My latest commit was just squashing my 2 commits into one, don't think we need to rerun the perf-ci

jtaleric:
> My latest commit was just squashing my 2 commits into one, don't think we need to rerun the perf-ci

Better safe than sorry!

jtaleric (approved):
LGTM

PR #570, opened 2021-05-07
Auto osd cache drop

bengland2:
### Description

Depends on benchmark-wrapper PR 269.

This adds support for ceph OSD cache dropping, without each workload having to specify a pod IP.  
At present, it cannot automatically start the Ceph toolbox or the cache dropping pod because of a need for authorization
from openshift-storage/rook-ceph namespaces, but the user can start these pods and leave them running and the rest of it works.

for openshift, ceph toolbox can be started with 
```

$ oc patch OCSInitialization ocsinit -n openshift-storage --type json --patch  '[{ "op": "replace", "path": "/spec/enableCephTools", "value": true }]'

```
for the cache drop pod, you can start it by filling in the vars in roles/ceph_osd_cache_drop/rook_ceph_drop_cache_pod.yaml
and running the pod with kubectl/oc.

### Fixes

lack of ceph cache dropping makes benchmark-operator incomplete for OCS.


bengland2:
I'd merge it now but this depends on bohica quay.io/cloud-bulldozer/ceph-cache-dropper being updated and that hasn't rebuilt.  How to trigger rebuild?  Also I'd like to figure out how to be able to put the ceph_cache_dropper pod in my-ripsaw namespace instead of openshift-storage/rook-ceph namespace.

bengland2:
I think I'm going to have to change this so that the user is responsible for actually starting the cache dropper pod and the ceph toolbox pod, but the rest of it can stay the same.   This is because the benchmark-operator doesn't have authorization to start pods in the openshift-storage namespace.   Will try to get to this next week and finish up.

jtaleric:
@bengland2 ack - so are we good with your current implementation? LGTM..

bengland2:
@jtaleric current implementation will not work for reasons discussed above, but if I make the user create the ceph toolbox pod and the cache dropper pod using a provided YAML file, then the authorization problem goes away and the rest of it should work fine.  I still think there is a way to get some sort of token and do it but for now that's beyond what I know how to do.   At least the CR will not need to change to contain the IP addr of the cache dropper pod - benchmark-operator will discover the pod and make use of it automatically if the user specifies that Ceph OSD cache dropping is desired.  This change is small and I should be able to test it this week and merge it.  Sorry for the delays, the Multus Alias allocation required me to postpone work on this.  Thanks for the other merge.

bengland2:
I'm merging because the test_fiod.sh script passed when run against an AWS cluster, and the test result in terms of passes and fails is identical to what it was before I included OSD cache dropping tests in the PR.  Specifically there was an fio failure back 28 days ago when none of the test_crs used OSD cache dropping.   So none of the OSD cache dropping code actually executed, and fio test passed 2nd time.   This makes me think there was a timeout due to slow image load instead of actual problem with the code.   I would like to see the CI save the logs for each test run so that it would be possible to further diagnose things instead of guessing.

PR #572, opened 2021-05-10
adding cerberus validate certs parameter

kedark3:
Signed-off-by: Kedar Vijay Kulkarni <kkulkarni@redhat.com>

### Description
Cerberus may not always be running with valid certs when it is setup as an ephemeral instance, and it would be useful to provide ability to pass `validate_certs` parameter to the `uri` call. 

### Fixes
Extends CRD to add new boolean for `cerberus_validate_certs` and extends the role to accept that variable. It is being set to `false` as default as its more likely what users would want in my experience. URI Module defaults to `true` otherwise. 

CC:  @mffiedler 

Please review @chaitanyaenr

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -37,14 +37,6 @@ In the interactive rebase screen, set the first commit to `pick` and all others 
 
 Push your rebased commits (you may need to force), then issue your PR.
 
-## Container Images
-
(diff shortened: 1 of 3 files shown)


jtaleric (approved):
LGTM thanks for fixing our docs.

Issue #574, opened 2021-05-18
Add looped matrix execution to the log generator workload


dustinblack:
Many (most?) other benchmark-operator workloads employ lists for parameters in the CR files and nested loops, mostly managed via jinja2 in templates, in order to execute large matrices of tests across every unique parameter combination. The log generator workload currently only allows for single values for each parameter.

dry923:
@dustinblack we could certainly look into adding that (it should be a fairly trivial lift). Would you be looking to loop on anything besides the message size?

dustinblack:
Our test plan for bare metal, replicating your work on CloudWatch, calls for permutations through log rate, log size, and pod count, so if I understand correctly that that is a full matrix of tests it would mean we want all three of those to be lists.

PR #575, opened 2021-05-18
Add loop handling for vars to log-generator workload

dustinblack:
### Description
This adds looping through variables to the log generator workload for pods, size, and message rate.

This is tested now and I believe everything works properly with the nested loops. Thanks for the help @dry923 .

### Fixes
Issue #574

### DIFF
--- a/roles/log_generator/templates/log_generator.yml
+++ b/roles/log_generator/templates/log_generator.yml
@@ -2,14 +2,14 @@
 apiVersion: batch/v1
 kind: Job
 metadata:
-  name: 'log-generator-{{ trunc_uuid }}'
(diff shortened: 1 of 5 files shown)


roles/log_generator/templates/log_generator.yml
```diff
@@ -73,7 +73,11 @@ spec:
         args:
           - > 
 {% if workload_args.pod_count | default(1) | int > 1 %}
-            echo "Waiting for all pods to be Ready";
+{% for pods in workload_args.pod_count %}
```
dry923, line 76: as mentioned in slack the pods are spun up in the role so you would have to loop the pods there..

dry923:
@dustinblack whats the plan for this guy?

PR #577, opened 2021-05-19
Quiesce logging in pod for log generator workload

smalleni:
This along with https://github.com/cloud-bulldozer/benchmark-wrapper/pull/273
helps suppress any unneeded logging in the pod, so that we can accurately and easily count
the number of log emssages recevied in a backend like kafka merely by using the offsets.
The plan is to deploy the log generator pods in a separate namespace and forward those logs
to a topic in kafka. That way we would be able to reliably count the messages received just
by looking at kafka topic offset. Otherwise there would be other logs from the log generator pods
as well as benchmark-operator pod that would make it hard to reliably count logs received just by
kafka offset.


dry923 (approved):
LGTM!

PR #587, opened 2021-06-10
Osd cache drop rbac

bengland2:
### Description

add an RBAC YAML that permits cache dropping role to detect cache dropper pod and toolbox pod in openshift-storage namespace.   Tell OSD cache dropping users to use this YAML.   A similar thing had to be done for the kernel cache dropper.  

### Fixes

benchmark-operator got HTTP 403 access denied errors from K8S trying to detect pods running in openshift-storage namespace.

### DIFF
--- a/docs/cache_dropping.md
+++ b/docs/cache_dropping.md
@@ -58,21 +58,35 @@ Benchmarks supported for kernel cache dropping at present are:
 
 ## how to drop Ceph OSD cache
 
-for this to work with OpenShift Container Storage, you must start both the Ceph toolbox pod in OCS
-and the cache dropper pod.   You can do this with:
+for this to work with OpenShift Container Storage, you must do these steps once the benchmark-operator is running:
+and the cache dropper pod, and enable benchmark-operator to see into the openshift-storage namespace.   
+You can do this with:
 
 ```
+
+# enable the benchmark operator to look for pods in the openshift-storage namespace
+oc create -f roles/ceph_osd_cache_drop/ocs-cache-drop-clusterrole.yaml
+
+# start the Ceph toolbox pod in openshift-storage namespace
 oc patch OCSInitialization ocsinit -n openshift-storage --type json --patch \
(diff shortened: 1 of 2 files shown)


docs/cache_dropping.md
```diff
@@ -58,22 +58,36 @@ Benchmarks supported for kernel cache dropping at present are:
 
 ## how to drop Ceph OSD cache
 
-for this to work with OpenShift Container Storage, you must start both the Ceph toolbox pod in OCS
-and the cache dropper pod.   You can do this with:
+for this to work with OpenShift Container Storage, you must do these steps once the benchmark-operator is running:
```
dry923, line 61: Extremely minor nit but capital F at the start :-D

bengland2:
@dry923 does it make sense that an RBAC like this would be needed for see what pods are in the openshift-storage namespace from benchmark-operator pod?

dry923:
> @dry923 does it make sense that an RBAC like this would be needed for see what pods are in the openshift-storage namespace from benchmark-operator pod?

@bengland2 Yes. Anything outside its own namespace the default benchmark-operator account wont have access to. If you look at something like the kube-burner service account, it goes as far as assigning cluster-admin to the service account to mange the various resources outside of its namespace.

bengland2:
thx.  In that case I'm ready for this to merge, but I don't have the merge button.  it doesn't need CI testing, the CI won't run this because of its dependency on OCS, and I tested with an AWS cluster running OCS and both fio and smallfile were able to trigger cache dropping.   Will try to test for OSD memory changes resulting from this using a larger test, would just be nice to confirm that the whole thing is having the desired effect.

bengland2:
why don't I see merge button?

dry923 (approved):
LGTM!

bengland2:
the check shouldn't have failed, since there are no tests executed.

mohit-sheth (approved):
LGTM

PR #598, opened 2021-06-29
operator sdk upgrade

whitleykeith:
upgrades to the latest operator sdk version


test.sh
```diff
@@ -69,37 +76,20 @@ UUID=${NEW_UUID%-*}
...
-  sed -i "s/my-ripsaw/my-ripsaw-$UUID-$ci_dir/g" `grep -Rl my-ripsaw`
-  sed -i "s/benchmark-operator-role/benchmark-operator-role-$UUID-$ci_dir/g" `grep -Rl benchmark-operator-role`
-  cd ..
 done
 
-# Update the operator image
-update_operator_image
+delete_operator || true
+oc delete namespace benchmark-operator --ignore-not-found
+kubectl delete benchmarks -n benchmark-operator --ignore-not-found --all
```
rsevilla87, line 90: Maybe we can skip this line?, deleting the namespace benchmark-operator should remove all objects in that namespace

Makefile
```diff
@@ -0,0 +1,179 @@
...
+ifeq (,$(wildcard $(KUSTOMIZE)))
+ifeq (,$(shell which kustomize 2>/dev/null))
+	@{ \
+	set -e ;\
+	mkdir -p $(dir $(KUSTOMIZE)) ;\
+	curl -sSLo - https://github.com/kubernetes-sigs/kustomize/releases/download/kustomize/v3.5.4/kustomize_v3.5.4_$(OS)_$(ARCH).tar.gz | \
+	tar xzf - -C bin/ ;\
+	}
+else
+KUSTOMIZE = $(shell which kustomize)
```
rsevilla87, line 106: We can prescind of the kustomize binary and use `kubectl kustomize` rather than `kustomize build`

test.sh
whitleykeith: added

Makefile
whitleykeith: yeah potentially, my only concern is that the current kustomize used in [kubectl is v4.2](https://github.com/kubernetes/kubectl/blob/master/go.mod#L44) versus this one which is v3.5

whitleykeith: yeah looks like we can't just use the kubectl built kustomize as the operator-sdk stuff uses commands not present in the kubectl

see https://stackoverflow.com/questions/65439037/can-we-use-k8s-kustomize-with-kubectl-to-set-image-and-create-secret-file

rsevilla87 (approved):
LGTM!

rsevilla87:
As mentioned in Slack, let's add the following env vars to the operator deployment: 

ANSIBLE_VERBOSITY_RIPSAW_CLOUDBULLDOZER_IO: 4
MAX_CONCURRENT_RECONCILES_RIPSAW_CLOUDBULLDOZER_IO: 1

sjug (commented):
Everything else we discussed you've fixed, just a few more nits.

Makefile
```diff
@@ -0,0 +1,179 @@
+# VERSION defines the project version for the bundle.
+# Update this value when you upgrade the version of your project.
+# To re-generate a bundle for another specific version without changing the standard setup, you can:
+# - use the VERSION as arg of the bundle target (e.g make bundle VERSION=0.0.2)
+# - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
+VERSION ?= 0.0.1
```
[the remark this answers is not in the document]
whitleykeith: I believe this version may correspond to versions we use with OLM, which we don't currently do yet

README.md
```diff
@@ -3,25 +3,43 @@
...
 cd benchmark-operator/charts/benchmark-operator
-helm install benchmark-operator . -n my-ripsaw --create-namespace
+helm install benchmark-operator . -n benchmark-operator --create-namespace
 ```
 
 To delete this release, you can do so with the following command:
 
 ```bash
-helm delete benchmark-operator -n my-ripsaw --purge
+helm delete benchmark-operator -n benchmark-operator --purge
```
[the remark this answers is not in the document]
whitleykeith: will fix after the CI finish it's run so it doesn't restart

.gitignore
```diff
@@ -1,2 +1,14 @@
-# Ignore
+
+# Binaries for programs and plugins
+*.exe
```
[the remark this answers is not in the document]
whitleykeith: will fix alongside helm doc update

PR #601, opened 2021-07-06
Add nodeSelector to kernel-cache-drop-daemonset.yaml daemonset

ekuric:
For kernel cache dropper to work nodes must be labeled with 

`kernel-cache-dropper=yes`

However, in my tests these pods were starting on infra nodes
even there was not this label present on these nodes.

Adding

```
    nodeSelector:
        kernel-cache-dropper: "yes"
```

to kernel-cache-drop-daemonset.yaml ensured that kernel cache dropper
pods to start only on nodes with specified labels.

### DIFF
--- a/resources/kernel-cache-drop-daemonset.yaml
+++ b/resources/kernel-cache-drop-daemonset.yaml
@@ -18,6 +18,8 @@ spec:
       #tolerations:
       #- key: node-role.kubernetes.io/master
       #  effect: NoSchedule
+      nodeSelector:
+        kernel-cache-dropper: "yes"
       containers:
       - name: kernel-cache-dropper
         image: quay.io/cloud-bulldozer/kernel_cache_dropper:latest


ekuric:
@bengland2 PTAL

whitleykeith:
@ekuric I have included this change in https://github.com/cloud-bulldozer/benchmark-operator/pull/598

https://github.com/cloud-bulldozer/benchmark-operator/pull/598/files#diff-1801f062d62f37d376815dc7cfb64dc587d2e99d240b5cdc98bca83c2daa9f17R18

bengland2:
@ekuric  sorry I was on PTO last week just saw this.    But why do we need the node selector again?   There is already a label that identifies which nodes the kernel cache dropper is to run on.  if you look at resources/kernel-cache-drop-daemonset.yaml, I thought this was taken care of.   Is that spec:selector not sufficient?  if we need the nodeselector, then do we not need the other one?

resources/kernel-cache-drop-daemonset.yaml
```diff
@@ -18,6 +18,8 @@ spec:
       #tolerations:
       #- key: node-role.kubernetes.io/master
       #  effect: NoSchedule
+      nodeSelector:
```
bengland2, line 21: what about the spec:selector:matchLabels: component?  Does that have no effect?   Perhaps I misunderstood what that does.   [See docs](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/).  As long as you have tested it and it works, I have no problem with using the node selector.

bengland2 (approved):
I must have thought that spec:selector:matchLabels took care of that somehow, but nodeSelector may be the right way to go.

bengland2:
@ekuric Do the CI tests that use the kernel cache dropper also set the appropriate label on at least some of the K8S nodes?   Otherwise the kernel cache dropper will fail to deploy and the CI will fail.  Why did the above CI test fail when none of the tests run use the kernel cache dropper?   Also, why are these CI tests failing - they don't use the kernel cache dropper AFAICT?

PR #605, opened 2021-07-15
Enable connect type workloads in uperf

josecastillolema:
### Description

Enable connect type workloads in uperf

### Notes
For this new connect `test_type`, it is possible to provide the `sizes` values as a list of two values where the first value is the write size and the second value is the read size.

It is also possible to define a zero size in order to test connections/second (CPS) (the number of L4 connections established per second). If there is a size defined, the metric obtained will be request/second (RPS) or transactions/second (TPS) (the number of L5/L7 requests satisfied per second).

For this kind of testing, it is necessary to apply some sysctl tunning, both to the worker nodes and to the pods (see [Using sysctls in containers](https://docs.openshift.com/container-platform/4.7/nodes/containers/nodes-containers-sysctls.html)). If these parameters are not applied, two things can happen as a result of port exhaustion issues:
 - The uperf client pod will crash
 - The results will be too low

Enable the `tuned` parameter in order to get sysctl pod level tunning:
```yaml
      test_types:
        - connect
      protos:
        - tcp
      sizes:
        - [0, 128]
      nthrs:
        - 1
      runtime: 20
      tuned: true
```

For this `tuned` parameter to work, the `KubeletConfig` should allow the following sysctl parameters configuration:
 - `net.ipv4.ip_local_port_range`
 - `net.ipv4.tcp_tw_reuse`
 - `net.netfilter.nf_conntrack_tcp_timeout_time_wait`

### DIFF
--- a/docs/uperf.md
+++ b/docs/uperf.md
@@ -132,6 +132,35 @@ For example:
 Will run the `rr` test with `tcp`, first with a symmectic size of `1024` and then with an
 asymmetric size of `8192` write and `4096` read.
 
+### Connect
+
+For the connect `test_type`, it is possible to provide the `sizes` values as a list of two values where the first value is the write size and the second value is the read size.
+
+It is also possible to define a zero size in order to test connections/second (CPS) (the number of L4 connections established per second). If there is a size defined, the metric obtained will be request/second (RPS) or transactions/second (TPS) (the number of L5/L7 requests satisfied per second).
+
+For this kind of testing, it is necessary to apply some sysctl tunning, both to the worker nodes and to the pods (see [Using sysctls in containers](https://docs.openshift.com/container-platform/4.7/nodes/containers/nodes-containers-sysctls.html)). If these parameters are not applied, two things can happen as a result of port exhaustion issues:
+ - The uperf client pod will crash
+ - The results will be too low
+
+Enable the `tuned` parameter in order to get sysctl pod level tunning:
+```yaml       
+      test_types:
+        - connect
+      protos:
+        - tcp
+      sizes:
+        - [0, 128]
+      nthrs:
+        - 1
+      runtime: 20
+      tuned: true
+```
+
+For this `tuned` parameter to work, the `KubeletConfig` should allow the following sysctl parameters configuration:
+ - `net.ipv4.ip_local_port_range`
+ - `net.ipv4.tcp_tw_reuse`
+ - `net.netfilter.nf_conntrack_tcp_timeout_time_wait`
+
 ### Multus
 
 If the user desires to test with Multus, use the below Multus `NetworkAtachmentDefinition`
--- a/roles/uperf/templates/server.yml.j2
+++ b/roles/uperf/templates/server.yml.j2
@@ -57,6 +57,18 @@ items:
             - name: net.ipv4.ip_local_port_range
               value: 20000 20011
 {% endif %}
+{% if workload_args.tuned is sameas true %}
+          serviceAccountName: benchmark-operator
+          securityContext:
+            privileged: true
+            sysctls:
+            - name: net.ipv4.ip_local_port_range
+              value: 1024 65535
+            - name: net.ipv4.tcp_tw_reuse
+              value: "1"
+            - name: net.netfilter.nf_conntrack_tcp_timeout_time_wait
+              value: "1"
+{% endif %}
 {% macro metadata() %}{% include "metadata.yml.j2" %}{% endmacro %}
     {{ metadata()|indent }}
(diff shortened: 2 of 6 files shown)


roles/uperf/templates/server.yml.j2
```diff
@@ -57,6 +57,18 @@ items:
...
+          serviceAccountName: benchmark-operator
+          securityContext:
+            privileged: true
+            sysctls:
+            - name: net.ipv4.ip_local_port_range
+              value: 1024 65535
+            - name: net.ipv4.tcp_tw_reuse
+              value: "1"
+            - name: net.netfilter.nf_conntrack_tcp_timeout_time_wait
+              value: "1"
```
rsevilla87, line 70: I think this sysctl parameters won't work on this context as they are not namespaced in the Linux Kernel and can't be set at container level.
You should set them using node tuning operator with something similar to:

```yaml
apiVersion: tuned.openshift.io/v1
kind: Tuned
metadata:
  name: openshift-ingress-performance
  namespace: openshift-cluster-node-tuning-operator
spec:
  profile:
  - data: |
      [main]
      summary=Network performance profile
      include=openshift
      [sysctl]
      net.ipv4.ip_local_port_range="1024 65535"
      net.ipv4.tcp_tw_reuse=1
      net.netfilter.nf_conntrack_tcp_timeout_time_wait=1
    name: openshift-ingress-performance
  recommend:
  - match:
    - label: {node-role.kubernetes.io/workload: node-role.kubernetes.io/worker
    priority: 10
    profile: openshift-network-performance
```

josecastillolema: Hi @rsevilla87 !
Yes, those node tuning operator templates are certainly needed, but unfortunately they are not enough for this scenario, you need both (the tuning operator templates and the pod level sysctl adjustments). Otherwise, you will get the sysctl tuning applied to the workers and not into the pods (which we need for this specific testing).
I agree with you on that creating the tuned profile (and the corresponding kubelet configuration for this scenario) should be done outside benchmark-operator, and that is why we created the `tuned` parameter, that should only be enabled once proper configurations outside the scope of the operator are in place.

docs/uperf.md
```diff
@@ -132,6 +132,35 @@ For example:
 Will run the `rr` test with `tcp`, first with a symmectic size of `1024` and then with an
 asymmetric size of `8192` write and `4096` read.
 
+### Connect
+
+For the connect `test_type`, it is possible to provide the `sizes` values as a list of two values where the first value is the write size and the second value is the read size.
+
+It is also possible to define a zero size in order to test connections/second (CPS) (the number of L4 connections established per second). If there is a size defined, the metric obtained will be request/second (RPS) or transactions/second (TPS) (the number of L5/L7 requests satisfied per second).
+
+For this kind of testing, it is necessary to apply some sysctl tunning, both to the worker nodes and to the pods (see [Using sysctls in containers](https://docs.openshift.com/container-platform/4.7/nodes/containers/nodes-containers-sysctls.html)). If these parameters are not applied, two things can happen as a result of port exhaustion issues:
```
jtaleric, line 141: For this kind of testing, it is necessary to enable the `tuned` flag, to apply some sysctl tuning.

docs/uperf.md
```diff
@@ -132,6 +132,35 @@ For example:
...
+        - tcp
+      sizes:
+        - [0, 128]
+      nthrs:
+        - 1
+      runtime: 20
+      tuned: true
+```
+
+For this `tuned` parameter to work, the `KubeletConfig` should allow the following sysctl parameters configuration:
```
smalleni, line 159: @josecastillolema this perhaps needs some rework based on the recent tunings you received from TIm?

roles/uperf/templates/workload.yml.j2
```diff
@@ -195,6 +195,18 @@ items:
...
                 name: uperf-test-{{ trunc_uuid }}
           restartPolicy: OnFailure
+{% if workload_args.tuned is sameas true %}
+          serviceAccountName: benchmark-operator
+          securityContext:
+            privileged: true
+            sysctls:
+            - name: net.ipv4.ip_local_port_range
+              value: 1024 65535
+            - name: net.ipv4.tcp_tw_reuse
```
smalleni, line 205: I don't think we want this based on recent comments from Tim?

josecastillolema: After testing both tuning profiles we'll keep the original one.

docs/uperf.md
josecastillolema: After testing both tuning profiles we'll keep the original one.

PR #607, opened 2021-07-16
Fix flaky uperf

rsevilla87:
### Fixes
Uperf CI times out sometimes, it needs more time to complete.


jtaleric (commented):
Hm, I think with all the other flakes, we need to consider how we are testing vs just increasing the timeout.

rsevilla87:
> Hm, I think with all the other flakes, we need to consider how we are testing vs just increasing the timeout.

Working at the moment on a full refactor of the CI, closing this.

PR #617, opened 2021-07-28
Ripsaw CLI

whitleykeith:
### Description

Adds `ripsaw-cli` python project to the repo that is meant to be a lightweight cli tool used to install the operator and run benchmarks.


cli/ripsaw/clients/k8s.py
```diff
@@ -0,0 +1,144 @@
+from kubernetes import client, config, watch, utils
```
jtaleric, line 1: Missing License ;)

cli/ripsaw/commands/benchmark.py
```diff
@@ -0,0 +1,55 @@
+import click
```
jtaleric, line 1: License

cli/ripsaw/commands/operator.py
```diff
@@ -0,0 +1,85 @@
+import click
```
jtaleric, line 1: License

cli/ripsaw/clients/k8s.py
whitleykeith: added

.pre-commit-config.yaml
```diff
@@ -0,0 +1,73 @@
...
+  - repo: https://github.com/asottile/pyupgrade
+    rev: v2.21.2
+    hooks:
+      - id: pyupgrade
+
+  - repo: https://github.com/PyCQA/bandit
+    rev: 1.7.0
+    hooks:
+      - id: bandit
+        args: ["--exclude", "cli/tests/", "--skip", "B404,B603"]
```
learnitall, line 63: I think that this exclude might be leftover from snafu's pre-commit config? Might be ok to chuck it

.github/workflows/cli.lint.yml
```diff
@@ -0,0 +1,28 @@
+name: Code Quality Check
```
learnitall, line 1: If we want to use GHA for pre-commit rather than the pre-commit CI (https://pre-commit.ci/) than we should remove the CI configuration from the pre-commit config. I personally would prefer the pre-commit CI because it's really useful, and it's the direction I want to take benchmark-wrapper, but up to y'all.

.pylintrc
```diff
@@ -0,0 +1,23 @@
+[MASTER]
+jobs=0
+ignore-paths=
+    cli/tests
```
learnitall, line 4: This seems to also be leftover from snafu, could remove this ignore path.

cli/.coveragerc
```diff
@@ -0,0 +1,5 @@
+[run]
+omit =
+    **/__init__.py
+    ripsaw/cli.py
```
learnitall, line 4: Why are we ignoring the CLI and ``commands/`` sub-dir? Are those out of scope for the unit tests?

cli/ripsaw/clients/k8s.py
```diff
@@ -0,0 +1,169 @@
...
+                continue
+
+            if len(pods) == 0 and timeout_interval > 60:
+                raise PodsNotFoundError(f"Found no pods with label selector {label_selector}")
+
+            _ = [
+                logger.info(f"{pod.metadata.namespace}\t{pod.metadata.name}\t{pod.status.phase}")
+                for pod in pods
+            ]
+            waiting_for_pods = any(  # pylint: disable=use-a-generator
```
learnitall, line 96: This should be an easy fix that doesn't require an ignore? By replacing the below brackets with parens?

```python
any((pod.status.phase != "Running" for pod in pods))
```

cli/setup.cfg
```diff
@@ -0,0 +1,46 @@
...
+    click==8.0.1
+    kubernetes==17.17.0
+    urllib3==1.25.7
+    PyYAML==5.4.1
+    elasticsearch==7.13.4
+    GitPython==3.1.18
+python_requires = >=3.8
+
+[options.extras_require]
+tests = pytest; pytest-cov; tox; pytest-kind; ElasticMock; pytest-env
```
learnitall, line 35: Can we add these on separate lines to match the format from above?

cli/ripsaw/util/exceptions.py
```diff
@@ -0,0 +1,46 @@
...
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+
+class BenchmarkFailedError(Exception):
+    def __init__(self, name, uuid, msg=None):
+        if msg is None:
+            msg = f"The benchmark {name} with uuid {uuid} failed"
+        super(BenchmarkFailedError, self).__init__(msg)  # pylint: disable=super-with-arguments
```
learnitall, line 20: Is there a reason why this was disabled rather than changing the line to the following?

```python
super().__init__(msg)
```

learnitall (approved):
Really well done, looks great! Got my LGTM. I left a couple of nitpicky comments, but other than that this is great. Easy to read and understand.

.github/workflows/cli.lint.yml
whitleykeith: Does pre-commit CI require each dev to install/setup pre-commit on their machines? TBH I want both. I want the pre-commit hooks so that if I enable it I can catch my lint errors earlier, but I want to ensure that every PR gets linted anyways

.pylintrc
whitleykeith: same thing as the problem with the bandit exclusions. The cli testsuses fixtures which trigger `unused-argument` or `unsued-import` if the fixture is ffrom a different module. Instead of littering all the tests with pylint disables I just excluded the tests from linting since I didn't have much luck catching many issues there with them

.pre-commit-config.yaml
whitleykeith: The tests uses asserts which made bandit fail but I didn't want to remove skip the rule entirely since I want it to fail if asserts are in the actual code base

cli/.coveragerc
whitleykeith: for unit tests, yeah. The `cli.py` file is just the code entrypoint and only really executes if you run the cli itself. `commands` I will remove from there because even though it's out of scope for unit tests we shouldn't ignore it from coverage

cli/ripsaw/util/exceptions.py
whitleykeith: fixed

cli/setup.cfg
whitleykeith: fixed

cli/ripsaw/clients/k8s.py
whitleykeith: fixed

.pre-commit-config.yaml
learnitall: My bad on these `cli/tests` path comments that I left. I need to go get a pair of reading glasses or something- I mistook that for `ci/tests`. Please ignore, my bad :+1:

cli/.coveragerc
learnitall: Sounds good! Thanks for the clarification.

.github/workflows/cli.lint.yml
learnitall: Yep, yeah the only thing that the pre-commit CI will do is replace the Github Workflow. All devs should still install, setup and use pre-commit. The pre-commit CI will run on each push and on each sync of a pull-request, auto-fix pull-requests, create PRs that auto-fix code when possible, and create PRs to upgrade hook versions when possible. Additionally it's really fast because they do a lot of install caching. 

And to clarify, by "auto-fix" I mean that if a hook modifies files during its execution, pre-commit will take those modifications and submit a PR for them.

config/manager/kustomization.yaml
```diff
@@ -5,6 +5,12 @@ generatorOptions:
...
-  files:
+- files:
   - controller_manager_config.yaml
+  name: manager-config
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+images:
+- name: controller
+  newName: quay.io/cloud-bulldozer/benchmark-operator
+  newTag: latest-amd64
```
rsevilla87, line 16: I think you uploaded this file by mistake, (It gets modified whenever you make deploy)

.github/workflows/cli.lint.yml
whitleykeith: Could we use the pre-commit CI to fix the PR before merging? I really just want PRs to not get merged unless they've been linted. If we can do that then i'm fine with not using GHA

config/manager/kustomization.yaml
whitleykeith: fixed

cli/ripsaw/commands/benchmark.py
```diff
@@ -0,0 +1,75 @@
...
+
+import click
+from ripsaw.clients import elastic
+from ripsaw.clients.k8s import Cluster
+from ripsaw.models.benchmark import Benchmark
+
+
+@click.group("benchmark")
+def benchmark_group():
+    """Passthrough function for Click CLI groups, used as a decorator to declare functions for the CLI"""
```
sjug, line 28: Is this just placeholder text? (x3)

cli/ripsaw/commands/operator.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+def _perform_operator_action(repo, branch, command, kubeconfig=None):
+    """run command from operator repo"""
+    shell_env = os.environ.copy()
+    if kubeconfig is not None:
+        shell_env["KUBECONFIG"] = kubeconfig
+    local_git_repo = _find_git_repo(os.getcwd())
+    result = None
+    if "benchmark-operator" in local_git_repo:
```
sjug, line 72: Does this override the default repo/branch based on the git repo & branch if you're in one locally as it's not reflected in that initial "Installing Operator from repo..." print statement?

sjug: When not installing/deleting outside of a git repo it's not too happy:

```
$ ripsaw operator install
Installing Operator from repo https://github.com/cloud-bulldozer/benchmark-operator and branch master
  File "/home/user/go/src/github.com/cloud-bulldozer/benchmark-operator/venv/bin/ripsaw", line 33, in <module>
    sys.exit(load_entry_point('ripsaw-cli==0.0.1', 'console_scripts', 'ripsaw')())
  File "/home/user/go/src/github.com/cloud-bulldozer/benchmark-operator/venv/lib/python3.9/site-packages/click/core.py", line 1137, in __call__
    return self.main(*args, **kwargs)
  File "/home/user/go/src/github.com/cloud-bulldozer/benchmark-operator/venv/lib/python3.9/site-packages/click/core.py", line 1062, in main
    rv = self.invoke(ctx)
  File "/home/user/go/src/github.com/cloud-bulldozer/benchmark-operator/venv/lib/python3.9/site-packages/click/core.py", line 1668, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/user/go/src/github.com/cloud-bulldozer/benchmark-operator/venv/lib/python3.9/site-packages/click/core.py", line 1668, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/user/go/src/github.com/cloud-bulldozer/benchmark-operator/venv/lib/python3.9/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/user/go/src/github.com/cloud-bulldozer/benchmark-operator/venv/lib/python3.9/site-packages/click/core.py", line 763, in invoke
    return __callback(*args, **kwargs)
  File "/home/user/go/src/github.com/cloud-bulldozer/benchmark-operator/venv/lib/python3.9/site-packages/ripsaw/commands/operator.py", line 42, in install
    _perform_operator_action(repo, branch, command, kubeconfig)
  File "/home/user/go/src/github.com/cloud-bulldozer/benchmark-operator/venv/lib/python3.9/site-packages/ripsaw/commands/operator.py", line 72, in _perform_operator_action
    if "benchmark-operator" in local_git_repo:
TypeError: argument of type 'NoneType' is not iterable
```

dry923 (approved):
LGTM!

cli/ripsaw/commands/operator.py
```diff
@@ -0,0 +1,114 @@
...
+        shell_env["VERSION"] = "latest"
+        logger.info("Detected CLI is running from local repo, using that instead of remote")
+        result = subprocess.run(
```
sjug, line 75: Can we handle `CalledProcessError` for this?

cli/ripsaw/commands/benchmark.py
whitleykeith: This function leverages a feature of Click CLI to "group" commands which creates a decorator you can use (see L31). The docstring is really just there to explain why this function has to exist.

cli/ripsaw/commands/operator.py
whitleykeith: yep the conditional needed to check to make sure `local_git_repo is not None`. Just pushed a fix

Also fixed the logs around telling you what repo/branch you're on.

cli/ripsaw/commands/operator.py
whitleykeith: yeah we can, what do you think we should do with it? log/reraise/exit?

cli/ripsaw/commands/operator.py
sjug: In the case where the user provides `ripsaw operator install` a `--repo` or `--branch` inside of the git repo, can we have the CLI args override even the git repo?

cli/ripsaw/commands/operator.py
sjug: Whatever you prefer, the stack trace doesn't look like an expected behavior.

cli/ripsaw/commands/benchmark.py
sjug: Right, I understand that. Why don't we use that space to explain why this function has to exist to make this decorator string useful?

cli/ripsaw/commands/operator.py
```diff
@@ -0,0 +1,115 @@
...
+            env=shell_env,
+            cwd=local_git_repo,
+            capture_output=True,
+            check=True,
+            encoding="utf-8",
+        )
+    else:
+        with tempfile.TemporaryDirectory(dir=".") as temp_dir:
+            click.echo(f"Installing Operator from repo {repo} and branch {branch}")
+            _clone_repo(repo, branch, temp_dir.name)
```
sjug, line 87:
```
  File "/home/user/go/src/github.com/cloud-bulldozer/benchmark-operator/venv/lib/python3.9/site-packages/ripsaw/commands/operator.py", line 87, in _perform_operator_action
    _clone_repo(repo, branch, temp_dir.name)
AttributeError: 'str' object has no attribute 'name'
```

cli/ripsaw/commands/operator.py
whitleykeith: the only real way to do that is to add a flag to force using a remote repo, which I just added as `--force-remote`. This has the added benefit of being able to force using the remote repo with the default repo/branch

cli/ripsaw/commands/operator.py
whitleykeith: updated it to make it look cleaner and also refactored it a bit

cli/ripsaw/commands/operator.py
whitleykeith: fixed

PR #620, opened 2021-07-29
Arbitrary env vars to kube-burner

rsevilla87:
### Description

Mechanism to inject arbitrary env vars to the kube-burner benchmark.


dry923:
@rsevilla87 not sure what happened here but how can we make this fail faster and not eat 2 hours of CI time?

PR #623, opened 2021-07-30
Lower reconcilePeriod

rsevilla87:
### Description
We need to set reconcilePeriod as low as possible. As there are reconcile actions that does not modify the status of the Benchmark CR (modifying the Benchmark object triggers a reconcile),  (I observed this behaviour in the uperf benchmark), the benchmark-operator could wait up to 1 minute to perform the next actions of that Benchmark.

With this change we should dramatically drop the benchmark duration.

### DIFF
--- a/watches.yaml
+++ b/watches.yaml
@@ -4,7 +4,7 @@
   group: ripsaw.cloudbulldozer.io
   kind: Benchmark
   playbook: ./playbooks/benchmark.yml
-  reconcilePeriod: 1m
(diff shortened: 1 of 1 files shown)


dry923:
@rsevilla87 looks like this was introduced in a recent PR (likely the version upgrade). It used to be 3 seconds. I propose we go back to that number.

whitleykeith (approved):
lgtm

dry923 (approved):
LGTM!

PR #624, opened 2021-08-02
Fix system-metrics collection

rsevilla87:
### Description

The current implementation of system-metrics was broken due to recent changes in kube-burner's metrics-collection config structure.


sjug (approved):
/lgtm

PR #626, opened 2021-08-04
Use IMG to push the image

rsevilla87:
### Description

We shouldn't append the architecture when IMG is set.

With the previous implementation, executing `$ make image-build IMG=quay.io/bla/blabla:latest`

Ended up with the image `quay.io/bla/blabla:latest-amd64` being built and pushed, which is incorrect, we shouldn't append the architecture suffix when IMG is set.


sjug (approved):

whitleykeith (approved):
lgtm

PR #627, opened 2021-08-04
fio prefill should use same numjobs value as fio samples

bengland2:
move numjobs loop up above prefill section
pass numjobs var into prefill fio job file
This fixes problem where fio reported read throughput greater than the actual block device throughput that fio was reading.   The root cause was that the files were not completely populated so that reads to the unwritten portions of the file return immediately with zeroes, resulting in higher throughput from the files than the block device.

### DIFF
--- a/roles/fio_distributed/templates/configmap.yml.j2
+++ b/roles/fio_distributed/templates/configmap.yml.j2
@@ -5,6 +5,8 @@ metadata:
   name: fio-test-{{ trunc_uuid }}
   namespace: '{{ operator_namespace }}'
 data:
+# FIXME: I don't think prefill works correctly for a list of numjobs values, only for 1 of them
+{% for numjobs in workload_args.numjobs %}
 {% if workload_args.prefill is defined and workload_args.prefill is sameas true %}
   fiojob-prefill: |
     [global]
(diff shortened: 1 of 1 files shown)


bengland2:
Using the program [analyze-fio-test-results.py](http://nfs-storage01.scalelab.redhat.com/bengland/public/openshift/ripsaw/analyze-fio-test-results.py) I queried elasticsearch and analyzed [results of fio benchmark run here](http://nfs-storage01.scalelab.redhat.com/bengland/public/openshift/ripsaw/fio-blkdev/fio-numjobs-20-fix.log), with the [CR here](http://nfs-storage01.scalelab.redhat.com/bengland/public/openshift/ripsaw/fio-blkdev/numjobs-20.yaml) 
numjobs-20-fixed-2021-08-04-12-55-41, and you can see that fio reports exactly what the block device throughput is reported to be below, 3K IOPS for a 1-TiB AWS gp2 PV

bengland2:
waiting for PR 628 to merge before re-running CI

dry923 (approved):
LGTM!

rsevilla87 (approved):
LGTM!

PR #634, opened 2021-08-09
Remove IMG_ARCH references

rsevilla87:

### Description

Make the variable IMAGE_ARCH optional. By default the containers built, pushed and deployed don't not have the the -$(ARCH) suffix. Passing IMAGE_ARCH to Makefile adds this suffix

The release workflow has been updated to pass this variable.


dry923 (approved):
LGTM!

PR #638, opened 2021-08-11
Benchmark status handler

rsevilla87:
### Description

Create two new task files in the benchmark_state role to be used as boilerplate to establish the statuses _Building, Running Completed/Failed_ in the Benchmarks

Thanks to the addition of these "functions", we can delete hundreds repeated lines across the different benchmarks available in benchmark-operator as well as establishing a basis to set the status of the benchmarks correctly.

Working benchmarks:

- [x] backpack
- [x] byowl
- [x] cyclictest
- [x] fio_distributed
- [x] flent
- [x] fs-drift
- [x] hammerdb
- [x] image_pull
- [x] iperf3
- [x] kube-burner
- [x] log_generator
- [x] oslat
- [x] pgbench
- [x] scale_openshift
- [x] servicemesh
- [x] smallfile
- [x] stressng
- [x] sysbench
- [x] testpmd
- [x] uperf
- [x] vegeta
- [x] ycsb

Required updates:
- Adding the label `benchmark-uuid` to all Jobs deployed by benchmark-operator, since the ansible code uses this label to look for completed jobs.
- Turn "server" Jobs such as iperf3 and fio into Pods. (We want Jobs to be used as ephemeral/client pods)
- Added a timeout to all jobs with `activeDeadlineSeconds: {{ workload_args.job_timeout|default(3600) }}`.
- Jobs should not be restared when the fail. `backoffLimit: 0`


learnitall (commented):
This is awesome! I added a few comments on adding some additional task files think I think could reduce even more lines of boilerplate code.

I also think that after we make more progress with the signal work for benchmark-wrapper (https://github.com/cloud-bulldozer/benchmark-wrapper/pull/305), it would be cool to revisit this and adjust how we set benchmark-state at the redis level too. That's way down the line though for sure.

roles/benchmark_state/tasks/completed.yml
```diff
@@ -0,0 +1,37 @@
...
+      api_version: ripsaw.cloudbulldozer.io/v1alpha1
+      kind: Benchmark
+      name: "{{ ansible_operator_meta.name }}"
+      namespace: "{{ operator_namespace }}"
+      status:
+        state: Failed
+        complete: true
+    when: job_state|json_query('resources[].status[]')|selectattr('failed', 'defined')|list|length > 0
+
+  when: resource_state.resources[0].status.state == "Running"
```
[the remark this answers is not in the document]
rsevilla87: yeah, that's right, after doing some tests I'm gonna move the task below to the playbook pre_tasks section.

```ansible
- name: Get current state
  k8s_info:
    api_version: ripsaw.cloudbulldozer.io/v1alpha1
    kind: Benchmark
    name: '{{ ansible_operator_meta.name }}'
  register: resource_state
```

roles/byowl/tasks/main.yml
```diff
@@ -43,27 +28,9 @@
         state: Running
```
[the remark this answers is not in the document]
whitleykeith: +1 to this, I think the reasoning is that not every task may use the "Running' state at the moment (or have more steps)

For instance uperf starts servers and clients so it has a state for "Starting Servers" and "Starting Clients". That being said I think we should have some universal status flag for Running

whitleykeith (approved):
lgtm, shared a small nit @learnitall had but i don't think it's something we need to block this PR on so here's the :heavy_check_mark:

bengland2:
weird that smallfile required 2 attempts to pass, why?  is it not reporting benchmark state correctly?

roles/byowl/tasks/main.yml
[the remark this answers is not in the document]
rsevilla87: done, I've modified those tasks to a more generic approach, we now can use:

```yaml
- include_role: benchmark_state
  tasks_from: set_state
  vars:
    state: Running
```

rsevilla87:
> weird that smallfile required 2 attempts to pass, why? is it not reporting benchmark state correctly?

It seems like it sometimes times out because smallfile CI runs in parallel with other I/O intensive workloads.

mohit-sheth (approved):
LGTM, awesome work @rsevilla87

playbooks/templates/metadata_pod.yml.j2
```diff
@@ -0,0 +1,43 @@
+{% if metadata.collection is sameas true and metadata.targeted is sameas true %}
```
dry923, line 1: @rsevilla87 how is this any different from playbooks/templates/metadata.yml.j2 ? A side by side looks the same at a glance.

rsevilla87: indentation ;)

dry923: Ahhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhh. Gothca thanks :-)

roles/iperf3/tasks/main.yml
```diff
@@ -1,75 +1,43 @@
...
-        state: "Starting Clients"
+  - include_role:
+      name: benchmark_state
+      tasks_from: set_state
+    vars:
+      state: "Starting Clients"
     when: "workload_args.pairs|default('1')|int == (server_pods | json_query('resources[].status.podIP')|length) and (server_pods | json_query('resources[].status.phase') is defined and workload_args.pairs|default('1')|int == server_pods | json_query('resources[].status[]')|selectattr('phase','match','Running')|list|length)"
 
-  when: resource_state.resources[0].status.state == "Starting Server"
+  when: benchmark_state.resources[0].status.state == "Starting Server"
```
sjug, line 35: Unrelated again and minor, the other statuses are `Starting Servers` (plural) regardless of count of servers started. `iperf3` is the only one that's singular.

sjug (commented):
Looks good @rsevilla87!

playbooks/benchmark.yml
```diff
@@ -82,31 +82,39 @@
...
       name: system-metrics
     
     vars:
       workload_args: "{{ workload.args }}"
     when:
-    - cr_state.resources[0].status.state is defined
-    - cr_state.resources[0].status.state == "Complete"
+    - benchmark_state.resources[0].status.state is defined
+    - benchmark_state.resources[0].status.state == "Complete"
     - system_metrics.collection | bool
```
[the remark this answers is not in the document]
rsevilla87: That's a different PR ;)

roles/iperf3/tasks/main.yml
rsevilla87: You're right, I just tried to keep the current semantics here, the major changes are that all benchmarks set now a Completed or Failed state

sjug (approved):
/lgtm

Issue #642, opened 2021-08-16
iperf3 Benchmark fails: Running benchmark of iperf3 for UDP doesn't work and Pods go in error state.  


MuhammadMunir12:
I have run the iperf3 benchmark with default configs as mentioned in the benchmark manifest given below: 

```
apiVersion: ripsaw.cloudbulldozer.io/v1alpha1
kind: Benchmark
metadata:
  name: iperf3-benchmark
  namespace: my-ripsaw
spec:
  clustername: ocp-cluster
  workload:
    name: iperf3
    args:
      pairs: 1
      hostnetwork: false
      pin: true
      pin_server: "worker-1"
      pin_client: "worker-1"
      port: 5201
      transmit_type: time
      transmit_value: 60
      omit_start: 0
      length_buffer: 128K
      window_size: 64k
      ip_tos: 0
      mss: 900
      streams: 1
      extra_options_client: ' '
      extra_options_server: ' '
      #retries: 200
```

The above manifest works fine. As mentioned here: [https://github.com/cloud-bulldozer/benchmark-operator/blob/master/docs/iperf.md]
When I add extra options for the client & server for UDP as "-u", both server and client pods go in an error state. Whereas, upon using "u" flag I see no difference in default (TCP) results and results with "u" flag for UDP. 

```
apiVersion: ripsaw.cloudbulldozer.io/v1alpha1
kind: Benchmark
metadata:
  name: iperf3-benchmark
  namespace: my-ripsaw
spec:
  clustername: r189a
  workload:
    name: iperf3
    args:
      pairs: 1
      hostnetwork: false
      pin: true
      pin_server: "worker-1"
      pin_client: "worker-1"
      port: 5201
      transmit_type: time
      transmit_value: 60
      omit_start: 0
      length_buffer: 128K
      window_size: 64k
      ip_tos: 0
      mss: 900
      streams: 1
      extra_options_client: "-u"   #or "u" 
      extra_options_server: "-u"  #or "u"
      #retries: 200

```

Could you please guide me on using the UDP-based traffic patterns iperf3 benchmark?

Also, I was not able to run iperf3 benchmark on nodes in MCP for DPDK. Is there any restriction that iperf3 benchmark could only be run on nodes with no DPDK enabled?

jtaleric:
> https://github.com/cloud-bulldozer/benchmark-operator/blob/master/docs/iperf.md



[...]

PR #616, opened 2021-07-28
Fix broken links in the docs

Harshith-umesh:
### Description
Fix broken links in majority of the workload docs
### Reviewers
@whitleykeith

PR #643, opened 2021-08-16
Small CR iperf3 CR change & docs update

sjug:
### Description
Some of the links in the docs were fixed previously by https://github.com/cloud-bulldozer/benchmark-operator/pull/616, however, there was a few misses.
Some of the docs were inconsistent in their sample sections.
Discovered while taking a look at https://github.com/cloud-bulldozer/benchmark-operator/issues/642.

### Fixes
- Small change to comment out length of buffer in iperf3 test, as 128K is the default for TCP but hardcoding the value breaks UDP.
- Minor corresponding changes to docs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,7 @@ ARCH ?= $(shell uname -m | sed 's/x86_64/amd64/')
 # Image URL to use all building/pushing image targets
 REGISTRY ?= quay.io
 ORG ?= cloud-bulldozer
-# Get the current branch/tag name
+# Get the current branch name
 # In case this is the master branch, rename it to latest
 VERSION ?= $(shell git describe --tags 2>/dev/null || git rev-parse --abbrev-ref HEAD | sed 's/master/latest/g')
 IMG ?= $(REGISTRY)/$(ORG)/benchmark-operator:$(VERSION)
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -5,6 +5,12 @@ generatorOptions:
   disableNameSuffixHash: true
 
 configMapGenerator:
-- name: manager-config
-  files:
+- files:
   - controller_manager_config.yaml
+  name: manager-config
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+images:
+- name: controller
+  newName: quay.io/cloud-bulldozer/benchmark-operator
+  newTag: latest
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -71,7 +71,7 @@ Example `playbook.yml`:
 
 ### Workload triggers
 [CRD](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/) holds the definition of the resource.
-The operator triggers roles based on the conditions defined in a CR ([example](resources/crds/ripsaw_v1alpha1_uperf_cr.yaml)) which will influence which roles the
+The operator triggers roles based on the conditions defined in a CR ([example](config/samples/uperf/cr.yaml) which will influence which roles the
 [playbook](playbook.yml) executes.
 Other vars may be defined that can modify the workload run conditions.
 
@@ -154,7 +154,7 @@ Redefine CRD
 ```
 Apply a new CR
 ```bash
-# kubectl apply -f resources/crds/ripsaw_v1alpha1_uperf_cr.yaml
+# kubectl apply -f config/samples/uperf/cr.yaml
 ```
 
 ## CI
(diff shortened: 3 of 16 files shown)


jtaleric (approved):
LGTM Thanks for reviewing and fixing the issues!

rsevilla87:
Hey @sjug , this PR will break the release workflow https://github.com/cloud-bulldozer/benchmark-operator/blob/master/.github/workflows/release.yml. 
The current implementation sets VERSION to the current tag or branch name, this workflow builds container images when new code is commited in the master branch or a new tag is pushed to the repo.

In case you're working on a new branch and you want to deploy benchmark-operator you can set IMG and point it to the desired image, otherwise it will try to deploy quay.io/cloub-bulldozer/benchmark-operator:<branch-name>.

```
IMG=quay.io/cloub-bulldozer/benchmark-operator:latest make deploy
```

config/manager/kustomization.yaml
```diff
@@ -5,6 +5,12 @@ generatorOptions:
   disableNameSuffixHash: true
 
 configMapGenerator:
-- name: manager-config
-  files:
+- files:
```
rsevilla87, line 8: I think you commited this file by mistake, as this file is modified by kustomize when you deploy the operator

sjug: No mistake, I'm aware that it's modified by kustomize when the operator is deployed. However as the code stands this is what would be in the file.

If changes are made this file will still be modified. However if one were to just checkout the code and `make deploy` this file would not be modified. I feel that's a reasonable expectation if one doesn't modify the code otherwise.

sjug:
> Hey @sjug , this PR will break the release workflow https://github.com/cloud-bulldozer/benchmark-operator/blob/master/.github/workflows/release.yml.
> The current implementation sets VERSION to the current tag or branch name, this workflow builds container images when new code is commited in the master branch or a new tag is pushed to the repo.
> 
> In case you're working on a new branch and you want to deploy benchmark-operator you can set IMG and point it to the desired image, otherwise it will try to deploy quay.io/cloub-bulldozer/benchmark-operator:.
> 
> ```
> IMG=quay.io/cloub-bulldozer/benchmark-operator:latest make deploy
> ```

Interesting, thanks for the heads up. I figured it was used for something, just not sure what.
I guess the issue is essentially that this `VERSION` is tailored only for CI.

The `master` branch in this original repo does not have tags so therefore it for some reason becomes `latest`.
However on any forks, the master branch does have a tag, and therefore you get something like: 
```
quay.io/cloud-bulldozer/benchmark-operator:v0.1-31-g2d3dc8d
```

I feel like it makes more sense for CI to set the version explicitly rather than requiring the user to do so. What do you think?

Makefile
```diff
@@ -40,9 +40,9 @@ ARCH ?= $(shell uname -m | sed 's/x86_64/amd64/')
 # Image URL to use all building/pushing image targets
 REGISTRY ?= quay.io
 ORG ?= cloud-bulldozer
-# Get the current branch/tag name
+# Get the current branch name
 # In case this is the master branch, rename it to latest
-VERSION ?= $(shell git describe --tags 2>/dev/null || git rev-parse --abbrev-ref HEAD | sed 's/master/latest/g')
+VERSION ?= $(shell git rev-parse --abbrev-ref HEAD | sed 's/master/latest/g')
```
rsevilla87, line 45: I don't understand this change. The original code was able to set VERSION to something like v1.1 if you are in the the tag v.1.1 or failback to set VERSION to the current branch name (in case of master this is updated to latest).

You have drop detecting the tag name in your implementation, e.g.:

Output using the original implementation:
```shell
rsevilla@wonderland ~/labs/ripsaw ((v0.1)) $ git --version
git version 2.31.1
rsevilla@wonderland ~/labs/ripsaw ((v0.1)) $ git describe --tags 2>/dev/null || git rev-parse --abbrev-ref HEAD | sed 's/master/latest/g'
v0.1
rsevilla@wonderland ~/labs/ripsaw (master) $ git checkout e2e-testing
Switched to branch 'e2e-testing'
rsevilla@wonderland ~/labs/ripsaw (e2e-testing) $ git describe --tags 2>/dev/null || git rev-parse --abbrev-ref HEAD | sed 's/master/latest/g'
e2e-testing
```

sjug: Yes indeed, I too can understand the bash involved. Let me rephrase my previous message; the issue was/is that the "original code" would resolve to a tag even on an untagged branch such as `master`.
```
$ git branch
* master
$ git describe --tags
v0.1-31-g2d3dc8d
```

This isn't the case on other forks, so it must be an issue on just mine. I'll revert the change.

config/manager/kustomization.yaml
rsevilla87: Ok, seems to work, I thought this file wouldn't be modified once kustomize had changed it. :)

rsevilla87 (approved):
LGTM!

sjug:
The way we set `VERSION` seems to be [breaking the prow CI jobs](https://prow.ci.openshift.org/job-history/gs/origin-ci-test/pr-logs/directory/pull-ci-openshift-kubernetes-release-4.8-configmap-scale).

PR #647, opened 2021-08-18
make undeploy should depend on make kustomize

RobertKrawitz:
### Description
make undeploy needs kustomize; the undeploy target should depend on it.


whitleykeith (approved):
lgtm

PR #650, opened 2021-08-23
update cache dropping doc

bengland2:
get rid of a now-unnecessary step
explain better how to set labels for kernel cache dropper


bengland2:
can someone please merge this?   I can't do it.

Issue #649, opened 2021-08-23
Unable to install benchmark-operator using default method and helm chart for ppc64le


piyushgupta1551:
Hi Team,

I am unable to install the benchmark-operator using both provided methods on ppc64le clusters. See issues below - 

**- Default method provided -** 
```
The easiest way to install the operator is through the operator-sdk methods provided in the Makefile.

git clone https://github.com/cloud-bulldozer/benchmark-operator
cd benchmark-operator
make deploy
```

What I get - 

```
[root@rdr-ap-49-mon01-bastion-0 benchmark-operator]# make deploy

gzip: stdin: not in gzip format
tar: Child returned status 1
tar: Error is not recoverable: exiting now
make: *** [Makefile:120: kustomize] Error 2
```
Why I get this error - There is no kustomize release tar available upstream for ppc64le https://github.com/cloud-bulldozer/benchmark-operator/blob/ef4b6638c806969db99f1dae4cfc9578fc7db3c5/Makefile#L123

**- Helm -** 
Default instructions provided - 
```
Installing the benchmark-operator is easiest by using the helm chart and can be done with the following commands. This requires your machine to have Helm installed. Install Helm

Note: If running on openshift you'll need to run this command before installing the chart. oc adm policy -n benchmark-operator add-scc-to-user privileged -z benchmark-operator

git clone https://github.com/cloud-bulldozer/benchmark-operator
cd benchmark-operator/charts/benchmark-operator
helm install benchmark-operator . -n benchmark-operator --create-namespace
```

What I get - 
```
[root@rdr-ap-49-mon01-bastion-0 benchmark-operator]# oc adm policy -n benchmark-operator add-scc-to-user privileged -z benchmark-operator
clusterrole.rbac.authorization.k8s.io/system:openshift:scc:privileged added: "benchmark-operator"
[root@rdr-ap-49-mon01-bastion-0 benchmark-operator]# helm install benchmark-operator . -n benchmark-operator --create-namespace
WARNING: Kubernetes configuration file is group-readable. This is insecure. Location: /root/.kube/config
WARNING: Kubernetes configuration file is world-readable. This is insecure. Location: /root/.kube/config
walk.go:74: found symbolic link in path: /home/piyush/new/benchmark-operator/charts/benchmark-operator/crds/crd.yaml resolves to /home/piyush/new/benchmark-operator/config/crd/bases/ripsaw.cloudbulldozer.io_benchmarks.yaml
[...]

PR #651, opened 2021-08-23
Chart fix

whitleykeith:
### Fixes

* https://github.com/cloud-bulldozer/benchmark-operator/issues/649


sjug (approved):
Good diff & catch with all manager updates.

/lgtm

PR #664, opened 2021-08-31
Set restartPolicy Never to all jobs deployed by benchmark-operator

rsevilla87:

### Description
Our jobs must be one shot, and we should keep failed or completed pods to ease debugging.

Restart policy must be set to Never to prevent garbage collecting failed pods, more info at https://github.com/kubernetes/kubernetes/issues/74848#issuecomment-475178355


sjug (approved):
/lgtm

PR #665, opened 2021-08-31
Elasticsearch URL

rsevilla87:

### Description

With this change, elasticsearch indexing is enabled when the elasticsearch.url parameter is filled with something different from ""

### Fixes

The current implementation is not possible to disable ES indexing since the elasticsearch dictionary defaults to {}

### DIFF
--- a/config/crd/bases/ripsaw.cloudbulldozer.io_benchmarks.yaml
+++ b/config/crd/bases/ripsaw.cloudbulldozer.io_benchmarks.yaml
@@ -29,8 +29,10 @@ spec:
                 properties:
                   url:
                     type: string
+                    default: ""
                   index_name:
                     type: string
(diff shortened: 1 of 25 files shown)


dry923 (approved):
LGTM!

sjug:
Why is it not possible to use the current implementation?

```
>>> bool({})
False
>>> bool("")
False
```

rsevilla87:
> Why is it not possible to use the current implementation?
> 
> ```
> >>> bool({})
> False
> >>> bool("")
> False
> ```

You're right, however the current implementation checks if the elasticsearch dictionary is defined, (that condition always happens), in addition that dictionary has some nested fields with default values too, which will make it always "True".

sjug (approved):
/lgtm

PR #668, opened 2021-09-01
Adds initial config for api-load tests

chentex:

### Description

Adds initial run of api-load tests in a container.


rsevilla87 (approved):
As discussed in slack with @chentex, this is the initial implementation and further changes will be required in the future to implement other features such as ES indexing. LGTM!

PR #669, opened 2021-09-01
Removing default value for image_name as it was causing the real defa…

dry923:
…ults defined in the yaml template not to be used

### Description
Recently merged PR https://github.com/cloud-bulldozer/benchmark-operator/pull/665 set the index_name default to "" for Null. This is causing the defaults in the yaml template of the workload to no longer be honored.


learnitall (approved):
Can you make this change smaller? Really hard to review.

learnitall:
lgtm!

PR #673, opened 2021-09-02
Add somaxconn parameter to vegeta tests

e-minguez:
### Description

The net.core.somaxconn sysctl is 128 by default (4096 in newer kernels https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?id=19f92a030ca6d772ab44b22ee6a01378a8cb32d4)

You can specify sysctls at pod level as https://docs.openshift.com/container-platform/4.7/nodes/containers/nodes-containers-sysctls.html#nodes-containers-sysctls-setting_nodes-containers-using 

### Fixes
N/A, new feature

### DIFF
--- a/config/samples/vegeta/cr.yaml
+++ b/config/samples/vegeta/cr.yaml
@@ -13,6 +13,7 @@ spec:
     name: vegeta
     args:
       # node_selector: "vegeta=true"
+      # net_core_somaxconn: 4096
       hostnetwork: false
       clients: 2
       targets:
--- a/docs/vegeta.md
+++ b/docs/vegeta.md
@@ -14,6 +14,8 @@ intended for Kata containers.
 The **node_selector** option can be used to limit the nodes where
 the vegeta pods are deployed.
 
+Use **net_core_somaxconn** to tweak the maximum number of "backlogged sockets" (`net.core.somaxconn`).  Default is 128.
+
 ```yaml
 apiVersion: ripsaw.cloudbulldozer.io/v1alpha1
 kind: Benchmark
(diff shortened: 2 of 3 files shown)


e-minguez:
In fact, I've just seen that particular sysctl is considered unsafe https://kubernetes.io/docs/tasks/administer-cluster/sysctl-cluster/#setting-sysctls-for-a-pod

Not sure about this PR then... thoughts?

rsevilla87:
Hey @e-minguez , several things wrt this PR.

- The parameter  net.core.somaxconn is indeed considered unsafe, so it must be whitelisted by the cluster kubelet's in order to apply it at pod level.
-  This parameter won't take any effect on vegeta's behaviour, since it's referent to to the sever-side of the connection.

In case you want to _improve_  the number of sockets the vegeta pods are able to open I suggest you to uset hostNetwork based pods, and apply some kernel tuning using tuned-operator:

```yaml
apiVersion: tuned.openshift.io/v1
kind: Tuned
metadata:
  name: openshift-ingress-performance
  namespace: openshift-cluster-node-tuning-operator
spec:
  profile:
  - data: |
      [main]
      summary=Network performance profile
      include=openshift
      [sysctl]
      net.ipv4.ip_local_port_range="1024 65535"
      net.ipv4.tcp_tw_reuse=1
    name: openshift-network-performance
  recommend:
  - match:
    - label: "node-role.kubernetes.io/worker"
    priority: 10
    profile: openshift-network-performance
```

e-minguez:
Gotcha. I guess it would be nice to document those tweaks :)

PR #679, opened 2021-09-08
Update CONTRIBUTING.md

jeniferh:
Add changes since v0.1 to update the operator image

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -121,41 +121,15 @@ case of failure or when disabled. This ensures no interference with subsequent w
 
 ### The operator container image
 Any changes to the [roles](roles/) tree or to the [playbook](playbook.yml) file will necessitate a new build of the operator container image.
-The benchmark-operator container can be built using [podman](https://podman.io/) and pushed to a public repository.
-The public repository could be [quay](https://quay.io) in which case you'll need to:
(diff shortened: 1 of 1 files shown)


jtaleric (approved):
LGTM, not sure of a more "proper way" either.

CONTRIBUTING.md
```diff
@@ -124,6 +124,7 @@ Any changes to the [roles](roles/) tree or to the [playbook](playbook.yml) file
 The benchmark-operator container can be built using [podman](https://podman.io/) and pushed to a public repository.
 The public repository could be [quay](https://quay.io) in which case you'll need to:
 
+#### v0.1
```
rsevilla87, line 127: Hey @jeniferh, the docs for the v0.1 tag are correct, I think we shouldn't include the v0.1 section in the master branch,

I'd just move the comments from the "v0.1" you've created here to the "The operator container image" section

rsevilla87 (changes requested):

CONTRIBUTING.md
jeniferh: Good point, I will adjust so the master version only documents the current method since the v0.1 tag covers the old method.

rsevilla87 (approved):
LGTM!

sjug (approved):
/lgtm

PR #684, opened 2021-09-13
add rosa parameters to scale template

morenod:
Add rosa parameters on snafu execution if they are detected as workload arguments

Adding also aws parameters also as workload arguments as they are required by rosa to run scaling process


rsevilla87 (approved):
LGTM!

dry923 (approved):
LGTM!

PR #685, opened 2021-09-13
Fio server_vm image updates

### LINKED ISSUE #527: fio with kind: vm no longer works
the fio benchmark with kind: vm will not work anymore.  This is because the fio versions of the client and server must match EXACTLY, but the fio benchmark of the client is fio-3.19-3 whereas the fio benchmark of the server container image was frozen in time at quay.io/mulbc/fed-fio , so you'd see errors like this in the client log:

```
fio: client/server version mismatch (84 != 82)
fio: bad server cmd version 84
```

 I fixed this by generating my own fio image using Chris Blum's handy script and modifying it slightly.   The version I'm using right now is at 

```
#!/bin/sh

IMAGE_URL=https://cloud.centos.org/centos/8/x86_64/images/CentOS-8-GenericCloud-8.3.2011-20201204.2.x86_64.qcow2
IMAGE=centos8.qcow2
LOCAL_IMAGE=mycentos8.qcow2
QUAY_IMAGE=quay.io/bengland2/centos8-fio:latest

set -x
set -e

sudo dnf install podman podman-docker wget -y

wget --continue $IMAGE_URL -O $IMAGE
cp -f $IMAGE $LOCAL_IMAGE

if ! virt-customize --help >/dev/null; then
  sudo yum install -y virt-customize || sudo yum install -y libguestfs-tools
fi

virt-customize -a $LOCAL_IMAGE \
  --install wget,libaio,curl,python3,python3-pip \
  --run-command "wget https://raw.githubusercontent.com/cloud-bulldozer/bohica/master/stockpile-wrapper/stockpile-wrapper.py && pip3
 install elasticsearch-dsl openshift kubernetes redis" \
  --copy-in fio-3.19-centos8:/usr/local/bin \
  --selinux-relabel \
  --root-password password:yourPassword \
  --firstboot-command '/usr/local/bin/fio-3.19-centos8 --server'
virt-sysprep -a $LOCAL_IMAGE

cat <<END >Dockerfile
FROM kubevirt/container-disk-v1alpha:v0.13.7
ADD $LOCAL_IMAGE /disk/
END
docker rmi $QUAY_IMAGE || echo "image already removed"
docker build -t $QUAY_IMAGE .
docker push $QUAY_IMAGE
```

And I built the fio image by logging into one of the VMs :

```
[kni@f20-h25-000-r640 benchmark-operator]$ oc get vmis
NAME                    AGE     PHASE     IP             NODENAME
fio-server-1-e3abc763   2m14s   Running   10.131.0.171   f21-h02-000-r640.rdu2.scalelab.redhat.com
...
[kni@f20-h25-000-r640 benchmark-operator]$ virtctl console fio-server-3-e3abc763
Successfully connected to fio-server-3-e3abc763 console. The escape sequence is ^]

CentOS Linux 8
Kernel 4.18.0-240.1.1.el8_3.x86_64 on an x86_64
...
fio-server-3-e3abc763 login: root
Password: 
[root@fio-server-3-e3abc763 ~]# 
```

and doing this:

```
and then did the following to build fio and install it:

 dnf install -y git gcc make libaio-devel zlib-devel
 git clone https://github.com/axboe/fio
 git checkout fio-3.19
 cd fio
 git checkout fio-3.19
./configure --disable=librados --disable=librbd
make -j
 ./fio --version
fio-3.19
```

and [saved the fio program here](http://perf1.perf.lab.eng.bos.redhat.com/pub/bengland/tmp/ms-etcd/fio-3.19-centos8).   

Eventually someone should make a PR and incorporate all this into the fio benchmark, but for now this is a usable workaround, I think.

jeniferh:
### Description
Adjusting server_vm config so that any VM containerDisk image can be used since the workload is run from chroot into a fio container image (i.e. no custom VM images need to be maintained).

Also adding a fallback VM option if no PVC/HostDisk is specified, use an emptyDisk (ephemeral storage).
Adding nodeSelector support for VMs.

### Fixes
#527

### DIFF
--- a/roles/fio_distributed/templates/server_vm.yml.j2
+++ b/roles/fio_distributed/templates/server_vm.yml.j2
@@ -22,13 +22,22 @@ spec:
       cores: {{ workload_args.vm_cores | default(1) }}
     devices:
       disks:
-        - name: registrydisk
-        - name: cloudinitdisk
+        - disk:
+            bus: {{ workload_args.vm_bus | default('virtio') }}
+          name: registrydisk
+        - disk:
+            bus: {{ workload_args.vm_bus | default('virtio') }}
+          name: cloudinitdisk
 {% if workload_args.storageclass is defined
     or workload_args.hostpath is defined %}
         - disk:
-            bus: virtio
+            bus: {{ workload_args.vm_bus | default('virtio') }}
           name: data-volume
+          serial: data
+{% else %}
+        - disk:
+            bus: {{ workload_args.vm_bus | default('virtio') }}
+          name: emptydisk
           serial: data
 {% endif %}
     resources:
@@ -37,37 +46,23 @@ spec:
   volumes:
     - name: registrydisk
       containerDisk:
-        image: {{ workload_args.vm_image | default('kubevirt/fedora-cloud-container-disk-demo:latest') }}
+        image: {{ workload_args.vm_image | default('quay.io/kubevirt/fedora-container-disk-images:latest') }}
     - name: cloudinitdisk
       cloudInitNoCloud:
         userData: |-
           #cloud-config
-          password: fedora
+          password: ripsaw
           chpasswd: { expire: False }
           bootcmd:
-{% if workload_args.storageclass is defined
-    or workload_args.hostpath is defined %}
             - "mkdir -p {{ fio_path }} || true"
-            - mkfs.ext4 /dev/disk/by-id/virtio-data
-            - "mount /dev/disk/by-id/virtio-data {{ fio_path }}"
-{% endif %}
+            - "[ -e /dev/disk/by-id/*data ] && disk=$(shopt -s nullglob; basename /dev/disk/by-id/*data) && mkfs.ext4 /dev/disk/by-id/$disk && mount /dev/disk/by-id/$disk {{ fio_path }}"
           runcmd:
-{% if workload_args.vm_image is undefined %}
-            - dnf install -y fio wget curl python3 python3-pip
-{% endif %}
-            - cd /tmp
-{% if workload_args.metadata is defined
-    and (workload_args.metadata.collection|default(false))
-    and (workload_args.metadata.targeted|default(true)) %}
-            - export ES_SERVER={{ elasticsearch.url }}
-{% if workload_args.vm_image is undefined %}
-            - wget https://raw.githubusercontent.com/cloud-bulldozer/bohica/master/stockpile-wrapper/stockpile-wrapper.py
-            - pip3 install elasticsearch-dsl openshift kubernetes redis
-{% endif %}
-            # TODO: start stockpile-wrapper script
-            python3 stockpile-wrapper.py -s {{ elasticsearch.url }} -u {{ uuid }} -n "$my_node_name" -N "$my_pod_name" --redisip {{ bo.resources[0].status.podIP }} --redisport 6379 --force
-{% endif %}
-            - fio --server
+            - dnf install -y podman
+            - "img=`podman create {{ workload_args.image | default('quay.io/cloud-bulldozer/fio:latest') }}`"
+            - fs=`podman mount $img`
+            - "mkdir -p $fs/{{ fio_path }} || true"
+            - "mount -o bind /{{ fio_path }} $fs/{{ fio_path }}"
+            - "chroot $fs bash -c 'cd /{{ fio_path }}; fio --server'"
 {% if workload_args.storageclass is defined %}
     - name: data-volume
       persistentVolumeClaim:
@@ -75,7 +70,17 @@ spec:
 {% elif workload_args.hostpath is defined %}
     - name: data-volume
       hostDisk:
-        path: {{ workload_args.hostpath }}
+        path: "{{ workload_args.hostpath }}/fio-server-{{item | string}}-{{ trunc_uuid }}"
         capacity: {{ workload_args.storagesize | default("5Gi") }}
         type: DiskOrCreate
+{% else %}
+    - name: emptydisk
+      emptyDisk:
+        capacity: {{ workload_args.storagesize | default("5Gi") }}
 {% endif %}
+{% if workload_args.nodeselector is defined %}
+  nodeSelector:
+{% for label, value in  workload_args.nodeselector.items() %}
+    {{ label | replace ("_", "-" )}}: {{ value }}
+{% endfor %}
+{% endif  %}
--- a/config/samples/fio/vm-cr.yaml
+++ b/config/samples/fio/vm-cr.yaml
@@ -33,7 +33,10 @@ spec:
       # vm_cores: 2
       # Memory that will be available inside the VM
       # Default: 5G
-      # vm_cores: 10G
+      # vm_memory: 10G
+      # VM bus type: virtio, sata, or scsi
+      # Default: virtio
+      # vm_bus: virtio
       # test types, see fio documentation
       jobs:
         - write
--- a/docs/fio_distributed.md
+++ b/docs/fio_distributed.md
@@ -108,12 +108,13 @@ The workload loops are nested as such from the CR options:
   > the specified annotations on the client pods' metadata.
 - **kind**: Can either be `pod` or `vm` to determine if the fio workload is run in a Pod or in a VM
   > Note: For VM workloads, you need to install Openshift Virtualization first
-- **vm_image**: Whether to use a pre-defined VM image with pre-installed requirements. Necessary for disconnected installs.
-  > Note: You can use my fedora image here: quay.io/mulbc/fed-fio \
-  > Note: Only applies when kind is set to `vm`
+- **vm_image**: VM image to use for the test, benchmark dependencies are provided by a pod container image (Default pod image: quay.io/cloud-bulldozer/fio:latest). Container images should be pre-loaded for disconnected installs. Default: quay.io/kubevirt/fedora-container-disk-images:latest
+  > Note: Only applies when kind is set to `vm`, current assumption is the firewall is disabled in the VM (default for most cloud images)
 - **vm_cores**: The number of CPU cores that will be available inside the VM. Default=1
   > Note: Only applies when kind is set to `vm`
-- **vm_cores**: The amount of Memory that will be available inside the VM in Kubernetes format. Default=5G
+- **vm_memory**: The amount of Memory that will be available inside the VM in Kubernetes format. Default=5G
+  > Note: Only applies when kind is set to `vm`
+- **vm_bus**: The VM bus type (virtio, scsi, or sata). Default=virtio
   > Note: Only applies when kind is set to `vm`
 - **bs**: (list) blocksize values to use for I/O transactions
   > Note: We set the `direct=1` fio option in the jobfile configmap. In order to avoid errors, the `bs` values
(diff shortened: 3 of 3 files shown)


jeniferh:
Still to-do: Fix stockpile part, adjust docs.

cc: @bengland2 @mulbc Feedback welcome!

bengland2:
Thank you @jeniferh !  I'm still reviewing how the code works, but 2 comments are: 

- it only supports a block-device storage class such as ocs-storageclass-ceph-rbd (or iscsi?), since you are making an ext4 filesystem on top of the PV, and this restriction should be documented in docs/fio_distributed.md (it probably was a restriction of the previous version also).   Shared filesystems like ocs-storageclass-cephfs or nfs would not work with this at present, but perhaps this approach could be made to work with them at a later date?

- Also, is there going to be any CI for this so that we know it works now and in the future?  Existing tests/test_crs/ won't cover it..  What could be done - write a tests/test_crs/fio_cnv_ocs.yaml that only gets run in tests/test_fiod.sh if CNV is installed and OCS is installed (easy to check).  I did that with tests/test_crs/valid_fiod_ocs_cache_drop.yaml .

roles/fio_distributed/templates/server_vm.yml.j2
```diff
@@ -13,52 +13,53 @@ spec:
       cores: {{ workload_args.vm_cores | default(1) }}
     devices:
       disks:
-        - name: registrydisk
-        - name: cloudinitdisk
+        - disk:
+            bus: virtio
```
bengland2, line 17: I agree this should be default but would it be useful to allow user to override this?  Does KVM or CNV support other forms of block device driver and would we ever need to test them?

roles/fio_distributed/templates/server_vm.yml.j2
```diff
@@ -13,52 +13,53 @@ spec:
...
     - name: registrydisk
       containerDisk:
-        image: {{ workload_args.vm_image | default('kubevirt/fedora-cloud-container-disk-demo:latest') }}
+        image: {{ workload_args.vm_image | default('quay.io/kubevirt/fedora-cloud-container-disk-demo:latest') }}
     - name: cloudinitdisk
       cloudInitNoCloud:
         userData: |-
           #cloud-config
-          password: fedora
+          password: ripsaw
```
bengland2, line 44: this may not be relevant to this PR but in general wouldn't it be easier to have the user supply a public key to enable login rather than a password?  These are so guessable.

roles/fio_distributed/templates/server_vm.yml.j2
```diff
@@ -13,52 +13,53 @@ spec:
...
-            # TODO: start stockpile-wrapper script
-            python3 stockpile-wrapper.py -s {{ elasticsearch.url }} -u {{ uuid }} -n "$my_node_name" -N "$my_pod_name" --redisip {{ bo.resources[0].status.podIP }} --redisport 6379 --force
-{% endif %}
-            - fio --server
+            - dnf install -y podman
+            - "img=`podman create {{ workload_args.image | default('quay.io/cloud-bulldozer/fio:latest') }}`"
+            - fs=`podman mount $img`
+            - "mkdir -p $fs/{{ fio_path }} || true"
+            - "mount -o bind /{{ fio_path }} $fs/{{ fio_path }}"
+            - "chroot $fs bash -c 'cd /{{ fio_path }}; fio --server'"
```
bengland2, line 62: where does port 8765 (fio --server port) get opened up?  Or is the CNV VM just not using a firewall?

bengland2 (approved):
the questions I raised are not reasons to stop this PR from going through.   We need this PR.

jeniferh:
Hey @bengland2, thanks for the review.
>     * it only supports a block-device storage class such as ocs-storageclass-ceph-rbd (or iscsi?), since you are making an ext4 filesystem on top of the PV, and this restriction should be documented in docs/fio_distributed.md (it probably was a restriction of the previous version also).   Shared filesystems like ocs-storageclass-cephfs or nfs would not work with this at present, but perhaps this approach could be made to work with them at a later date?

Ah good point! I will work on adding a FS-mode compatible option.
> 
>     * Also, is there going to be any CI for this so that we know it works now and in the future?  Existing tests/test_crs/ won't cover it..  What could be done - write a tests/test_crs/fio_cnv_ocs.yaml that only gets run in tests/test_fiod.sh if CNV is installed and OCS is installed (easy to check).  I did that with tests/test_crs/valid_fiod_ocs_cache_drop.yaml .

We had a discussion w/ the scale team on this, the test structure may be changing in the near future so while we do want to add a vm test_cr here at that point, for now we will have e2e VM workload test coverage in the benchmark-runner [CI](https://github.com/redhat-performance/benchmark-runner/blob/main/.github/workflows/Nightly_Func_CI.yml) -- will need to add fio to that once we get this PR merged.

roles/fio_distributed/templates/server_vm.yml.j2
jeniferh: The other bus option is sata which can have very poor performance, however it is the default for usability reasons (for ex. if the guest OS didn't have virtio drivers installed). I can add it as an option, I just don't want to encourage it's usage due to the performance implications :)

roles/fio_distributed/templates/server_vm.yml.j2
jeniferh: Good suggestion, we could refactor this for all VM templates.. will take a look.

roles/fio_distributed/templates/server_vm.yml.j2
bengland2: I agree completely, that's why virtio should be the default here, but it is possible to allow the user to set the driver and this might be useful precisely so we can demonstrate to people how bad the alternatives are, for example.   THere is more than 1 alternative if I remember right.

jeniferh: Ah yeah I forgot about scsi: http://kubevirt.io/api-reference/master/definitions.html#_v1_disktarget  even more of a reason to add an option w/ default set to virtio

roles/fio_distributed/templates/server_vm.yml.j2
jeniferh: This is a very interesting point I did not think about, because the Kubevirt-provided containerDisk images use "cloud editions" that do not install a firewall (for ex. mentioned in fedora [docs](https://docs.fedoraproject.org/en-US/fedora-server/virtualization-vmcloud/)). But longer term we should consider how to extend this to *any* type of guest image and would need to handle the various firewall cmds in that case..

mulbc: I know it has been a while but adding my 2c:
Rather than mounting the containers root to the host, then bind-mounting the device, I would forward the device into the container with the `podman run --volume {{ fio_path}}:/target ` option.

Then I would expect that the entrypoint of the fio container would call `fio --server` anyways. That for me is a more straightforward way that future people will know and understand and we don't need the chroot.

With that, runcmd would shrink to:

```
            - dnf install -y podman
            - "podman run --volume {{ fio_path}}:/target --entrypoint 'fio --server --directory=/target' {{ workload_args.image | default('quay.io/cloud-bulldozer/fio:latest') }}"
```
(untested)

bengland2: @mulbc looks simpler and more like typical podman/docker usage to me, but maybe I'm missing something about why Jen did it the other way?  @jeniferh ?

jeniferh: @mulbc thanks for the review. I wanted to avoid testing inside a container environment if possible since we know it could add some overhead, we could probably carefully tune around that by making the container privileged, turning off seccomp, enabling all CAPs etc. but the idea was chroot is simply a way to use what the container image provides without actually adding another layer to the data-path by running from w/in a full container.

jeniferh:
@jtaleric @rsevilla87 this PR is (finally) ready if I could please get a review

jeniferh:
@RobertKrawitz @ebattat  Could one of you please give this a final review? 
I have tested all the bus type and disk options for VMs locally w/ the operator, next step will be adding a fio testcase to the runner CI (after the current cleanup effort).

roles/fio_distributed/templates/server_vm.yml.j2
```diff
@@ -22,60 +22,70 @@ spec:
...
+          password: ripsaw
           chpasswd: { expire: False }
           bootcmd:
+            - "mkdir -p {{ fio_path }} || true"
 {% if workload_args.storageclass is defined
     or workload_args.hostpath is defined %}
-            - "mkdir -p {{ fio_path }} || true"
-            - mkfs.ext4 /dev/disk/by-id/virtio-data
-            - "mount /dev/disk/by-id/virtio-data {{ fio_path }}"
+            - "[ -e /dev/disk/by-id/*data ] && disk=$(basename `ls /dev/disk/by-id/*data`) && mkfs.ext4 /dev/disk/by-id/$disk && mount /dev/disk/by-id/$disk {{ fio_path }}"
```
RobertKrawitz, line 59: Better to write

$(shopt -s nullglob; basename /dev/disk/by-id/*data)

Note that if you want to use ls like this that $(...) constructs nest; there's no need for the backquote construct.

RobertKrawitz (commented):
Minor change to shell code.

roles/fio_distributed/templates/server_vm.yml.j2
```diff
@@ -22,60 +22,70 @@ spec:
...
+          name: emptydisk
 {% endif %}
     resources:
       requests:
         memory: {{ workload_args.vm_memory | default('5G') }}
   volumes:
     - name: registrydisk
       containerDisk:
-        image: {{ workload_args.vm_image | default('kubevirt/fedora-cloud-container-disk-demo:latest') }}
+        image: {{ workload_args.vm_image | default('quay.io/kubevirt/fedora-cloud-container-disk-demo:latest') }}
```
ebattat, line 48: @jeniferh, Do you think we need to update it to the new fedora image ?
`quay.io/kubevirt/fedora-container-disk-images:34`

jeniferh: Good point, I will leave the operator tag as `:latest` using this newer image, and we can let the runner be opinionated using the `:34` tag and moving up when appropriate.

roles/fio_distributed/templates/server_vm.yml.j2
jeniferh: Thanks for the tip, works well. Also I found I can simplify the emptydisk fallback by adding the 'data' label which makes the code cleaner, but my testing is a bit delayed due to a wrapper issue ([PR376](https://github.com/cloud-bulldozer/benchmark-wrapper/pull/376) should fix it)

jeniferh:
Thanks for the reviews, fixes pushed and tested locally.

roles/fio_distributed/templates/server_vm.yml.j2
ebattat: OK, sounds good
/LGTM

chentex:
Needs this PR https://github.com/cloud-bulldozer/benchmark-operator/pull/711 to retest

jeniferh:
> Needs this PR #711 to retest

Thanks @chentex but I am already using:  default('quay.io/cloud-bulldozer/fio:latest')
I have been seeing the fio pod client complete & index results since your pip upgrade PR merged.

jeniferh:
> > Needs this PR #711 to retest
> 
> Thanks @chentex but I am already using: default('quay.io/cloud-bulldozer/fio:latest') I have been seeing the fio pod client complete & index results since your pip upgrade PR merged.

Ohh now I see, for the e2e check to pass. thanks!

chentex:
Hi @jeniferh can you rebase this PR, it should pull all the latest fixes

jeniferh:
Thanks @chentex for the recent wrapper fixes, I'm still getting e2e failures on non-fio tests, your PR #715 may help there?

PR #687, opened 2021-09-16
Adding multi test run to benchmark

chentex:

### Description

Adding multi test run to benchmark


roles/api_load/templates/api_load.yml
```diff
@@ -61,21 +61,33 @@ spec:
...
+{% for key, value in workload_args.test_list.items() %}
             ocm-load-test
 {% if workload_args.override is defined %}
             {{ workload_args.override }}
 {% else %}
             --test-id={{ uuid }}
             --gateway-url {{ workload_args.gateway_url }}
             --ocm-token={{ workload_args.ocm_token }}
+{% if value.duration is defined %}
+            --duration={{ value.duration}}
```
dry923, line 73: instead of using if's for these, what about just defaulting the value? ieL
```
--duration={{ value.duration | default(60) }}
```

roles/api_load/templates/api_load.yml
```diff
@@ -61,21 +61,33 @@ spec:
...
+{% else %}
             --duration={{ workload_args.duration}}
+{% endif %}
+{% if value.rate is defined %}
+            --rate={{ value.rate}}
```
dry923, line 78: Minor nit but since we have a space after the open {{ we should also have it before the close }}. There are a few of them like that.

roles/api_load/templates/api_load.yml
chentex: ohhh nice trick!

dry923:
A few minor things and then it needs some docs :-D Good work @chentex !!

PR #690, opened 2021-09-24
Kube burner configmap config

rsevilla87:
### Description

With this PR the kube-burner benchmark will be able to read its own configuration from configmaps in the same namespace where is running.

### DIFF
--- a/docs/e2e-ci.md
+++ b/docs/e2e-ci.md
@@ -27,7 +27,7 @@ A Bats test file is a Bash script with special syntax for defining test cases. U
 }
 ```
 
-If every command in the test case exits with a 0 status code (success), the test passes. This workflow uses bats framework to execute end to end tests, these tests are defined in files with the termination *.bats under the [e2e directory](./e2e)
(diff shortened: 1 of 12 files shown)


docs/e2e-ci.md
```diff
@@ -27,7 +27,7 @@ A Bats test file is a Bash script with special syntax for defining test cases. U
 }
 ```
 
-If every command in the test case exits with a 0 status code (success), the test passes. This workflow uses bats framework to execute end to end tests, these tests are defined in files with the termination *.bats under the [e2e directory](./e2e)
+If every command in the test case exits with a 0 status code (success), the test passes. This workflow uses bats framework to execute end to end tests, these tests are defined in files with the termination *.bats under the [e2e directory](../e2e)
```
dry923, line 30: I'd remove the ../e2e reference as it really can only add confusion when the user may be reading from a dir other than docs. If you want to point to it I'd use something like [e2e directory](benchmark-operator/2e2)

whitleykeith (approved):
lgtm

PR #693, opened 2021-10-05
Make completed.yml able to detect job timeouts

rsevilla87:

### Description

The current implementation is not able to detect job timeouts since these jobs are not marked as failed because the pod didn't actually fail, some context:

```shell
$ oc explain job.status.failed
KIND:     Job
VERSION:  batch/v1

FIELD:    failed <integer>

DESCRIPTION:
     The number of pods which reached phase Failed.
```

A good solution is to update this logic to check for job conditions.


jtaleric (approved):
LGTM

dry923:
Looks good. waiting on CI to run

PR #708, opened 2021-11-10
Changing activeDeadlineSeconds default for api_load

chentex:
### Description

When running a full load test this made the
process undergo a forced termination.

### DIFF
--- a/roles/api_load/templates/api_load.yml
+++ b/roles/api_load/templates/api_load.yml
@@ -6,7 +6,7 @@ metadata:
   namespace: '{{ operator_namespace }}'
 spec:
   backoffLimit: 0
-  activeDeadlineSeconds: {{ workload_args.job_timeout|default(3600) }}
+  activeDeadlineSeconds: {{ workload_args.job_timeout | default(28800) }}
   parallelism: {{ workload_args.pod_count | default(1) | int }}
   template:
     metadata:


roles/api_load/templates/api_load.yml
```diff
@@ -6,7 +6,6 @@ metadata:
   namespace: '{{ operator_namespace }}'
 spec:
   backoffLimit: 0
-  activeDeadlineSeconds: {{ workload_args.job_timeout|default(3600) }}
```
dry923, line 9: Is there no timeout anymore in the job? I'm not seeing this moved anywhere just removed.

chentex: I think for this benchmark it is not necessary. I feel it is difficult to predict one time, since depends on the number of test being run and the duration of each of them and the cooldown period.

dry923 (approved):
LGTM!

roles/api_load/templates/api_load.yml
sjug: It seems all jobs have this `activeDeadlineSeconds` set, what was the rationale for adding it in the first place?

chentex: There is not definitive time the test should take, it depends on the number of endpoints and the time they are set to run and the cooldowns defined. So all of the process should take a defined amount of time.
The ES indexing and Snappy upload don't have a time out, since they also depend on the amount of data that is going to be generated.

chentex: Currently it is taking around 6H30M just to run the test and some time extra to upload. if we keep this I would go with no less that a 8H timeout.

sjug: Yeah I'm hesitant to remove this given I don't know why it was added to begin with.
Also given it's a widespread setting, do we want it on all tests or is it something we want to remove from all?

chentex: I added the timeout, with an 8H default.

sjug (approved):
/lgtm

You know I love those 1 line PRs

PR #713, opened 2021-11-18
Standardizes using latest tag for images

chentex:
Depends-On: https://github.com/cloud-bulldozer/benchmark-wrapper/pull/396

### Description

image tags are pointing to master as default. But master tag is not being built when a new version is deployed, changing it to latest.

### DIFF
--- a/roles/fs-drift/tasks/main.yml
+++ b/roles/fs-drift/tasks/main.yml
@@ -59,7 +59,7 @@
             spec:
               containers:
                 - name: publisher-container
-                  image: quay.io/cloud-bulldozer/fs-drift:latest
+                  image: "{{ workload_args.image | default('quay.io/cloud-bulldozer/fs-drift:latest') }}"
                   tty: true
                   command: ["/bin/sh", "-c"]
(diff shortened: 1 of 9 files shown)


dry923:
Updated Depends-On requirement line in description

roles/fs-drift/tasks/main.yml
```diff
@@ -59,7 +59,7 @@
             spec:
               containers:
                 - name: publisher-container
-                  image: quay.io/cloud-bulldozer/fs-drift:master
+                  image: quay.io/cloud-bulldozer/fs-drift:latest
```
dry923, line 62: Can we update this line to be in line w/ the others to allow you to overwrite the image. ie:
```
image: "{{ workload_args.image | quay.io/cloud-bulldozer/fs-drift:latest }}"
```

sjug:
I think `pgbench` is setting the image via defaults?

sjug:
Thanks, I don't think `iperf3/templates/client_store.yml.j2` is used but maybe just change that image to take the `workload_args.image` parameter as well?

Otherwise lgtm.

PR #719, opened 2022-01-13
Swap the label selector for the operator

jtaleric:
### Description
Using the label "name=benchmark-operator". The other label wasn't being
applied in non-openshift deployments?


sjug (approved):
/lgtm

jtaleric:
`make deploy` applies the correct labels, however the helm installation does now.

PR #720, opened 2022-01-18
Add CPU resource request to kutomize manifest

jtaleric:
### Description

In smaller deployments (not number of nodes, but number of cores), I
have seen where the operator will not deploy.

Additionally, setting the limit to 2, kustomize will set requests to 2.

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -72,6 +72,8 @@ spec:
         ports:
           - containerPort: 6379
         resources:
+          requests: 
+            cpu: "100m"
           limits:
             cpu: "2.0"
         volumeMounts:


sjug:
Suggest keeping a limit for safety, as is best practice for OpenShift operators. Instead of removing can we set a small request instead? Maybe a half core?

sjug (approved):
/lgtm

100m is very little

jtaleric:
> /lgtm
> 
> 100m is very little

Agreed. However, I am not sure we have any idea how much we need... 100m is just a stake in the ground to start things... maybe we need 500m.

sjug:
Seems low but the goal is to enable the pod to be scheduled, which is pretty much guaranteed now.

PR #721, opened 2022-01-18
Uperf workload with nodeport and loadbalancer type svc

mukrishn:
### Description
To support different service types for uperf workload
NodePort: 
```
      serviceip: true
      servicetype: "nodeport"
```

Loadbalancer: only tested for `metallb` loadbalancers on baremetal.
```
      serviceip: true
      servicetype: "metallb"
      metallb:
        addresspool: "addresspool-l3"
        service_etp: "Cluster"
```

@rsevilla87 and @josecastillolema  I've updated the uperf server to listen on port `30000` for all service types as advised, that avoids complexity in the code.

### DIFF
--- a/docs/uperf.md
+++ b/docs/uperf.md
@@ -260,6 +260,56 @@ Once done creating/editing the resource file, you can run it by:
 # kubectl apply -f config/samples/uperf/cr.yaml # if edited the original one
 # kubectl apply -f <path_to_file> # if created a new cr file
 ```
+### Advanced Service types
+
+Benchmark operator now also supports different service types, it can create `NodePort` and `LoadBalancer` (only metallb) 
+type services along with the current default `ClusterIP` type.
+
+No pre-requisites needed for `NodePort` service, as long as the ports used by uperf(30000 to 30012) are allowed at the node level, 
+which is the cluster default.
+
+```
+NAME                TYPE           CLUSTER-IP      EXTERNAL-IP       PORT(S) 
+uperf-service-np    NodePort       172.30.177.81   <none>            30000:31955/TCP,30001:31935/TCP,30002:31942/TCP,30001:31935/UDP,30002:31942/UDP
+```
+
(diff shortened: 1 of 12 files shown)


jtaleric (commented):
Thanks @mukrishn - only nit is this handles the pod case, but not the VM case... If we want to full-stop move to port 30000 I suggest we modify all the templates.

mukrishn:
Thanks @jtaleric good point, updated the VM server port and that is the only place I could find. PTAL.

> Thanks @mukrishn - only nit is this handles the pod case, but not the VM case... If we want to full-stop move to port 30000 I suggest we modify all the templates.

jtaleric:
The only other headache I can think of is we need to open up the ports in each cloud for the testing...

mukrishn:
> The only other headache I can think of is we need to open up the ports in each cloud for the testing...

Yes, @rsevilla87 already found some and updated https://github.com/cloud-bulldozer/scale-ci-deploy/pull/183

rsevilla87 (changes requested):
Hey @mukrishn, can you update uperf docs to describe these couple of new scenarios?

On the other hand, I think you can add a test case for the nodePort scenario.

rsevilla87:
Hey @mukrishn, tests live now under the e2e directory, check this document https://github.com/cloud-bulldozer/benchmark-operator/blob/master/docs/e2e-ci.md. The tests directory is there since benchmark-wrapper still uses them.

mukrishn:
> Hey @mukrishn, tests live now under the e2e directory, check this document https://github.com/cloud-bulldozer/benchmark-operator/blob/master/docs/e2e-ci.md. The tests directory is there since benchmark-wrapper still uses them.

Thanks Raul, now I've included bats for uperf nodeport. PTAL

mohit-sheth (approved):
LGTM

mohit-sheth:
nodeport test is passing

PR #724, opened 2022-01-25
Helm Chart missing label

jtaleric:
The helm chart was missing a label, which was causing issues with deploy
UPerf. Fixing the template.


whitleykeith (approved):
lgtm

PR #727, opened 2022-02-07
Add new CNI focussed kubelet tests

smalleni:

### Description
1. kubelet-density-cni: This creates deployments of node.js
webservers fronted by services and another deployment of pods that
run curl against the webserver service for readiness probe. This ensures
that the curl pods only go into ready when they are able to reach the
webserver srvice. kube-burner accordingly accounts for this delay in the
podready metrics.

2. kubelet-density-cni-policy: This is similar to above workload
except  a deny-all policy is created in the namespace to begin with and
later a newotk policy to allow http on 8080 is introduced when the
pods/services are created. This also ensures that the curl pods'
readiness probe only succeeds when the flows dictated by the
network policy are applied. kube-burner podready metrics now account for
propagation of this network policy.


rsevilla87:
I think it needs some rebase, to use the new centos stream repo:
``` 
STEP 6/11: RUN dnf install -y --nodocs redis openssl --enablerepo=centos8-appstream-* && dnf clean all
Updating Subscription Management repositories.
Unable to read consumer identity

This system is not registered to Red Hat Subscription Management. You can use subscription-manager to register.

CentOS-8-Appstream-x86_64                        21 kB/s | 8.1 kB     00:00    
Errors during downloading metadata for repository 'centos8-appstream-x86_64':
  - Status code: 404 for http://mirror.centos.org/centos/8/AppStream/x86_64/os/repodata/repomd.xml (IP: 198.15.72.18)
Error: Failed to download metadata for repo 'centos8-appstream-x86_64': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried
Error: error building at STEP "RUN dnf install -y --nodocs redis openssl --enablerepo=centos8-appstream-* && dnf clean all": error while running runtime: exit status 1
make: *** [Makefile:92: image-build] Error 125
Error: Process completed with exit code 2.
```

rsevilla87 (approved):
New workload tests passed. LGTM!

PR #728, opened 2022-02-08
Update AppStream Repo for CentOS 8

smalleni:
CentOS 8 is now EOL, so moving to vault for now.

### DIFF
--- a/image_resources/centos8-appstream.repo
+++ b/image_resources/centos8-appstream.repo
@@ -1,17 +1,17 @@
 [centos8-appstream-x86_64]
 name=CentOS-8-Appstream-x86_64
-baseurl=http://mirror.centos.org/centos/8/AppStream/x86_64/os/
+baseurl=http://mirror.centos.org/centos/8-stream/AppStream/x86_64/os/
 enabled=0
 gpgcheck=0
 
 [centos8-appstream-aarch64]
 name=CentOS-8-Appstream-aarch64
-baseurl=http://mirror.centos.org/centos/8/AppStream/aarch64/os/
+baseurl=http://mirror.centos.org/centos/8-stream/AppStream/aarch64/os/
 enabled=0
 gpgcheck=0
 
 [centos8-appstream-ppc64le]
 name=CentOS-8-Appstream-ppc64le
-baseurl=http://mirror.centos.org/centos/8/AppStream/ppc64le//os/
+baseurl=http://mirror.centos.org/centos/8-stream/AppStream/ppc64le/os/
 enabled=0
 gpgcheck=0


dry923:
@smalleni did you check the new centos stream to see if they have the required packages? http://mirror.centos.org/centos/8-stream/AppStream/x86_64/os/ . It looks like it should be a 1-to-1 swap but i haven't tested it personally.

smalleni:
> @smalleni did you check the new centos stream to see if they have the required packages? http://mirror.centos.org/centos/8-stream/AppStream/x86_64/os/ . It looks like it should be a 1-to-1 swap but i haven't tested it personally.

It would be easy to move the repo over the CentOS stream, but untilwe upgrade the operator image itself, I'd much rather use the exact 1x1 RPMs in the same old repos we were using. This change essentially makes us use what we were using earlier. Moving to stream could potentially introduce newer packages.

dry923:
> > @smalleni did you check the new centos stream to see if they have the required packages? http://mirror.centos.org/centos/8-stream/AppStream/x86_64/os/ . It looks like it should be a 1-to-1 swap but i haven't tested it personally.
> 
> It would be easy to move the repo over the CentOS stream, but untilwe upgrade the operator image itself, I'd much rather use the exact 1x1 RPMs in the same old repos we were using. This change essentially makes us use what we were using earlier. Moving to stream could potentially introduce newer packages.

Yes it could add newer packages, etc but that is exactly like it was before it was EOLd. Updates come out, etc. If you want to remain on a static set of rpms thats one thing but if we want to treat it like it was before I'd argue it should go to the stream release. Not saying one way is more correct than the other.

smalleni:
> > > @smalleni did you check the new centos stream to see if they have the required packages? http://mirror.centos.org/centos/8-stream/AppStream/x86_64/os/ . It looks like it should be a 1-to-1 swap but i haven't tested it personally.
> > 
> > 
> > It would be easy to move the repo over the CentOS stream, but untilwe upgrade the operator image itself, I'd much rather use the exact 1x1 RPMs in the same old repos we were using. This change essentially makes us use what we were using earlier. Moving to stream could potentially introduce newer packages.
> 
> Yes it could add newer packages, etc but that is exactly like it was before it was EOLd. Updates come out, etc. If you want to remain on a static set of rpms thats one thing but if we want to treat it like it was before I'd argue it should go to the stream release. Not saying one way is more correct than the other.

But 8-stream is like saying it could be 8.5. or 8.6 <whatever>. I won't pretend like I know everything about stream but IMHO getting updated RPMs for Fedora 34 for a  package is different from getting the same package for Fedora 35 right? Kind of applying that logic here. Let Raul and others weight in, I'm not married to any particular way of doing this.

dry923:
> > > > @smalleni did you check the new centos stream to see if they have the required packages? http://mirror.centos.org/centos/8-stream/AppStream/x86_64/os/ . It looks like it should be a 1-to-1 swap but i haven't tested it personally.
> > > 
> > > 
> > > It would be easy to move the repo over the CentOS stream, but untilwe upgrade the operator image itself, I'd much rather use the exact 1x1 RPMs in the same old repos we were using. This change essentially makes us use what we were using earlier. Moving to stream could potentially introduce newer packages.
> > 
> > 
> > Yes it could add newer packages, etc but that is exactly like it was before it was EOLd. Updates come out, etc. If you want to remain on a static set of rpms thats one thing but if we want to treat it like it was before I'd argue it should go to the stream release. Not saying one way is more correct than the other.
> 
> But 8-stream is like saying it could be 8.5. or 8.6 . I won't pretend like I know everything about stream but IMHO getting updated RPMs for Fedora 34 for a package is different from getting the same package for Fedora 35 right? Kind of applying that logic here. Let Raul and others weight in, I'm not married to any particular way of doing this.

So there's nothing in the old line "http://mirror.centos.org/centos/8/AppStream/x86_64/os/" that indicates a version so its the same thing unless I'm missing something. That said, it's really not a big deal to me either way.

rsevilla87 (approved):
Seems it works, LGTM. Let's ignore the flaky tests cases

PR #732, opened 2022-02-14
Support for dualstack loadbalancer type services

josecastillolema:
### Description
`dualstack` option will use IPv6 for establishing the connectivity between uperf client(s) and server(s). The environment has to be properly configured with dualstack previous to running the test. For now, it is only supported with LoadBalancer type services.

The same logic can be easily extended to support other scenarios, like pod2pod or pod2clusterIPsvc.

Did not include a test because the loadbalancer type service is platform specific.


rsevilla87:
@josecastillolema seems like with the latest changes this one needs some rebase

josecastillolema:
Does the e2e cluster already support loadbalancer type services?
If not, would it be possible/worth it to [install the MetalLB operator](https://docs.openshift.com/container-platform/4.10/networking/metallb/metallb-operator-install.html) on the e2e cluster in order to include loadbalancer type tests?

PR #734, opened 2022-02-16
SCTP support for pod2service scenarios in uperf workloads

josecastillolema:
### Description
Besides `tcp` and `udp`, we have added `sctp` support for uperf workloads service scenarios.


roles/uperf/templates/service_metallb.yml.j2
```diff
@@ -82,6 +82,47 @@ items:
...
+{% endif %}
+      externalTrafficPolicy: '{{ workload_args.metallb.service_etp | default("Cluster") }}'
+      type: LoadBalancer
+      ports:
+      - name: uperf
+        port: 30000
+        targetPort: 30000
+        protocol: SCTP
+{% for num in range(30001,30012,1) %}
+      - name: uperf-control-tcp-{{num}}
```
jtaleric, line 120: You mention `uperf-control-tcp` here, but have the protocol as SCTP? 

Is the proto sctp or tcp here? This is for the uperf control - not sure if it should be sctp or tcp... I haven't messed with sctp, so flying blind :(

rsevilla87:
@josecastillolema seems like with the latest changes this one needs some rebase

roles/uperf/templates/service_metallb.yml.j2
josecastillolema: We are following the same naming convection for SCTP like the one used for TCP and UDP: https://github.com/cloud-bulldozer/benchmark-operator/blob/84d7f13ba2b0de41ad359ddb1c8b529f0a3ba1e2/roles/uperf/templates/service_metallb.yml.j2#L75-L83

josecastillolema:
I understand [bats](https://github.com/bats-core/bats-detik) is being used to run the k8s e2e tests. Would it be possible (or even worth it) to [enable SCTP](https://docs.openshift.com/container-platform/4.10/networking/using-sctp.html#nw-sctp-enabling_using-sctp) on the e2e cluster to include SCTP tests?
We have to take into consideration that the enablement of SCTP requires a reboot of the host.

rsevilla87:
> I understand [bats](https://github.com/bats-core/bats-detik) is being used to run the k8s e2e tests. Would it be possible (or even worth it) to [enable SCTP](https://docs.openshift.com/container-platform/4.10/networking/using-sctp.html#nw-sctp-enabling_using-sctp) on the e2e cluster to include SCTP tests? We have to take into consideration that the enablement of SCTP requires a reboot of the host.

Yeah, it would be nice to have an e2e test for this scenario, uperf tests are specified at https://github.com/cloud-bulldozer/benchmark-operator/blob/master/e2e/013-uperf.bats, it will be required to add your testcase there.

In the meantime, Im gonna enable sctp in the CI cluster, I'll let you know when ready.

rsevilla87:
Hey @josecastillolema, just finished enabling SCP in the CI cluster, feel free to add the test case now.

josecastillolema:
Thanks @rsevilla87 !
About the tests, I was thinking in including them here: 
https://github.com/cloud-bulldozer/benchmark-operator/blob/master/tests/test_crs/valid_uperf_serviceip.yaml#L34, is that ok?

rsevilla87:
> Thanks @rsevilla87 ! About the tests, I was thinking in including them here: https://github.com/cloud-bulldozer/benchmark-operator/blob/master/tests/test_crs/valid_uperf_serviceip.yaml#L34, is that ok?

The tests directory is deprecated and just used by benchmark-wrapper. If you want to add a new test case you have to update https://github.com/cloud-bulldozer/benchmark-operator/blob/master/e2e/013-uperf.bats and add your new CR file to https://github.com/cloud-bulldozer/benchmark-operator/tree/master/e2e/uperf

rsevilla87 (approved):
Seems like the e2e test passed w/o issues, LGTM!

BTW, I guess that this PR is also assing sctp support to the pod2pod scenario, doesn't it?

josecastillolema:
> Seems like the e2e test passed w/o issues, LGTM!
> 
> BTW, I guess that this PR is also adding sctp support to the pod2pod scenario, doesn't it?

We are about to find out

PR #736, opened 2022-03-02
Adds ability to use different kubernetes configs

chentex:

### Description

Adds ability to use different kubernetes configs

By setting the KUBECONFIG variable you can target different clusters, without having to change contexts on your main kubeconfig.


smalleni:
> Not sure about the need of this change. I think this could be easily addressed by just exporting KUBECONFIG before executing any makefile target.
> 
> ```shell
> export KUBECONFIG=/whatever/kubeconfig
> make deploy
> ```

Agree

PR #735, opened 2022-02-21
corrected indentation with `filter`

mukrishn:
### Description
`macro` is not passing all current variables to the `include`'d template, so `metadata` dict was always undefined and job definition was missing `initContainers` section. Replaced that with regular `include` with `filter` to indent properly

I could see it includes `metadata.yml` on both server and client pod definition now,
```
...
  initContainers:
  - args:
    - |
      python3 stockpile-wrapper.py -s=https://search-perfscale-dev-chmf5l4sh66lvxbnadi4bznl3a.us-west-2.es.amazonaws.com:443 -u=13a31b07-4c90-535f-90cd-37c8bda42c8b -n=${my_node_name} -N=${my_pod_name} --redisip=10.129.40.107 --redisport=6379 --tags=common,k8s,openshift
    command:
    - /bin/sh
    - -c
...
```

### Fixes
#725

PR #738, opened 2022-03-03
patching client pod indentation

### LINKED ISSUE #737: Uperf pod fails when using default pod net + pinning + multi pair
**Describe the bug**
Running a uperf pod test w/ more than one pair + node pinning fails with:

```
 TASK [Start Client(s) w/o serviceIP] ********************************
[...]
yaml.parser.ParserError: while parsing a block mapping
  in "<unicode string>", line 12, column 7:
          template:
          ^
expected <block end>, but found '-'
  in "<unicode string>", line 219, column 7:
          - kind: Job
          ^
fatal: [localhost]: FAILED! => {
    "changed": false
}
```

**To Reproduce**
ex params:
```
  workload:
    name: uperf
    args:
      debug: true
      serviceip: false
      hostnetwork: false
      networkpolicy: false
      multus:
        enabled: false
      pin: true
      pin_server: "worker02"
      pin_client: "worker01"
      samples: 1
      kind: pod
      pair: 10
```

**Expected behavior**
Clients start

**Additional context**
I reverted PR #735 and that appears to resolve the issue, but I don't understand what the actual problem is with the filter yet..

mukrishn:
### Description
Recent PR #735 had an issue in formatting the client_pod.yml when creating multi pair + pin uperf pods.

### Fixes
#737

### DIFF
--- a/roles/uperf/templates/workload.yml.j2
+++ b/roles/uperf/templates/workload.yml.j2
@@ -241,4 +241,5 @@ items:
 {% filter indent(width=4, first=True) %}
 {% include "metadata.yml.j2" %}
 {% endfilter %}
+
 {% endfor %}


mukrishn:
#737 will get fixed by this patch, tested and confirmed by @jeniferh thanks!

rsevilla87 (approved):
LGTM!

PR #740, opened 2022-03-08
Remove redundant metadata section in client

dry923:
### Description
Remove redundant metadata collection from uperf client. If metadata is set to true we will already gather it when server pods launch.


mukrishn (approved):
:+1:

rsevilla87:
Let's ignore fs-drift issues for now -> https://github.com/cloud-bulldozer/benchmark-operator/issues/741

Merging!

PR #742, opened 2022-03-10
fix firewall for centos stream8

ebattat:
### Description
Firewall-cmd fails in cloud-init in centos stream8

### Fixes
Use firewall-offline-cmd instead of Firewall-cmd

### DIFF
--- a/roles/hammerdb/templates/db_mssql_workload_vm.sh.j2
+++ b/roles/hammerdb/templates/db_mssql_workload_vm.sh.j2
@@ -54,8 +54,10 @@ export HOME=/root;
     or workload_args.client_vm.hostpath is sameas true %}
 sudo systemctl stop mssql-server;
 sudo ACCEPT_EULA='Y' MSSQL_PID='Enterprise' MSSQL_SA_PASSWORD='s3curePasswordString' /opt/mssql/bin/mssql-conf setup;
-sudo firewall-cmd --zone=public --add-port=1433/tcp --permanent;
-sudo firewall-cmd --reload;
+# Fixed firewall-cmd failure in cloud-init, for more details https://forums.centos.org/viewtopic.php?f=47&t=52162&p=220915
+sudo firewall-offline-cmd --add-port=1433/tcp;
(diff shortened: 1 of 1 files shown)


jeniferh:
LGTM!

roles/hammerdb/templates/db_mssql_workload_vm.sh.j2
```diff
@@ -54,8 +54,9 @@ export HOME=/root;
     or workload_args.client_vm.hostpath is sameas true %}
 sudo systemctl stop mssql-server;
 sudo ACCEPT_EULA='Y' MSSQL_PID='Enterprise' MSSQL_SA_PASSWORD='s3curePasswordString' /opt/mssql/bin/mssql-conf setup;
-sudo firewall-cmd --zone=public --add-port=1433/tcp --permanent;
-sudo firewall-cmd --reload;
+sudo firewall-offline-cmd --add-port=1433/tcp;
```
RobertKrawitz, line 57: Another option would be to simply move the firewall-cmd below the systemctl start firewalld.

ebattat: @RobertKrawitz, Tried it, **didnt work**. 
For more details [see](https://forums.centos.org/viewtopic.php?f=47&t=52162&p=220915)

RobertKrawitz: Fair enough, please add a comment about that so nobody misunderstands later.

ebattat: I preferred not to add comment inside this Jinja file (too sensitive)

RobertKrawitz: That's simply shell script code.  You can add a comment there.

ebattat: Add comment and test it locally

RobertKrawitz (approved):
LGTM

PR #758, opened 2022-05-09
Allow the user to override the label to find nodes

jtaleric:
Currently benchmark-operator only looks for a specific node label which
I have only seen with OpenShift.

This patch allows the user to provide other labels.


jtaleric:
> @jtaleric LGTM however can we also add a small note in the uperf docs for this? https://github.com/cloud-bulldozer/benchmark-operator/blob/master/docs/uperf.md

Is all CI just generally broken?

PR #760, opened 2022-05-25
Reduce test scope

rsevilla87:
### Description

Should improve CI reliability and reduce test duration. 
- Only perform all FIO I/O operations in one of the tests 
- Disable metadata collection in benchmarks with more than one test case.
- Enable prometheus metric collection in only one kube-burner test


Harshith-umesh:
LGTM

PR #764, opened 2022-05-30
Run tests for modified files

rsevilla87:
### Description

The file test_map has the following structure:

regex:test-filter

Where a regex should be a valid regular expression that is going to be passed to `grep -E` and test-filter is a valid test filter passed to the BATS flag -f.

```
$ bats  -h
  <tests> is the path to a Bats test file, or the path to a directory
  containing Bats test files (ending with ".bats")
  -f, --filter <regex>      Only run tests that match the regular expression
```

---

The behavior should be as is at follows:

If one of the expressions of test_map matched, we append the test_filter to the BATS_TEST variable (Note that there'is a skip test_filter, meant to skip the tests for the files it matches, i.e documentation files), if no expression matches, we run all tests.

Some test files were renamed in order to match with the ansible role names and simplify test_map expressions. i.e rather than `(fs-drift|fs_drift)` we just have `fs-drift`

Also replacing the action `redhat-actions/oc-login@v1` by a simple oc command, this actions is sometimes faulty for reasons I unknown at the moment

I managed to test it in my own fork, where I modified an uperf file, and only uperf tests were executed:

https://github.com/rsevilla87/ripsaw/pull/5
https://github.com/rsevilla87/ripsaw/runs/6657138080?check_suite_focus=true


mkarg75:
LGTM!

PR #765, opened 2022-06-02
Removing unused client_store template from iperf template assets

vishnuchalla:
### Description
Removing client_store.yml.j2 template in iperf workload claiming that it is not used anywhere.

#### Story
While playing around with iperf workload in benchmark-operator i came across this template [client_store.yml.j2](https://github.com/cloud-bulldozer/benchmark-operator/blob/master/roles/iperf3/templates/client_store.yml.j2) and noticed that it does the same job as [client.yml.j2](https://github.com/cloud-bulldozer/benchmark-operator/blob/master/roles/iperf3/templates/client.yml.j2) expect for storing the results in a results.path directory. But the strange thing is `{{results}}` isn't defined anywhere and the template `client_store.yml.j2` isn't used anywhere.

#### Experiment done to confirm the above thoughts
- I have removed `client_store.yml.j2` and run the workload and it ran without any failures as it is using `client.yml.j2` but not `client_store.yml.j2`
- And also i have checked replacing `client.yml.j2` with `client_store.yml.j2` code to check if there is an error on the undefined variable `{{results.path}}`. And it failed with same error as expected. So if the template were being used we must be seeing failures and since there is no known failure and known usage of this file, I have arrived to a conclusion that `client_store.yml.j2` isn't being used any where.

#### Supporting Logs
`fatal: [localhost]: FAILED! => {
    "msg": "The task includes an option with an undefined variable. The error was: 'results' is undefined\n\nThe error appears to be in '/opt/ansible/roles/iperf3/tasks/main.yml': line 48, column 5, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n  - name: Start Client(s)\n    ^ here\n"
}`

`001b[0;32m            }\u001b[0m\r\n\u001b[0;32m        }\u001b[0m\r\n\u001b[0;32m    ]\u001b[0m\r\n\u001b[0;32m}\u001b[0m\n\r\nTASK [iperf3 : Start Client(s)] ************************************************\r\n\u001b[1;30mtask path: /opt/ansible/roles/iperf3/tasks/main.yml:48\u001b[0m\n\u001b[0;34mFile lookup using /opt/ansible/roles/iperf3/templates/client.yml.j2 as file\u001b[0m\n\u001b[0;31mfatal: [localhost]: FAILED! => {\u001b[0m\r\n\u001b[0;31m    \"msg\": \"The task includes an option with an undefined variable. The error was: 'results' is undefined\\n\\nThe error appears to be in '/opt/ansible/roles/iperf3/tasks/main.yml': line 48, column 5, but may\\nbe elsewhere in the file depending on the exact syntax problem.\\n\\nThe offending line appears to be:\\n\\n\\n  - name: Start Client(s)\\n    ^ here\\n\"\u001b[0m\r\n\u001b[0;31m}\u001b[0m\n\r\nTASK [include_role : benchmark_state] ******************************************\r\n\u001b[1;30mtask path: /opt/ansible/playbooks/benchmark.yml:76\u001b[0m\n\r\nTASK [benchmark_state : Failure State] *****************************************`

#### Supporting Points
- No usage of this code using a `"{{ lookup('template', 'client_store.yml.j2') | from_yaml }}" ` in any of the role tasks.
- No code usages: https://github.com/cloud-bulldozer/benchmark-operator/search?q=client_store.yml.j2
- `client.yml.j2` does the same job as `client_store.yml.j2` with very slight difference of not storing results in a `results.path ` directory: https://www.diffchecker.com/YOAeoAUu. And as mentioned above `{{results.path}}` variable in the `client_store.yml.j2` is undefined and the template itself is not being used anywhere.

### Fixes
I think we can simply remove client_store.yml.j2 as the code is buggy and unused. Please feel free to correct me if i am wrong.


mukrishn (approved):

dry923 (approved):
LGTM! Good catch!

PR #752, opened 2022-04-07
Point redis image to multi-arch image on docker.io to support arm architecture

svetsa-rh:
### Description
Point to redis multi-arch image to support arm architecture.

PR #766, opened 2022-06-02
Don't follow :latest for redis

jtaleric:
Hit a weird issue with the latest Redis breaking our automation.

```
redis.exceptions.ResponseError: DENIED Redis is running in protected mode because protected mode is enabled and no password is set for the default user. In this mode connections are only accepted from the loopback interface. If you want to connect from external computers to Redis you may adopt one of the following solutions: 1) Just disable protected mode sending the command 'CONFIG SET protected-mode no' from the loopback interface by connecting to Redis from the same host the server is running, however MAKE SURE Redis is not publicly accessible from internet if you do so. Use CONFIG REWRITE to make this change permanent. 2) Alternatively you can just disable the protected mode by editing the Redis configuration file, and setting the protected mode option to 'no', and then restarting the server. 3) If you started the server manually just for testing, restart it with the '--protected-mode no' option. 4) Setup a an authentication password for the default user. NOTE: You only need to do one of the above things in order for the server to start accepting connections from the outside.
```

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -63,7 +63,7 @@ spec:
           initialDelaySeconds: 5
           periodSeconds: 10
       - name: redis-master
-        image: bitnami/redis:latest
+        image: bitnami/redis:6.2.7
         env:
           - name: MASTER
             value: "true"


jtaleric:
All checks have failed, good time to merge.

rsevilla87:
Hey @jtaleric, I think we can close this one in favor of https://github.com/cloud-bulldozer/benchmark-operator/pull/752, which is based in version 6.2.6

jtaleric:
We fixed it locally in our manifests! Glad yall have it fixed too.

PR #769, opened 2022-06-07
Template serviceAccountName

rsevilla87:

### Description

Rather than setting serviceAccountName: benchmark-operator in every YAML, let's use the metadata template to set it.


dry923 (approved):
I like it. LGTM!

PR #781, opened 2022-07-06
Use uperf pod antiAffinity rules

rsevilla87:
Use podAntiaffinity rules to place uperf client pods in different worker
nodes where the servers are running on

Use podAffinity rules to place all client and server pods in the same node, useful when running more than one pair
All client pods will be scheduled in the same node and servers in another


mukrishn (approved):
LGTM

jtaleric (approved):
lgtm

Issue #780, opened 2022-07-05
Test type 'bidirec' not working in uperf


vishnuchalla:
**Describe the bug**
'bidirec' test type is not working in the uperf.

**Root Cause**
In this test type the client  should open two TCP connections with the server: one is used for the forward test and one for the reverse. But it is trying to use same port and running into error. Attaching the screenshots below.

**To Reproduce**
CR to reproduce
```
apiVersion: ripsaw.cloudbulldozer.io/v1alpha1
kind: Benchmark
metadata:
  name: uperf-burstable
  namespace: benchmark-operator
spec:
  system_metrics:
    collection: true
    prom_url: https://prometheus-k8s.openshift-monitoring.svc.cluster.local:9091
    es_url: https://search-perfscale-dev-chmf5l4sh66lvxbnadi4bznl3a.us-west-2.es.amazonaws.com:443
    prom_token: ${PROM_TOKEN}
    metrics_profile: node-metrics.yml
  elasticsearch:
    url: https://search-perfscale-dev-chmf5l4sh66lvxbnadi4bznl3a.us-west-2.es.amazonaws.com:443
    index_name: ripsaw-uperf
  metadata:
    collection: true
  cleanup: false
  workload:
    name: uperf
    args:
      hostnetwork: false
      serviceip: false
      pin: false
      multus:
        enabled: false
      samples: 2
      pair: 1
      test_types:
        - bidirec
      protos:
        - tcp
        - udp
      sizes:
        - 1024
        - [1024, 512]
      nthrs:
        - 1
        - 2
      runtime: 2
      debug: true
```

**Screenshots**
 - <a href="https://drive.google.com/file/d/1jYEyJZDQiVuwppQr5rbPxGqeM2xWI0ob/view?usp=sharing"><img src="https://drive.google.com/file/d/1jYEyJZDQiVuwppQr5rbPxGqeM2xWI0ob/view?usp=sharing" style="width: 500px; max-width: 100%; height: auto" title="Port binding error due to conflicts" /></a>
 - <a href="https://drive.google.com/file/d/1XroAtcEJSRwlnaeGtoDqSnMpz0vdq4wF/view?usp=sharing"><img src="https://drive.google.com/file/d/1XroAtcEJSRwlnaeGtoDqSnMpz0vdq4wF/view?usp=sharing" style="width: 500px; max-width: 100%; height: auto" title="configmap indicating same ports for both forward and backward test" /></a>

**Fix Suggestion**
[...]

PR #786, opened 2022-07-12
fixing port conflict for uperf bidrec test

vishnuchalla:
### Description
Context: https://github.com/cloud-bulldozer/benchmark-operator/issues/780
### Fixes
Assigning a different port for the reverse test so that it doesn't conflict with the forward test port.
#### Testing screenshot
Screenshot from 2022-07-11 21-42-15


mukrishn:
@vishnuchalla LGTM, squash all commits in to single commit.

PR #795, opened 2022-08-10
Benchmark-operator's affinity preference to workload node

rsevilla87:
### Description

Let's try to take more advantage of the workload node by setting benchmark-operator's affinity to preferably be scheduled on workload nodes.

I think we should hardcore it, as it doesn't affect the default behavior


jtaleric (approved):
seems safe enough :shipit:

Issue #793, opened 2022-07-28
Distributed-FIO: PV not getting mounted to server POD when created with pvcvolumemode: Block


shekharberry:
**Describe the bug**

When in distributed-FIO CR we specify pvcvolumemode: Block (Default is Filesystem), the PV is not getting mounted on the server POD. The PVC gets bound successfully.

The reason for this is for block mode the POD should have -devicePath but it has -mountPath. (https://kubernetes.io/docs/concepts/storage/persistent-volumes/#pod-specification-adding-raw-block-device-path-in-container)

Here's the output of POD in yaml when  pvcvolumemode is Block: http://pastebin.test.redhat.com/1067721
 
**To Reproduce**
1) Deploy Benchmark-operator 
2) Set  pvcvolumemode: Block in distributed-FIO CR
3) 

**Expected behavior**

the pod yaml, should have something like this for block mode: (Note: its an example)

volumeDevices:
    - devicePath: "/dev/xvda"
      name: lvm-block-pvc

but it has

volumeDevices:
    - mountPath: "/dev/xvda"
      name: lvm-block-pvc

Note, for pvcvolumemode: Filesystem (Default Config), the server POD comes up and tests finish successfully.

mykaul:
How would a distributed fio work against a single block? Who'll coordinate the writes and reads?

shekharberry:
This issue came up when I was evaluating LVMO Block performance on SNO (by specifying VolumeMode = block in PVC). The FIO tests wouldn't run because of issue described above
It again came up when I had to measure LSO performance where block PVs are created.
 My plan is to assign one PV (from 1 NVMe) to a single FIO server and measure performance. This way the storage wouldn't have the shared access.

mulbc:
@mykaul - Ripsaw never shares the same PVC between Pods. Ripsaw creates as many Pods as we define in the `servers` variable (in the benchmark CR). Each Pod gets its own PVC. We can tune the amount of parallel IO per PVC with the `numjobs` and `iodepth` variables.
Because of this, block PVCs should work just fine

mulbc:
@shekharberry I think you just need to fix this here:
https://github.com/cloud-bulldozer/benchmark-operator/blob/master/roles/fio_distributed/templates/servers.yaml#L51
Add another if clause for block mode. (could you send a PR please?)

As far as I can see, fio on VMs should work fine already and is not affected by this bug

PR #797, opened 2022-08-19
WIll allow FIO to run test with pvcvolumemode as Block

shekharberry:
### Description
This PR will allow FIO to run tests with pvcblockmode as Block. 

### Fixes
https://github.com/cloud-bulldozer/benchmark-operator/issues/793


config/manager/kustomization.yaml
```diff
@@ -12,5 +12,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: quay.io/cloud-bulldozer/benchmark-operator
-  newTag: latest
+  newName: quay.io/shekharberry/bo
```
jtaleric, line 15: Need to revert.

roles/fio_distributed/tasks/main.yml
```diff
@@ -2,8 +2,8 @@
 - block:
   - name: Provide path for fio - PV
     set_fact:
-      fio_path: /mnt/pvc
-    when: workload_args.storageclass is defined
+      fio_path: /dev/xvda
```
jtaleric, line 5: So, if we don't use block is this still correct?

jtaleric: meaning should we look at the type of storageclass and if it isn't block we keep the default to `/mnt/pvc` ?

jtaleric (changes requested):
question about always defaulting to `/dev/xvda` as well as removing the personal operator path.

roles/fio_distributed/tasks/main.yml
shekharberry: It worked for all 3 cases, pvcvolumemode is undefined, Block or Filesystem. When It's Filesystem /dev/xvda is considered as directory by FIO and for block /dev/xvda is considered as location of Block device.

config/manager/kustomization.yaml
shekharberry: Done

shekharberry:
Yes. This fix works and it would be great if someone can verify the same on
their setup so that the PR is merged.

--Shekhar

jtaleric (approved):
/lgtm

jtaleric:
@shekharberry did you share this PR with the ODF/OCS Team to ensure it doesn't break any of their automation?

PR #799, opened 2022-11-17
Added StorageClass support to Sysbench

kdvalin:
### Description
This PR adds the ability to run sysbench in a CR defined SC.  If a SC is not provided, the volume defaults to `emptyDir` type, as it was doing previously.

It also adds the ability to define options for each phase of `fileio` test.  The current `parameters` act as global parameters, and then it breaks out into `<prepare/run/cleanup>_parameters` that override the global parameters, if you want to remove a global parameter from a phase of the test, providing `null` for the option in the YAML file achieves this.

### Fixes
Adds CR defend SC selection for sysbench.
Creates PVC for the selected SC with CR defined size for sysbench.

### DIFF
--- a/config/samples/sysbench/cr.yaml
+++ b/config/samples/sysbench/cr.yaml
@@ -11,6 +11,8 @@ spec:
       #kind: vm
       # If you want to run this as a VM uncomment the above
       #pin_node: "worker-0.mylab.example.com"
+      #storageclass: ocs-storagecluster-ceph-rbd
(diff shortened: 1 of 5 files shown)


shekharberry:
@kdvalin Can you provide an update here?? IMO this worked smoothly during my testing and it should be merged. 

@jtaleric who would be responsible to merge this PR ??

shekharberry (commented):
LGTM.. +1

PR #802, opened 2022-12-13
FIO Prefill State

kdvalin:
### Description

This PR adds a `Prefilling` state to the FIO benchmark.  Previously prefilling and running the actual benchmark were both under the state of `Running`, with this PR the actual benchmark is running during the `Running` state, and prefilling during the `Prefilling` state.

This update has not been tested using `kind: vm`.


jtaleric:
Confirmed by @shekharberry - merging...

PR #809, opened 2023-04-04
Add es_index parameter to log-generator workload template as it can be set on e2e-benchmarking as ES_BACKEND_INDEX

morenod:
e2e-benchmarking allow to select the index_name to use as ES_BACKEND_INDEX.

But this parameter is not managed on the template so it is never used.

Adding it to the template


dry923 (approved):
LGTM!

PR #811, opened 2023-04-27
zone affinity for uperf clients

mukrishn:
### Description
Pod affinity to keep server and client within same zone to avoid external variability.

### DIFF
--- a/roles/uperf/templates/workload.yml.j2
+++ b/roles/uperf/templates/workload.yml.j2
@@ -56,6 +56,13 @@ spec:
                 values:
                 - {{ ansible_operator_meta.name }}-bench-client-{{ trunc_uuid }}
             topologyKey: kubernetes.io/hostname
+          - labelSelector:
+              matchExpressions:
+              - key: type
+                operator: In
+                values:
+                - {{ ansible_operator_meta.name }}-bench-server-{{ trunc_uuid }}
+            topologyKey: topology.kubernetes.io/zone            
         podAntiAffinity:
           preferredDuringSchedulingIgnoredDuringExecution:
           - weight: 100
--- a/roles/uperf/templates/workload.yml.j2
+++ b/roles/uperf/templates/workload.yml.j2
@@ -56,6 +56,7 @@ spec:
                 values:
                 - {{ ansible_operator_meta.name }}-bench-client-{{ trunc_uuid }}
             topologyKey: kubernetes.io/hostname
+          preferredDuringSchedulingIgnoredDuringExecution:
           - labelSelector:
               matchExpressions:
               - key: type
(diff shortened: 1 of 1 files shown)


jtaleric (commented):
Since this is required during scheduling, if the AZ only has a single node, will the client ever launch?

rsevilla87:
> Since this is required during scheduling, if the AZ only has a single node, will the client ever launch?

I think not since the policy is `requiredDuringSchedulingIgnoredDuringExecution`, I think we should add this policy as `preferredDuringSchedulingIgnoredDuringExecution`

mukrishn:
This is keeping the server and client in the same zone, so why can't we enforce it by keeping `requiredDuring` instead of `preferredDuring` ? at all cases we will have a single zone isn't it.

rsevilla87:
> This is keeping the server and client in the same zone, so why can't we enforce it by keeping `requiredDuring` instead of `preferredDuring` ? at all cases we will have a single zone isn't it.

In bare metal environments by default there's only one zone, for example

mukrishn:
> In bare metal environments by default there's only one zone, for example

Yes so server and client will use the same zone in that case.

jtaleric:
> > Since this is required during scheduling, if the AZ only has a single node, will the client ever launch?
> 
> I think not since the policy is `requiredDuringSchedulingIgnoredDuringExecution`, I think we should add this policy as `preferredDuringSchedulingIgnoredDuringExecution` to prevent single node on AZ test cases to fail

Yeah, I think we need to move to preferred vs required.

rsevilla87 (approved):
@smalleni oh, that's right!; somehow I thought the goal was to deploy the clients on different zones, makes sense now. LGTM

mukrishn:
Note: With this affinity we need at least 2 nodes in a single zone. so [airflow](https://github.com/mukrishn/airflow-kubernetes/commit/d33146da177bdecd0c9c4ac217e0ba8a82fb9bde) worker counts have to be raised to 6.

mukrishn:
@smalleni  just confirmed that baremetal nodes are not having this `topology.kubernetes.io/zone` label, so it might break BM runs. better to be `preferred` ?

jtaleric:
> @smalleni just confirmed that baremetal nodes are not having this `topology.kubernetes.io/zone` label, so it might break BM runs. better to be `preferred` ?

Or we apply the label prior to execution... 

I ran into this running these scripts against non-ocp deployments too (EKS and GKE in particular).

mukrishn:
> Or we apply the label prior to execution...

Not sure if adding labels would be an appropriate solution without knowing the placement of machines. Never played with zoning on baremetal node before, happy to investigate further if that is a way.

For time being lets make it `preferred` ?

rsevilla87:
> > Or we apply the label prior to execution...
> 
> Not sure if adding labels would be an appropriate solution without knowing the placement of machines. Never played with zoning on baremetal node before, happy to investigate further if that is a way.
> 
> For time being lets make it `preferred` ?

makes sense to me

rsevilla87 (approved):
Ignoring FIO failures, lgtm!

PR #812, opened 2023-06-21
iPerf: resolve ipv6 client pod deployment issue with a name format

elenagerman:
### Description
When the ipv6 address is used for traffic generation, and this is the only ip version defined for the server:
 item.status.podIP getting ipv6 ip format value, "podIP": "fd01:0:0:6::bd" as an example
As a result, the client pod failed to start because of the name format failure; only '.' and '_' are acceptable in a name string (not ':').

### Fixes
remove item.status.podIP variable from the name definition

### DIFF
--- a/roles/iperf3/templates/client.yml.j2
+++ b/roles/iperf3/templates/client.yml.j2
@@ -2,7 +2,7 @@
 kind: Job
 apiVersion: batch/v1
 metadata:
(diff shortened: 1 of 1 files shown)


elenagerman:
it is still relevant; the problem exists for single ipv6 openshift deployments and in case the openshift priority on ipv4 will change to the ipv6

Issue #814, opened 2023-06-21
iPerf: client pod failure on ipv6 traffic required for dual-stack cluster


elenagerman:
**Describe the bug**
In the case of a dual-stack cluster, when ipv6 traffic is required, and even the '-6' option is provided, the 'iperf3 -c ' command will always use an ipv4 address.
The openshift preference in the case of dual-stack is always ipv4. And this is a reason the server pod will always have ipv4 as a main pod address (podIP value).

As a result, when '-6' option is used in this case and 'iperf3 -c ' will run to the ipv4 address, the pod execution fails.

**To Reproduce**
1. Deploy dual-stack cluster
2. apply this benchmark:
  `

apiVersion: ripsaw.cloudbulldozer.io/v1alpha1
kind: Benchmark
metadata:
  name: iperf3-benchmark-iperf3
  namespace: benchmark-operator
spec:
  uuid: "3ac4f6c1-69f3-469d-aa9f-a81382a77007"
  elasticsearch:
    url: 'localhost'
  clustername: kni-qe-3
  metadata:
    collection: true
    privileged: true
    serviceaccount: 'benchmark-operator'
    targeted: false
  workload:
    name: iperf3
    args:
      image: "registry.kni-qe-3.lab.eng.rdu2.redhat.com:5000/cloud-bulldozer/iperf3:latest"
      run_id: "100056"
      pair: 1
      port: 60001
      transmit_type: time
      transmit_value: 100
      length_buffer: 500
      streams: 1
      extra_options_client: '-u -b 0 -6 -J -V'
      pin: true
      pin_server: "openshift-worker-3.kni-qe-3.lab.eng.rdu2.redhat.com"
      pin_client: "openshift-worker-2.kni-qe-3.lab.eng.rdu2.redhat.com"

`
4. wait until client pod failure
5. oc describe client pod to see 'iperf3 -c addr' was used

**Expected behavior**
option to run iPerf ipv6 traffic on dual-stack cluster 

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

elenagerman:
maybe possible solution: https://github.com/cloud-bulldozer/benchmark-operator/pull/815

elenagerman:
Thanks a lot!

PR #815, opened 2023-06-21
iPerf: work-around for ipv6 traffic type on dual-stack cluster

elenagerman:
### Description
In the case of a dual-stack cluster, when ipv6 traffic is required, and even the '-6' option is provided, the 'iperf3 -c ' command will always use an ipv4 address.
The openshift preference in the case of dual-stack is always ipv4. And this is a reason the server pod will always have ipv4 as a main pod address (podIP value).

As a result, when '-6' option is used in this case and 'iperf3 -c ' will run to the ipv4 address, the pod execution fails.

### Fixes
switch to the item.status.podIPs list handling instead of item.status.podIP when ipv6 is required

### DIFF
--- a/roles/iperf3/templates/client.yml.j2
+++ b/roles/iperf3/templates/client.yml.j2
@@ -46,7 +46,9 @@ spec:
         imagePullPolicy: Always
         command: ["/bin/sh", "-c"]
         args:
-          - "iperf3 -c {{ item.status.podIP }}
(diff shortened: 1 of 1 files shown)


elenagerman:
possible solution for the https://github.com/cloud-bulldozer/benchmark-operator/issues/814

rsevilla87 (approved):
lgtm!

